PR #4, opened 2015-11-25
Review of DM-4230 (Port HSC-1355: Improved fringe subtraction)

jdswinbank:


python/lsst/ip/isr/fringe.py
```diff
@@ -112,26 +112,25 @@ def run(self, exposure, fringes, seed=None):
...
+        if not hasattr(fringes, '__iter__'):
             fringes = [fringes]
 
+        mask = exposure.getMaskedImage().getMask()
+        for fringe in fringes:
+            fringe.getMaskedImage().getMask().__ior__(mask)
+            if self.config.pedestal:
+                self.removePedestal(fringe)
+
+        # placeholder implementation for multiple fringe frames
```
jdswinbank: I realise you're only moving this comment around, rather than adding it, but: is there a JIRA ticket for replacing this implementation? If so, could you link to it please? If not, could you create one and then link to it?

hchiang2: Created a new ticket DM-4441 for multi-mode fringe correction and add the ticket number to the comment here.

PR #10, opened 2016-03-10
DM-5287: Port HSC ip_isr changesets to LSST

timj:


python/lsst/ip/isr/isrTask.py
```diff
@@ -610,8 +621,10 @@ def saturationDetection(self, exposure, amp):
         """
         if not math.isnan(amp.getSaturation()):
             maskedImage = exposure.getMaskedImage()
-            dataView = maskedImage.Factory(maskedImage, amp.getRawBBox())
-            isr.makeThresholdMask(
+            for box in (amp.getRawBBox(), amp.getRawHorizontalOverscanBBox(),\
```
timj: Remove backslash.

timj:
I think this ticket branch needs to be rebased. Github is complaining.

python/lsst/ip/isr/isrTask.py
```diff
@@ -597,7 +599,16 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
...
             exp = dataRef.get(datasetType, immediate=immediate)
         except Exception, e:
-            raise RuntimeError("Unable to retrieve %s for %s: %s" % (datasetType, dataRef.dataId, e))
+            try:
+                if not self.config.fallbackFilterName:
+                    raise ValueError("No fallback filter specified")
+                exp = dataRef.get(datasetType, filter=self.config.fallbackFilterName, immediate=immediate)
+                self.log.warn("Using fallback calibration from filter %s" % self.config.fallbackFilterName)
+            except ValueError as e:
+                raise RuntimeError("Unable to retrieve %s for %2: %s and not fallbackFilterName defined" %\
```
timj: The `%2` looks wrong here. Given the code being replaced I think it should be `%s`.

timj: "and no fallbackFilterName" rather than "and not fallbackFilterName"?

python/lsst/ip/isr/isrTask.py
```diff
@@ -597,7 +599,16 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
...
+                        (datasetType, dataRef.dataId, e))
+            except:
+                raise RuntimeError("Unable to retrieve %s for %s: %s" % (datasetType, dataRef.dataId, e))
```
timj: The error message here is identical to the error message for the `ValueError` clause so why does the `ValueError` clause exist?

timj:
Is it possible to have a test for the fall back code? Including the case where no fallback is found.

python/lsst/ip/isr/isrTask.py
```diff
@@ -200,7 +200,9 @@ class IsrTaskConfig(pexConfig.Config):
         default = True,
         doc = "Should the gain be applied when applying the brighter fatter correction?"
         )
-
+    fallbackFilterName = pexConfig.Field(dtype=str, doc="Fallback default filter name for "
+        "calibrations", optional=True)
```
timj: Would the string not need to be split if the `doc=` was on a line itself?

python/lsst/ip/isr/isrTask.py
natelust: One was intended to let the user know that something had failed (maybe that dataref was  bad) and the other was intended to inform the user that the failure was specifically may be fixed simply by supplying a fallback filter. I will think about this a bit and see if another way comes to me.

python/lsst/ip/isr/isrTask.py
natelust: done

python/lsst/ip/isr/isrTask.py
```diff
@@ -610,8 +621,10 @@ def saturationDetection(self, exposure, amp):
         """
         if not math.isnan(amp.getSaturation()):
             maskedImage = exposure.getMaskedImage()
-            dataView = maskedImage.Factory(maskedImage, amp.getRawBBox())
-            isr.makeThresholdMask(
+            for box in (amp.getRawBBox(), amp.getRawHorizontalOverscanBBox(),\
+                        amp.getRawVerticalOverscanBBox()):
+                dataView = maskedImage.Factory(maskedImage, box)
+                isr.makeThresholdMask(
                 maskedImage = dataView,
```
natelust: I am not sure what you mean here, they look to be correct syntax to me, but maybe not preferred style wise?

timj: I'm saying that the maskedImage line should be indented relative to the `isr.makeThresholdMask` above it. Without that it's hard to tell at first glance that the `maskedImage` line contains arguments for the previous line. Previously there was the correct indentation but the `for` loop moved the `isr.makeThresholdMask` in without adjusting the arguments to it.

timj: Regarding the spaces. It seems that the LSST python style guide inexplicably insists on the spaces so you have to leave them in. It's the indenting that is the main problem.

python/lsst/ip/isr/isrTask.py
natelust: I think I have a cleaner way to convey the information now.

python/lsst/ip/isr/isrTask.py
```diff
@@ -597,7 +599,15 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
...
-            raise RuntimeError("Unable to retrieve %s for %s: %s" % (datasetType, dataRef.dataId, e))
+            try:
+                exp = dataRef.get(datasetType, filter=self.config.fallbackFilterName, immediate=immediate)
+                self.log.warn("Using fallback calibration from filter %s" % self.config.fallbackFilterName)
+            except:
+                extraMessage = ''
+                if not self.config.fallbackFilterName:
+                    extraMessage += 'And no fallback filter specified'
+                raise RuntimeError("Unable to retrieve %s for %s %s: %s" % (datasetType, dataRef.dataId,
+                                                                            extraMessage, e))
```
timj: I think this will lead to a space between the text and the colon if no `extraMessage` is provided. I'd rather it didn't but your choice.

python/lsst/ip/isr/isrTask.py
```diff
@@ -597,7 +599,15 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
...
             exp = dataRef.get(datasetType, immediate=immediate)
         except Exception, e:
-            raise RuntimeError("Unable to retrieve %s for %s: %s" % (datasetType, dataRef.dataId, e))
+            try:
+                exp = dataRef.get(datasetType, filter=self.config.fallbackFilterName, immediate=immediate)
+                self.log.warn("Using fallback calibration from filter %s" % self.config.fallbackFilterName)
+            except:
+                extraMessage = ''
+                if not self.config.fallbackFilterName:
+                    extraMessage += 'And no fallback filter specified'
```
timj: lower case "A".

python/lsst/ip/isr/isrTask.py
```diff
@@ -597,7 +599,16 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
         try:
             exp = dataRef.get(datasetType, immediate=immediate)
         except Exception, e:
```
timj: Please consider removing this `e` as I don't think it's used.

python/lsst/ip/isr/isrTask.py
timj: Is the `e` here the correct `e`?

python/lsst/ip/isr/isrTask.py
```diff
@@ -610,13 +620,15 @@ def saturationDetection(self, exposure, amp):
...
-                maskedImage = dataView,
-                threshold = amp.getSaturation(),
-                growFootprints = 0,
-                maskName = self.config.saturatedMaskName,
-            )
+            for box in (amp.getRawBBox(), amp.getRawHorizontalOverscanBBox(),
+                        amp.getRawVerticalOverscanBBox()):
+                dataView=maskedImage.Factory(maskedImage, box)
+                isr.makeThresholdMask(
+                maskedImage=dataView,
```
timj: The indenting is still wrong here.

python/lsst/ip/isr/isrTask.py
natelust: I was thinking you were talking about line 625/626, I see what you are saying now and they are now indented.

python/lsst/ip/isr/isrTask.py
natelust: This is the behavior of the code coming from HSC, but I am inclined to agree with you that it would make more sense if the second except generated the e

timj:
I have one question which I guess is for @jdswinbank; the commits on this PR are still associated with @PaulPrice but it seems to me that 2f3ed5e is not at all like the original commit cherry-picked from HSC (at least I assume that given how many changes resulted from this PR) that this attribution seems somewhat confusing.

PaulPrice:
I've had the same confusion on seeing commits supposedly written by me but bearing little resemblance in places to what I wrote.  I suspect that when squashing commits, git puts as author of the squashed commit either the author of the first commit or the author of the largest commit.
Thankfully, we'll soon be done with the HSC merge, and shouldn't have this problem on this scale again.

jdswinbank:
When dealing with code reviews on the HSC merge, we've encouraged folks to treat them as if they were code written natively to address LSST -- that is, changes based on the review comments are squashed back into the original. In some cases, as you point out, that means the original is substantially rewritten.

The alternative would be to use a somewhat different development process for the HSC merge, in which changes based on the review were made in separate commits. This would make it easier to see exactly how LSST code differs from that which was deployed on HSC, and it would avoid any ambiguity about authorship... at the cost of breaking our usual workflow.

My feeling is that neither of the above alternatives are perfect; in general, we've done the best we can with what we've got. I hope the original author of the code (Paul, in this case) doesn't feel they are unfairly represented by the newly-rewritten LSST history; if so, I apologise.

PaulPrice:
I don't feel strongly about the attribution, as we're all working towards the same goal, but it makes it difficult to identify the true author in cases where we want to ask, "why did you do it this way?"

I do think the decision to "treat them as if they were code written natively to address LSST" wasn't the right one, as it greatly increased the difficulty of comparing the two branches --- `git cherry` wouldn't work, and we were forced to compare diffs, resulting in several false positives.

PR #13, opened 2016-05-19
DM-6147

r-owen:


python/lsst/ip/isr/isrTask.py
```diff
@@ -619,7 +625,7 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
         return exp
 
     def saturationDetection(self, exposure, amp):
-        """!Detect saturated pixels and mask them using mask plane "SAT", in place
+        """!Detect saturated pixels and mask them using mask plane config.saturatedMaskName, in place
```
parejkoj: I like this change!

python/lsst/ip/isr/isrTask.py
```diff
@@ -102,6 +102,11 @@ class IsrTaskConfig(pexConfig.Config):
         doc = "Name of mask plane to use in saturation detection and interpolation",
         default = "SAT",
     )
+    suspectMaskName = pexConfig.Field(
+        dtype = str,
+        doc = "Name of mask plane to use for suspect pixels",
```
parejkoj: Probably worth either incorporating or linking to the text you're adding to afw about what suspect means, here.

r-owen: I agree. This is the most important place to document what the value means.

python/lsst/ip/isr/isrTask.py
r-owen: I'm not at all convinced this should even be a config param, but as it stands the old docs were wrong.

tests/testSuspectMasking.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        maxVal = 32000
+        fracSuspect = 0.3
+        suspectLevel = maxVal*(1 - fracSuspect)
+
+        bbox = self.ampInfo.getRawBBox()
+        self.ampInfo.setSuspectLevel(suspectLevel)
+        maskedImage = makeRampMaskedImage(bbox, 0, maxVal)
+        imArr = maskedImage.getImage().getArray()
+        desSetArr = np.where(imArr >= suspectLevel, 1, 0)
```
parejkoj: Why are you doing `foo=where(condition,1,0)` here instead of just `foo=condition`? The latter gives you the boolean array directly.

r-owen: Good catch. Thanks!

parejkoj:
I feel like there should be a test that does both saturation and suspect detection, to help define how those two interact. Whether that should live in testSuspectMasking.py or some other file, I'm not sure, but I feel like it's an important integration test. It would call isrTask.run(), but maybe that suggests that the stuff inside the "for amp in ccd:" should live in a "do_amp()" method, which can then be tested separately?

tests/testSuspectMasking.py
parejkoj: Note: there's a bunch of those in here.

r-owen:
I would like to leave that for DM-6161 if at all possible. We don't presently have any unit test for IsrTask and we really need one.

tests/testSuspectMasking.py
r-owen: Yes. If it's worth doing it's worth overdoing. I found and fixed 4 instances.

parejkoj:
Ok: glad there's a ticket about it. Should I (or you?) add my note above about the possible refactor to that ticket?

r-owen:
I don't understand about the refactoring. Perhaps you should add the note or explain it to me in person.

PR #14, opened 2016-05-26
Tickets/dm 5462

r-owen:


include/lsst/ip/isr.h
```diff
@@ -22,196 +22,9 @@
...
-        lsst::afw::image::MaskedImage<ImagePixelT> const& overscan,
-        double ssize=1.,
-        int sigma=1
-        );
-    
-
-}}} // namespace lsst::ip::isr
-	
-#endif // !defined(LSST_IP_ISR_ISR_H)
+#endif
```
PaulPrice: Needs a newline at the end.

include/lsst/ip/isr/isr.h
```diff
@@ -0,0 +1,217 @@
+// -*- LSST-C++ -*- 
+
+/* 
+ * LSST Data Management System
+ * Copyright 2008, 2009, 2010 LSST Corporation.
```
PaulPrice: Update copyleft year?

include/lsst/ip/isr/isr.h
```diff
@@ -0,0 +1,217 @@
...
+ 
+/**
+  * \file
+  *
+  * \ingroup isr
+  *
+  * \brief Implementation of the templated Instrument Signature Removal 
+  * stage of the nightly LSST Image Processing Pipeline.
+  *
+  * \author Nicole M. Silvestri / ACB, University of Washington
```
PaulPrice: Spell out "ACB"?

include/lsst/ip/isr/isr.h
```diff
@@ -0,0 +1,217 @@
...
+  *
+  * Contact: nms@astro.washington.edu
```
PaulPrice: Remove contact?

include/lsst/ip/isr/isr.h
```diff
@@ -0,0 +1,217 @@
...
+  *
+  * \version
+  *
+  * LSST Legalese here...
```
PaulPrice: Remove?

include/lsst/ip/isr/isr.h
```diff
@@ -0,0 +1,217 @@
...
+        ISR_DFLATid = 0x10,  ///< Dome flat
+        ISR_ILLUMid = 0x20,  ///< Illumination correction
+        ISR_BADPid  = 0x40,  ///< Bad pixel mask
+        ISR_SATid   = 0x80,  ///< Saturated pixels
+        ISR_FRINid  = 0x100, ///< Fringe correction
+        ISR_DARKid  = 0x200, ///< Dark correction
+        ISR_PUPILid = 0x400, ///< Pupil correction
+        ISR_CRREJid = 0x800, ///< Cosmic ray rejection
+        ISR_BACKSUBid = 0x1000, ///< Cosmic ray rejection
+    };
```
PaulPrice: Can we remove all the above (`StageId` and the "signature strings")?

(I realise this is more than what you originally signed up for, so feel free to tell me to take a hike.)

include/lsst/ip/isr/applyLookupTable.h
```diff
@@ -0,0 +1,62 @@
...
+    */
+    template<typename PixelT>
+    int applyLookupTable(
+        afw::image::Image<PixelT> &image,
+        ndarray::Array<PixelT, 1, 1> const &table,
+        PixelT indOffset);
+
+}}} // lsst::ip::isr
+
+#endif
```
PaulPrice: Needs newline at end of file.

include/lsst/ip/isr/applyLookupTable.h
```diff
@@ -0,0 +1,62 @@
...
+    @param[in] table  lookup table
+    @param[in] indOffset  scalar added to image value before truncating to lookup column
+
+    @return the number of pixels whose values were out of range
+    */
+    template<typename PixelT>
+    int applyLookupTable(
+        afw::image::Image<PixelT> &image,
+        ndarray::Array<PixelT, 1, 1> const &table,
+        PixelT indOffset);
```
PaulPrice: Shouldn't the `);` go on the next line?

include/lsst/ip/isr/applyLookupTable.h
```diff
@@ -0,0 +1,62 @@
...
+        return numOutOfRange
+
+    @param[in,out] image  image to which to add the values; modified in place
+    @param[in] table  lookup table
+    @param[in] indOffset  scalar added to image value before truncating to lookup column
+
+    @return the number of pixels whose values were out of range
+    */
+    template<typename PixelT>
+    int applyLookupTable(
```
PaulPrice: Suggest returning `size_t` or `unsigned int` instead of `int` --- negative numbers don't mean anything.

include/lsst/ip/isr/applyLookupTable.h
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+    Add the values in a lookup table to an image, e.g. for non-linearity correction
+
+    The algorithm is as follows:
+        numOutOfRange = 0
+        For each i,j of the image:
+            lookupInd = int(indOffset + image[i,j])
+            if lookupInd not in range [0, table.size() - 1]:
+                set lookupInd to nearest edge and increment numOutOfRange
```
PaulPrice: Is the mask modified to indicate the pixel was out of range?  I suggest `SUSPECT` might be appropriate?

src/applyLookupTable.cc
```diff
@@ -0,0 +1,65 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the LSST License Statement and 
+ * the GNU General Public License along with this program.  If not, 
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include <cstdint>
+#include <iostream>
```
PaulPrice: Shouldn't need `iostream`.  Suspect it was only included for printf-style debugging.

src/applyLookupTable.cc
```diff
@@ -0,0 +1,65 @@
...
+template<typename PixelT>
+int applyLookupTable(
+    afw::image::Image<PixelT> &image,
+    ndarray::Array<PixelT, 1, 1> const &table,
+    PixelT indOffset
+) {
+    int numOutOfRange = 0;
+    auto const maxLookupCol = table.size() - 1;
+    for (auto col = 0, imHeight = image.getHeight(); col < imHeight; ++col) {
+        for (auto imPtr = image.row_begin(col), end = image.row_end(col); imPtr != end; ++imPtr) {
```
PaulPrice: Is there a reason the C++11 range-based iteration wouldn't work?

    for (auto&& pixel : image) {

Failing that, might you not just iterate over the image, rather than the rows and columns separately?

    for (auto ptr = image.begin(); ptr != image.end(); ++ptr) {

tests/testApplyLookupTable.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    imArr = image.getArray()
+    indArr = np.array(imArr + indOffset, dtype=int)
+    maxInd = len(table) - 1
+    numBadPoints = np.sum(indArr < 0)
+    numBadPoints += np.sum(indArr > maxInd)
+    indArr = np.where(indArr < 0, 0, indArr)
+    indArr = np.where(indArr >= maxInd, maxInd, indArr)
+    imArr += table[indArr]
+    return numBadPoints
```
PaulPrice: Is there a reason not to use this as the actual implementation?  We should prefer a python implementation unless there's demonstrated performance advantage to having it in C++.  Given that there's a simple python implementation, I think you need to demonstrate that it's not too slow before we should accept the C++.

tests/testApplyLookupTable.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        bbox = afwGeom.Box2I(afwGeom.Point2I(-31, 22), afwGeom.Extent2I(100, 85))
+        imMin = -5
+        imMax = 2500
+        tableLen = 2000
+        tableSigma = 55
+        for indOffset in (0, -50, 234):
+            for imageClass in (afwImage.ImageF, afwImage.ImageD):
+                inImage = makeRampImage(bbox=bbox, minVal=imMin, maxVal=imMax, imageClass=imageClass)
+                table = np.random.normal(scale=tableSigma, size=tableLen)
```
PaulPrice: This is not deterministic --- the RNG hasn't been seeded.

tests/testApplyLookupTable.py
```diff
@@ -0,0 +1,84 @@
...
+                table = np.array(table, dtype=inImage.getArray().dtype)
+
+                refImage = imageClass(inImage, True)
+                refNumBad = referenceApply(image=refImage, table=table, indOffset=indOffset)
+
+                measImage = imageClass(inImage, True)
+                measNumBad = applyLookupTable(measImage, table, indOffset)
+
+                self.assertEqual(refNumBad, measNumBad)
+                self.assertImagesNearlyEqual(refImage, measImage)
```
PaulPrice: At the moment, you're testing that one implementation is the same as the "reference implementation".  But the "reference implementation" hasn't been validated, and so *both* implementations are not *known* to be free of bugs like off-by-one errors.  I think you need to add some tests that explicitly check that the right element from the table is being used, e.g., by having everything in the table being zero except for one element, and then the (single-pixel?) image has values around that corresponding to the non-zero entry in the table.

python/lsst/ip/isr/linearize.py
```diff
@@ -0,0 +1,228 @@
...
+
+        @warning only checks the first record of the amp info catalog
+
+        @param[in] detector  detector information (an instance of lsst::afw::cameraGeom::Detector)
+
+        @throw RuntimeError if anything doesn't match
+        """
+        ampInfoType = detector.getAmpInfoCatalog()[0].getLinearityType()
+        if self.LinearityType != ampInfoType:
+            raise RuntimeError("Linearity types don't match: %s != %s" % (self.LinearityType, ampInfoType))
```
PaulPrice: Would an `assert` be appropriate?  Is using the wrong linearity type a programming error?

python/lsst/ip/isr/linearize.py
```diff
@@ -0,0 +1,228 @@
...
+        """Correct non-linearity
+
+        @param[in] image  image to be corrected (an lsst.afw.image.Image)
+        @param[in] detector  detector information (an instance of lsst::afw::cameraGeom::Detector)
+        @param[in] log  logger (an lsst.pex.logging.Log), or None to disable logging;
+                    a warning is logged if amplifiers are skipped or other worrisome events occur
+
+        @return an lsst.pipe.base.Struct containing at least the following fields:
+        - numAmps number of amplifiers found
+        - numLinearized  number of amplifiers linearized
```
PaulPrice: I don't think there's any reason for this correction to iterate over amplifiers --- the `detector` isn't used except to get the data for an individual amplifier.  I suggest that the `__call__` should be devoted to just applying the correction for a single amplifier, and leave the iteration to the caller.  That removes the need to pull out the first amplifier in `checkLinearityType`, and would simplify other checks as well.

python/lsst/ip/isr/linearize.py
```diff
@@ -0,0 +1,228 @@
...
+from lsst.pipe.base import Struct
+from .isrLib import applyLookupTable
+
+__all__ = ["LinearizeBase", "LinearizeLookupTable", "LinearizeSquared"]
+
+class LinearizeBase(object):
+    """Abstract base class functor for correcting non-linearity
+    """
+    __metaclass__ = abc.ABCMeta
+    LinearityType = None
```
PaulPrice: Document `LinearityType`, if only with a comment.

tests/testLinearizeLookupTable.py
```diff
@@ -0,0 +1,232 @@
...
+import lsst.pex.logging as pexLog
+import lsst.afw.geom as afwGeom
+import lsst.afw.image as afwImage
+import lsst.afw.table as afwTable
+import lsst.afw.cameraGeom as cameraGeom
+from lsst.ip.isr import applyLookupTable, LinearizeLookupTable
+
+
+def refLinearize(image, detector, table):
+    """!Reference implementation of non-linearization correction
```
PaulPrice: This is not a reference implementation, it is a copy of the actual implementation.  A reference implementation must have been tested and validated; this has no pedigree.

tests/testLinearizeLookupTable.py
```diff
@@ -0,0 +1,232 @@
...
+
+class LinearizeLookupTableTestCase(lsst.utils.tests.TestCase):
+    """!Unit tests for LinearizeLookupTable"""
+    def setUp(self):
+        # the following values are all arbitrary, but sane and varied
+        self.bbox = afwGeom.Box2I(afwGeom.Point2I(-31, 22), afwGeom.Extent2I(100, 85))
+        self.numAmps = (2, 3)
+        self.colIndOffsets = np.array([[0, -50, 2.5], [37, 1, -3]], dtype=float)
+        self.rowInds = np.array([[0, 1, 4], [3, 5, 2]])
+        numCols = self.numAmps[0] * self.numAmps[1]
```
PaulPrice: No space around operator `*`.

tests/testLinearizeSquared.py
```diff
@@ -0,0 +1,179 @@
...
+        @param[in] linearityType  name of linearity type (a str)
+
+        @return a detector (an lsst.afw.cameraGeom.Detector)
+        """
+        bbox = bbox if bbox is not None else self.bbox
+        numAmps = numAmps if numAmps is not None else self.numAmps
+        sqCoeffs = sqCoeffs if sqCoeffs is not None else self.sqCoeffs
+        schema = afwTable.AmpInfoTable.makeMinimalSchema()
+        ampInfoCat = afwTable.AmpInfoCatalog(schema)
+        boxArr = BoxArr(self.bbox, self.numAmps)
```
PaulPrice: Leave off the `self.` for both `bbox` and `numAmps` or you risk inconsistency.

tests/testLinearizeSquared.py
```diff
@@ -0,0 +1,179 @@
+import unittest
```
PaulPrice: There is far too much here that is duplicated from tests/testLinearizeLookupTable.py .  Please consolidate the code.

python/lsst/ip/isr/isrTask.py
```diff
@@ -412,7 +421,8 @@ def readIsrData(self, dataRef, rawExposure):
                                )
 
     @pipeBase.timeMethod
-    def run(self, ccdExposure, bias=None, dark=None,  flat=None, defects=None, fringes=None, bfKernel=None):
+    def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, defects=None,
+        fringes=None, bfKernel=None):
```
PaulPrice: Check left alignment.

tests/testApplyLookupTable.py
r-owen: Thank you for finding that.

tests/testApplyLookupTable.py
r-owen: @RobertLuptonTheGood was very keen to have this code be as fast as possible. I felt that the only way to get proper control over speed was to hand code it in C++. The implementation is simple enough that I wasn't very worried about it being in C++. Also, the Python is rather clumsy by comparison.

src/applyLookupTable.cc
r-owen: My understanding is that the technique I used is significantly more efficient than linear iteration, as per this documentation from @RobertLuptonTheGood:

    // STL-compliant iterators are available.
    // However, they are not very efficient because the image data may not be contiguous
    // so these iterators must test for end-of-row on every increment.
    // (By the way, we do guarantee that an image's row data is contiguous).
    // iterator
    for (ImageT::iterator ptr = img.begin(), end = img.end(); ptr != end; ++ptr) {
        *ptr = 100;
    }

Unless that documentation is out of date I would rather not incur the cost.

include/lsst/ip/isr/isr.h
r-owen: I'll remove the author and stuff below.

include/lsst/ip/isr/isr.h
r-owen: Good idea. I had not noticed that stuff.

tests/testApplyLookupTable.py
r-owen: Far enough. I put something in to generate known values and test against those. It's a hair more complex than perhaps is ideal, but does the job.

include/lsst/ip/isr/applyLookupTable.h
r-owen: You are right, but I prefer to use int unless there is a real need to do otherwise, because C++ is picky about mixing signed and unsigned values. I'm not convinced this qualifies as a need. Also, this will primarily be used in Python, which does not do anything useful with unsigned ints (except in numpy arrays).

include/lsst/ip/isr/applyLookupTable.h
r-owen: It's an interesting idea, but I'm not happy about co-opting SUSPECT for this purpose. It does suggest that the code should operate on masked images so a mask bit could be set should that be wanted in the future.

python/lsst/ip/isr/linearize.py
r-owen: A wrong linearity type is a configuration error, not a programming error. The amp info table is inconsistent with the persisted lookup table.

src/applyLookupTable.cc
PaulPrice: Agreed.

python/lsst/ip/isr/linearize.py
r-owen: You raise a good point, but I respectfully disagree with your conclusion. The heart of the problem is that there ought to be only one linearity type for the whole detector, but I know of no good way to represent table-wide properties in a catalog. A bad way is to use the table's metadata, but that is not type safe. Another alternative is to make linearity type a property of `Detector` instead of `AmpInfoCatalog`, but that separates related data. Of these three options, I prefer the one I chose.

Getting back to the linearizer...I want to persist a whole detector's worth of lookup table for each linearizer. This feels logical to me, since the whole detector will be corrected in one pass through the ISR task, and strikes a good balance between persisting too much information in one blob (the whole focal plane worth of lookup tables) and too little (one amplifier's lookup table). Having done this, as a functor, I think it is logical for `__call__` to correct the whole CCD image. Finally, `Detector` contains some information that is useful, and it would be annoying to have to pass in both a detector and an amp info catalog.

python/lsst/ip/isr/linearize.py
r-owen: Excellent point

include/lsst/ip/isr/applyLookupTable.h
PaulPrice: Using the correct type goes beyond the compiler, but it's also a signal to the user about what it's for and what they can expect.

tests/testLinearizeSquared.py
r-owen: Good catch. Thanks!

python/lsst/ip/isr/linearize.py
PaulPrice: Why do you say that "there ought to be only one linearity type for the whole detector"?  That doesn't seem to be necessary, and if you take that away then I think it's very natural for the linearizer to only care about a single amp.

I see no reason why you can't persist the amp's lookup tables individually.  Alternatively, if you must have them all together then you can pickle a `list` of linearizers.

r-owen: I think it is a needless complication to allow multiple linearity models per amplifier since lookup tables can handle anything. Furthermore, I doubt it is possible to persist different classes using the same dataset type (e.g. two different linearizers) in which case we'd need to make a single class that can handle different models. In my opinion the current design is far simpler and should suffice. @RobertLuptonTheGood gave me permission to assume one linearity model for all amplifiers on a detector, so he may wish to weigh in.

tests/testLinearizeLookupTable.py
r-owen: Good point. I'll think about ways to improve this.

PR #18, opened 2016-08-15
DM-7233: Pytest updates

timj:


tests/testSaturationCorrection.py
```diff
@@ -21,24 +21,22 @@
...
-class IsrTestCases(unittest.TestCase):
+class IsrTestCases(lsst.utils.tests.TestCase):
 
     def testSaturation(self):
         saturation = 1000
 
         bbox = afwGeom.Box2I(afwGeom.Point2I(0, 0), afwGeom.Point2I(19, 19))
         maskedImage = afwImage.MaskedImageF(bbox)
         maskedImage.set(100, 0x0, 1)
-        exposure = afwImage.ExposureF(maskedImage, None)
```
pschella: Why is this here? Was it supposed to be tested?

tests/testSaturationCorrection.py
```diff
@@ -59,7 +57,6 @@ def testSaturation(self):
         )
 
         mask = maskedImage.getMask()
-        bitmaskBad = mask.getPlaneBitMask('BAD')
```
pschella: Was this supposed to be tested?

tests/testDefect.py
```diff
@@ -81,7 +80,9 @@ def testDefectBase(self):
         im = ccdImage.getImage()
         for d in defectList:
             intrp = im.Factory(im, d.getBBox())
-            self.assertImagesEqual(intrp, self.setVal)
+            expect = np.empty_like(intrp.getArray())
+            expect[:] = self.setVal
```
pschella: Why not `expect = np.copy(intrp.getArray())`?

tests/testDefect.py
```diff
@@ -31,10 +32,8 @@
 import lsst.afw.display.utils as displayUtils
 import lsst.ip.isr as ipIsr
 
-try:
-    type(display)
-except NameError:
-    display = False
+# set to True to display images in ds9
+display = False
```
pschella: Not sure if I like either solution. I assume the former was to allow this to be imported and pick up `display` if it existed?

tests/testMaskNans.py
```diff
@@ -20,22 +20,18 @@
...
+import lsst.utils.tests
 import lsst.afw.image as afwImage
 import lsst.afw.display.ds9 as ds9
 from lsst.ip.isr import maskNans
 
-try:
-    debug
-except NameError:
-    debug = False
+debug = False
```
pschella: Again, confused as to why it was done this way. And hence not sure I like the new solution.

tests/testSaturationCorrection.py
timj: Looks like the code using `exposure` was removed in commit dd64d974b781de74f813290a46f7e77c7686ea30

parejkoj: This line (and the other removed one) was flagged as unused. @timj found the related commit.

tests/testSaturationCorrection.py
parejkoj: This line (and the other removed one) was flagged as unused. @timj found the related commit.

tests/testDefect.py
parejkoj: I don't think it matters: I'm refilling it with self.setVal. I guess I'd rather just start with an uninitialized array, to ensure it doesn't accidentally use the values it wants to compare against.

tests/testDefect.py
parejkoj: I have no idea, but it's pretty strange. We don't have a good unified way to set a "display things in tests" variable. lsstDebug is kind of like that, but there are some tickets to rethink that system.

PR #19, opened 2016-09-09
DM-7296: Port to python 3

parejkoj:
Convert ip_isr to python3.

### DIFF
--- a/python/lsst/ip/isr/assembleCcdTask.py
+++ b/python/lsst/ip/isr/assembleCcdTask.py
@@ -39,8 +39,8 @@ class AssembleCcdConfig(pexConfig.Config):
     )
     doRenorm = pexConfig.Field(
         doc="renormalize to a gain of 1? (ignored if setGain false). "
-        "Setting to True gives 1 ADU per electron. "
-        "Setting to True is not recommended for mosaic cameras because it breaks normalization across "
+            "Setting to True gives 1 ADU per electron. "
+            "Setting to True is not recommended for mosaic cameras because it breaks normalization across "
             "the focal plane. However, if the CCDs are sufficiently flat then the resulting error "
             "may be acceptable.",
         dtype=bool,
@@ -132,14 +132,14 @@ class AssembleCcdTask(pipeBase.Task):
     \dontinclude exampleUtils.py
     Create some input images with the help of some utilities in examples/exampleUtils.py
     \skip makeAssemblyInput
-    \until inputData
+    @until inputData
     The above numbers can be changed.  The assumption that the readout corner is flipped on every other amp is
     hardcoded in createDetector.
 
     \dontinclude runAssembleTask.py
     Run the assembler task
     \skip runAssembler
-    \until frame += 1
+    @until frame += 1
 
     <HR>
     To investigate the \ref ip_isr_assemble_Debug, put something like
@@ -197,7 +197,7 @@ class AssembleCcdTask(pipeBase.Task):
         ccd = None
         if hasattr(assembleInput, "has_key"):
             # Get a detector object for this set of amps
-            ccd = assembleInput.values().next().getDetector()
+            ccd = next(assembleInput.values()).getDetector()
             # Sent a dictionary of input exposures, assume one amp per key keyed on amp name
 
             def getNextExposure(amp):
--- a/python/lsst/ip/isr/assembleCcdTask.py
+++ b/python/lsst/ip/isr/assembleCcdTask.py
@@ -197,7 +197,7 @@ class AssembleCcdTask(pipeBase.Task):
         ccd = None
         if hasattr(assembleInput, "has_key"):
             # Get a detector object for this set of amps
-            ccd = assembleInput.itervalues().next().getDetector()
+            ccd = assembleInput.values().next().getDetector()
             # Sent a dictionary of input exposures, assume one amp per key keyed on amp name
 
             def getNextExposure(amp):
--- a/python/lsst/ip/isr/assembleCcdTask.py
+++ b/python/lsst/ip/isr/assembleCcdTask.py
@@ -30,30 +30,31 @@ from .isr import calcEffectiveGain
 
 __all__ = ["AssembleCcdTask"]
 
+
 class AssembleCcdConfig(pexConfig.Config):
     setGain = pexConfig.Field(
-        doc = "set gain?",
-        dtype = bool,
-        default = True,
+        doc="set gain?",
+        dtype=bool,
+        default=True,
     )
     doRenorm = pexConfig.Field(
-        doc = "renormalize to a gain of 1? (ignored if setGain false). "
-            "Setting to True gives 1 ADU per electron. "
-            "Setting to True is not recommended for mosaic cameras because it breaks normalization across "
+        doc="renormalize to a gain of 1? (ignored if setGain false). "
+        "Setting to True gives 1 ADU per electron. "
+        "Setting to True is not recommended for mosaic cameras because it breaks normalization across "
             "the focal plane. However, if the CCDs are sufficiently flat then the resulting error "
             "may be acceptable.",
-        dtype = bool,
-        default = False,
+        dtype=bool,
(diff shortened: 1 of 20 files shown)


python/lsst/ip/isr/assembleCcdTask.py
```diff
@@ -30,30 +30,31 @@
...
+        dtype=bool,
+        default=True,
     )
     doRenorm = pexConfig.Field(
-        doc = "renormalize to a gain of 1? (ignored if setGain false). "
-            "Setting to True gives 1 ADU per electron. "
-            "Setting to True is not recommended for mosaic cameras because it breaks normalization across "
+        doc="renormalize to a gain of 1? (ignored if setGain false). "
+        "Setting to True gives 1 ADU per electron. "
+        "Setting to True is not recommended for mosaic cameras because it breaks normalization across "
```
timj: Look at the line below this one. Different indenting. I think I prefer the 4 character indent...

python/lsst/ip/isr/assembleCcdTask.py
```diff
@@ -195,13 +197,15 @@ def assembleCcd(self, assembleInput):
         ccd = None
         if hasattr(assembleInput, "has_key"):
             # Get a detector object for this set of amps
-            ccd = assembleInput.itervalues().next().getDetector()
+            ccd = assembleInput.values().next().getDetector()
```
timj: Hmm. Is that a real `.next()` method in the iterator "now known as `next()`" way? Or is that a method that just happens to be called `next()`?

parejkoj: I'm not sure it matters: we just want to grab one detector from the input stuff, so the fact that it's a generator should be fine.

python/lsst/ip/isr/assembleCcdTask.py
parejkoj: Interesting. autopep8 munged this. I'll fix it.

python/lsst/ip/isr/isrTask.py
```diff
@@ -139,94 +142,95 @@ class IsrTaskConfig(pexConfig.Config):
...
+        dtype=str,
+        doc="Dataset type for input data; users will typically leave this alone, "
+        "but camera-specific ISR tasks will override it",
+        default="raw",
     )
     fallbackFilterName = pexConfig.Field(dtype=str,
-            doc="Fallback default filter name for calibrations", optional=True)
+                                         doc="Fallback default filter name for calibrations", optional=True)
+
+# \addtogroup LSST_task_documentation
```
timj: aren't the `##` magic for doxygen?

python/lsst/ip/isr/linearize.py
```diff
@@ -20,7 +21,7 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
 import abc
-from itertools import izip
```
timj: `from builtins import zip`

python/lsst/ip/isr/isrTask.py
parejkoj: Oh. I have no idea. linter complains about them.

python/lsst/ip/isr/linearize.py
parejkoj: Ah, I see. Fixed.

python/lsst/ip/isr/assembleCcdTask.py
timj: Ideally all the `next()` methods would be replaced with `__next__()` in our python/C++ layer so that we can use the `next()` function where appropriate.

tests/testLinearizeLookupTable.py
```diff
@@ -203,7 +203,7 @@ def makeDetector(self, bbox=None, numAmps=None, rowInds=None, colIndOffsets=None
                 ampInfo.setName("amp %d_%d" % (i + 1, j + 1))
                 ampInfo.setBBox(boxArr[i, j])
                 ampInfo.setLinearityType(linearityType)
-                ampInfo.setLinearityCoeffs([rowInds[i, j], colIndOffsets[i, j], 0, 0])
+                ampInfo.setLinearityCoeffs(np.array([rowInds[i, j], colIndOffsets[i, j], 0, 0]))
```
timj: so this change is needed on python 3?

python/lsst/ip/isr/isrTask.py
timj: I know it does. Ask @r-owen.

python/lsst/ip/isr/assembleCcdTask.py
parejkoj: Sure. Is there something else to use in this case, though?

tests/testLinearizeLookupTable.py
parejkoj: setLinearityCoeffs raised that it didn't understand the arguments. Doing this converted them to float and it worked.

python/lsst/ip/isr/assembleCcdTask.py
timj:
```python
ccd = next(assembleInput.values()).getDetector()
```
but that will only work if it's an iterator with a `__next__()` method.

python/lsst/ip/isr/isrTask.py
```diff
@@ -224,12 +224,12 @@ class IsrTaskConfig(pexConfig.Config):
...
-## \page IsrTask
-## \ref IsrTask_ "IsrTask"
-## \copybrief IsrTask
-## \}
+# \addtogroup LSST_task_documentation
+# \{
+# \page IsrTask
+# \ref IsrTask_ "IsrTask"
+# \copybrief IsrTask
+# \}
```
timj: This patch is not in the right commit. Surely it wasn't needed to make the tests pass.

tests/testLinearizeLookupTable.py
timj: a bit weird it worked on python 2 before then.

timj: Is this some weirdness caused by python3 division?

timj:
Please ensure `from __future__ import division` is enabled where appropriate so we can get consistent answers on 2 and 3.

tests/testLinearizeLookupTable.py
parejkoj: I don't think so. `rowInds[i, j]` is an int, `colIndOffsets[i, j]` is a float, the others are ints. My guess is that swig got more strict about type conversions.

python/lsst/ip/isr/isrTask.py
r-owen: Yes, unfortunately you have to keep the ## for Doxygen as @timj says.

tests/testLinearizeLookupTable.py
r-owen: I suggest that you specify `dtype=float` for clarity. Otherwise it is not clear why `np.array` is used.

python/lsst/ip/isr/isrTask.py
r-owen: Please remove this change. The ## is required for Doxygen

python/lsst/ip/isr/assembleCcdTask.py
parejkoj: Ok, after looking at it a bit more, the object should be a dictionary in that case, so __next__ should be available. I am rather suspicious of this whole function, but I'm not going to root around in it.

python/lsst/ip/isr/isrTask.py
parejkoj: Sadness. I've restored them.

tests/testLinearizeLookupTable.py
parejkoj: Done. I've also added a note.

python/lsst/ip/isr/isrTask.py
parejkoj: Removed.

PR #23, opened 2016-10-03
ip_isr/DM-7736: Avoid signed/unsigned integer comparison warnings.

TallJimbo:


src/applyLookupTable.cc
```diff
@@ -38,10 +38,10 @@ int applyLookupTable(
     PixelT indOffset
 ) {
     int numOutOfRange = 0;
-    auto const maxLookupCol = table.size() - 1;
+    int const maxLookupCol = table.size() - 1;
```
kfindeisen: Hmm... will this work if `table` is empty? You end up with `table[-1]` down below...

TallJimbo: Easier to fix than to open a new ticket, so I'll just fix it here (by raising an exception if `table` is empty).

PR #25, opened 2017-01-12
DM-8982: Incorrect binning in overscan spline interpolation

PaulPrice:


python/lsst/ip/isr/isr.py
```diff
@@ -345,6 +345,8 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
         diff = numpy.abs(biasArray - medianBiasArr[:, numpy.newaxis])
         biasMaskedArr = numpy.ma.masked_where(diff > collapseRej*stdevBiasArr[:, numpy.newaxis], biasArray)
         collapsed = numpy.mean(biasMaskedArr, axis=1)
+        if collapsed.mask.sum() > 0:
+            collapsed.data[collapsed.mask] = numpy.mean(biasArray.data[collapsed.mask], axis=1)
```
laurenam: I realize you are adding the original values back into the masked pixels in order to plot them below, but might this be a gotcha for anyone later using `collapsed.data` and expecting the non-modified values?  I'm not familiar with the object here, so am not sure what the masked pixels would've been assigned in `collapsed.data`, so forgive my naïveté if this is a non-issue!

PaulPrice: Without this change, the masked pixels have a value of zero and are therefore useless.

laurenam: That's sort of my point, i.e. someone expecting a value of zero in `collapsed.data` for masked pixels (I don't imagine a use case here...but you never know!)

PaulPrice: I don't know why someone would want a zero value for masked pixels, except for my particular use of this below!  I use `collapsed.data*~collapsedMask`, which provides exactly that (`collapsedMask` is `True` for bad pixels).  So the possibility is still there just in case.

PR #27, opened 2017-01-27
DM-8461: Wrap pipe_tasks with pybind11

kfindeisen:
Standardizing on python 3 division, as per the style guide. Needed because the pybind11 wrappers for objects used in this file implement `__truediv__` but not `__div__`.


kfindeisen:
Done. I didn't see any problems from `absolute_import`.

PR #34, opened 2017-03-29
Tickets/dm 9990

RobertLuptonTheGood:


python/lsst/ip/isr/assembleCcdTask.py
```diff
@@ -231,6 +232,37 @@ def getNextExposure(amp):
...
             inMI = getNextExposure(amp).getMaskedImage()
             assemble(outMI, inMI, amp)
+
+        if not self.config.doTrim:      # we need to update the Detector
+            ampInfoCatalog = ccd.getAmpInfoCatalog().copy(deep=True)
+
+            for amp in ampInfoCatalog:
+                assert amp.getHasRawInfo()
+                awidth, aheight = amp.getRawBBox().getDimensions()
+                for bboxName in ("",
```
r-owen: I am uncomfortable having this much knowledge about the names of the raw bounding boxes this far from `AmpInfoRecord`, especially as you don't actually care which raw bounding boxes exist, you just need to process all of them. Furthermore we are planning to rename some of these, and I doubt the person implementing that ticket will find this code.

I suggest moving some or all of this code into AmpInfoRecord, since it knows all about the raw bounding boxes. Here are two alternatives, either of which I think would be clean:
- Provide a method that returns an processed amp info record (as a copy). Then build the new catalog one record at a time (without making a deep copy to start with).
- Add a method to `AmpInfoCatalog` that does all this and returns a new catalog.
Note that either of these could be written in pure Python if you like (though a C++ implementation would be quite simple and could even be defined in the pybind11 wrapper as a lambda).

The following would also work, but I don't really like either one:
- Provide an iterator over raw bounding box names. That seems like awkward API clutter to me.
- Assume that such bounding boxes will always be named "getRaw*BBox" and use grep in the code above to find all getters. That's ugly and could fail if we decide to break the naming convention, but perhaps a *bit* better than what you have done?

python/lsst/ip/isr/assembleCcdTask.py
```diff
@@ -231,6 +232,37 @@ def getNextExposure(amp):
         for amp in ccd:
             inMI = getNextExposure(amp).getMaskedImage()
             assemble(outMI, inMI, amp)
+
+        if not self.config.doTrim:      # we need to update the Detector
```
r-owen: A few more words in the comment would be much appreciated (and moving it to the next line). This is the sort of thing I have in mind (though some of this may already be in the help for `config.doTrim` -- I haven't found that yet):
`Update the amp info table in the Detector to describe untrimmed amplifier data assembled into one image, with the same flips as trimmed assembled data.`

r-owen:
Also please add a unit test for this new code.

PR #36, opened 2017-04-13
Tickets/dm 8108

natelust:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -85,7 +85,9 @@ def defectListFromFootprintList(fpList, growFootprints=1):
         if growFootprints > 0:
             # if "True", growing requires a convolution
             # if "False", its faster
-            fpGrow = afwDetection.growFootprint(fp, growFootprints, False)
+            tempSpans = fp.spans.dilated(growFootprints,
+                                        afwGeom.Stencil.MANHATTAN)
```
fred3m: I think that there should be an extra space in the indentation.

PR #37, opened 2017-04-19
Fix defects being mandatory

mfisherlevine:


python/lsst/ip/isr/isrTask.py
```diff
@@ -399,8 +403,8 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
             fringes = pipeBase.Struct(fringes=None)
         if self.config.doFringe and not isinstance(fringes, pipeBase.Struct):
             raise RuntimeError("Must supply fringe exposure as a pipeBase.Struct")
-
-        defects = [] if defects is None else defects
+        if self.config.doDefect and defects is None:
+            raise RuntimeError("Must supply defects if config.doDefect True")
```
RobertLuptonTheGood: I'd use a verb...

python/lsst/ip/isr/isrTask.py
```diff
@@ -64,6 +64,11 @@ class IsrTaskConfig(pexConfig.Config):
         doc="Apply fringe correction?",
         default=True,
     )
+    doDefect = pexConfig.Field(
+        dtype=bool,
+        doc="Apply defect correction?",
```
RobertLuptonTheGood: "Correct CCD defects such as hot pixels"?

python/lsst/ip/isr/isrTask.py
mfisherlevine: So would I, but none of other error messages there do, so I follow form.

python/lsst/ip/isr/isrTask.py
mfisherlevine: These are all very short, and if you don't know what CCD defects are then I don't think this is the place to learn.

mfisherlevine: OK, done.

PR #39, opened 2017-07-18
Don't warn about missing WCSs when none are expected

RobertLuptonTheGood:


python/lsst/ip/isr/isrTask.py
```diff
@@ -188,6 +188,11 @@ class IsrTaskConfig(pexConfig.Config):
         default=True,
         doc="Assemble amp-level exposures into a ccd-level exposure?"
     )
+    expectWcs = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Input science images are expected to have a WCS (not true for e.g. spectrographs)"
```
parejkoj: I'd phrase this "Expect input science images to have a WCS (set False for e.g. spectrographs)".

python/lsst/ip/isr/isrTask.py
```diff
@@ -426,6 +431,8 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
 
         if self.config.doAssembleCcd:
             ccdExposure = self.assembleCcd.assembleCcd(ccdExposure)
+            if not ccdExposure.getWcs() and self.config.expectWcs:
```
parejkoj: It doesn't really matter, but I'd reverse the order here to short-circuit on getting the WCS if it's not necessary.

python/lsst/ip/isr/isrTask.py
```diff
@@ -426,6 +431,8 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
...
+                self.log.warn("No WCS found in input exposure")
```
parejkoj: Should we include in the warning which exposure had the problem?

RobertLuptonTheGood: There isn't enough context, but if you run the task with `--longlog` you'll get what you need.

PR #40, opened 2017-07-25
Add crosstalk to ISR

SimonKrughoff:


python/lsst/ip/isr/crosstalk.py
```diff
@@ -0,0 +1,205 @@
...
+    def getCoeffs(self, numAmps):
+        """Return a 2-D numpy array of crosstalk coefficients of the proper shape"""
+        return np.array(self.coeffs).reshape((numAmps, numAmps))
+
+
+class CrosstalkTask(Task):
+    """Apply intra-CCD crosstalk correction"""
+    ConfigClass = CrosstalkConfig
+
+    def run(self, exposure):
```
SimonKrughoff: I think that since the coefficients can't be in the config, this signature needs to take either a `dataRef` or an opaque object that contains the details of the crosstalk.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -0,0 +1,205 @@
...
+        Mask planes to ignore.
+    minPixelToMask : `float`
+        Minimum pixel value in source amplifier for which to set
+        ``crosstalkStr`` mask plane in target amplifier.
+    crosstalkStr : `str`
+        Mask plane name for pixels greatly modified by crosstalk.
+    """
+    mi = exposure.getMaskedImage()
+    mask = mi.getMask()
+    bg = calculateBackground(mi, badPixels)
```
SimonKrughoff: Since this is before flatfield, shouldn't the background be calculated per amp?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -0,0 +1,205 @@
...
+
+            jImage = extractAmp(mi, jAmp, iAmp.getReadoutCorner())
+            jImage.getMask().getArray()[:] &= crosstalk  # Remove all other masks
+            jImage -= bg
+
+            iImage.scaledPlus(coeffs[ii, jj], jImage)
+
+    # Set crosstalkStr bit only for those pixels that have been significantly modified
+    # (not necessarily those that are bright originally).
+    mask.clearMaskPlane(crosstalkPlane)
```
SimonKrughoff: I'm missing something here.  Doesn't this clear the mask plane?  Where do we set the crosstalk bit?

parejkoj: I'm curious about @SimonKrughoff 's comment here too.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -0,0 +1,240 @@
...
+    exposure : `lsst.afw.image.Exposure`
+        Exposure for which to measure crosstalk.
+    threshold : `float`
+        Lower limit on pixels for which we measure crosstalk.
+    badPixels : `list` of `str`
+        Mask planes indicating a pixel is bad.
+
+    Returns
+    -------
+    ratios : `list` of `list` of `list` of `float`
```
parejkoj: list of list of list? Isn't there a better way to represent this?

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -0,0 +1,240 @@
...
+
+    for ii, iAmp in enumerate(ccd):
+        iImage = mi.Factory(mi, iAmp.getBBox())
+        iMask = iImage.getMask().getArray()
+        select = (iMask & detected > 0) & (iMask & bad == 0) & np.isfinite(iImage.getImage().getArray())
+        for jj, jAmp in enumerate(ccd):
+            if ii == jj:
+                continue
+            jImage = extractAmp(mi.getImage(), jAmp, iAmp.getReadoutCorner())
+            pixels = (jImage.getArray()[select] - bg)/iImage.getImage().getArray()[select]
```
parejkoj: How does this `select` apply to `jImage`, since it was created on `iMask`?

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -0,0 +1,240 @@
...
+    for ii, jj in itertools.product(range(numAmps), range(numAmps)):
+        if ii == jj:
+            continue
+        values = np.array(ratios[ii][jj])
+        values = values[np.abs(values) < 1.0]  # Discard unreasonable values
+        for rej in range(rejIter):
+            lo, med, hi = np.percentile(values, [25.0, 50.0, 75.0])
+            sigma = 0.741*(hi - lo)
+            good = np.abs(values - med) < rejSigma*sigma
+            print(ii, jj, rej, med, sigma, len(values), good.sum())
```
parejkoj: print statement here isn't going to be helpful for debugging: please remove it, or turn it into a `log.debug()`, with labels for the various bits.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -0,0 +1,240 @@
...
+        Number of rejection iterations.
+    rejSigma : `float`
+        Rejection threshold (sigma).
+
+    Returns
+    -------
+    coeff : `numpy.ndarray`
+        Crosstalk coefficients.
+    coeffErr : `numpy.ndarray`
+        Crosstalk coefficient errors.
```
parejkoj: Mention that these are calculated as the std-dev of the mean?

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -0,0 +1,240 @@
...
+    coeffErr = np.zeros((numAmps, numAmps))
+    coeffNum = np.zeros((numAmps, numAmps), dtype=int)
+
+    for ii, jj in itertools.product(range(numAmps), range(numAmps)):
+        if ii == jj:
+            continue
+        values = np.array(ratios[ii][jj])
+        values = values[np.abs(values) < 1.0]  # Discard unreasonable values
+        for rej in range(rejIter):
+            lo, med, hi = np.percentile(values, [25.0, 50.0, 75.0])
```
parejkoj: Should this be using `afw.stats` or some other standard stats package instead?

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -0,0 +1,240 @@
...
+    @classmethod
+    def _makeArgumentParser(cls):
+        parser = super(MeasureCrosstalkTask, cls)._makeArgumentParser()
+        parser.add_argument("--dump-ratios", dest="dumpRatios",
+                            help="Name of pickle file to which to write crosstalk ratios")
+        return parser
+
+    @classmethod
+    def parseAndRun(cls, *args, **kwargs):
+        """Implement scatter/gather"""
```
parejkoj: Should probably document the return value here, since it's a tuple of things.

python/lsst/ip/isr/measureCrosstalk.py
PaulPrice: I can't think of a better description. As mentioned in the docstring, it's a matrix of lists of floats.

python/lsst/ip/isr/measureCrosstalk.py
PaulPrice: The amps are assumed to be the same size, and I want to select the same pixels on `jImage` as on `iImage`.

python/lsst/ip/isr/measureCrosstalk.py
PaulPrice: That would be best, but we don't have these robust stats in afw (yet?).

python/lsst/ip/isr/measureCrosstalk.py
PaulPrice: Modified to "list of list of ndarray", with an explanation that it is a matrix of arrays.

PR #42, opened 2017-12-12
DM-10765: Replace existing WCS classes with SkyWcs

r-owen:


python/lsst/ip/isr/isrTask.py
```diff
@@ -789,6 +804,27 @@ def overscanCorrection(self, exposure, amp):
...
+        \param[in] camera  camera geometry; an lsst.afw.cameraGeom.Camera
+        """
+        self.log.info("Adding a distortion model to the WCS")
+        wcs = exposure.getWcs()
+        if wcs is None:
+            raise RuntimeError("exposure has no WCS")
+        if camera is None:
+            raise RuntimeError("camera is None")
+        detector = exposure.getDetector()
+        if wcs is None:
```
laurenam: wcs --> detector
(you've already checked for the wcs)

Also, should you move the check for the camera to the list here:
https://github.com/lsst/ip_isr/pull/42/commits/03dc329f5b1aa793166bc1f908905a97deadff38#diff-3a9973baa2fe4526d09a95e405f90269R443
i.e.
```python
if self.config.doAddDistortionModel and camera is None:
    raise RuntimeError("Must supply a camera if config.doAddDistortionModel True")
```
in order to bail out earlier?

python/lsst/ip/isr/isrTask.py
```diff
@@ -789,6 +804,27 @@ def overscanCorrection(self, exposure, amp):
             collapseRej=self.config.overscanRej,
         )
 
+    def addDistortionModel(self, exposure, camera):
+        """!Update the WCS in exposure with a distortion model based on camera geometry
```
laurenam: A tad more detail would be nice here (i.e. expand a bit on the "based on camera geometry" statement)

tests/test_addDistortionModel.py
```diff
@@ -0,0 +1,124 @@
+#
+# LSST Data Management System
+# Copyright 2017 LSST Corporation.
```
laurenam: 2018 (darn slow reviewers...!)

tests/test_addDistortionModel.py
```diff
@@ -0,0 +1,124 @@
...
+
+        desiredWcs = self.makeDesiredDistortedWcs()
+        self.assertWcsAlmostEqualOverBBox(desiredWcs, self.exposure.getWcs(), self.bbox)
+
+    def makeMinimalIsrConfig(self):
+        """Return an IsrConfig with all boolean flags disabled"""
+        isrConfig = IsrTask.ConfigClass()
+        for name in isrConfig:
+            if name.startswith("do"):
+                setattr(isrConfig, name, False)
```
laurenam: Is it guaranteed that all boolean flags will always start with "do"?

tests/test_addDistortionModel.py
```diff
@@ -0,0 +1,124 @@
...
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+import unittest
+
+import lsst.utils.tests
+import lsst.afw.geom as afwGeom
+from lsst.afw.coord import IcrsCoord
+from lsst.afw.image import ExposureF, ExposureInfo, Calib, VisitInfo
+from lsst.afw.geom import makeCdMatrix, makeSkyWcs
```
laurenam: I just seems silly to me to have these individual imports in addition to afwGeom.  I will again state my preference for the `import ... as ...` style so that it is clear where the functions came from when they are used lower down in a file without having to look back up to the top (in addition to being consistent with a lot of existing code).

tests/test_addDistortionModel.py
```diff
@@ -0,0 +1,124 @@
...
+class ApplyLookupTableTestCase(lsst.utils.tests.TestCase):
+    """Test IsrTask.addDistortionModel
+    """
+
+    def setUp(self):
+        self.camera = CameraWrapper().camera
+        self.detector = DetectorWrapper().detector
+        self.crpix = afwGeom.Point2D(50, 100)
+        self.crval = IcrsCoord(36 * afwGeom.degrees, 71 * afwGeom.degrees)
+        scale = 1.0 * afwGeom.arcseconds
```
laurenam: https://developer.lsst.io/coding/python_style_guide.html#binary-operators-should-be-surrounded-by-a-single-space-except-for

tests/test_brighterFatter.py
```diff
@@ -36,7 +36,10 @@ def setUp(self):
         self.filename = "bf_kernel.pkl"
         kernel = afwImage.ImageF(17, 17)
         kernel.set(9, 9, 1)
-        kernel.getArray().dump(self.filename)
+        kernelPickleString = kernel.getArray().dumps()
+        # kernel.getArray().dump(self.filename) triggers an "unclosed file" warning with numpy 1.13.1
+        with open(self.filename, 'wb') as f:
+            f.write(kernelPickleString)
```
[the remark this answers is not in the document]
r-owen: numpy.dump is only documented to accept file path string, not a file object.

I'll fix the commit message.

python/lsst/ip/isr/isrTask.py
r-owen: Thanks for catching that bug. I'll add the test you suggest. I wish it didn't have to be in 3 places, but I think you are right.

tests/test_addDistortionModel.py
r-owen: I think it is guaranteed that all flags that control whether an operation is to be performed will start with `do`, and that is what matters here. The idea is to minimally process the data.

r-owen: I don't think it matters in this case. If the test passes, great. If not, the test an easily be updated to add any new flags that control which stages get performed. In any case, the rule need only apply to IsrTask for this test, and self-consistency within one task is more likely that from task to task. It certainly doesn't rise to the level of an RFC.

PR #44, opened 2018-01-26
DM-12375: attach TransmissionCurves

TallJimbo:
Load and attach TransmissionCurves during ISR.  Disabled by default until more cameras actually have defined transmission curves.


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -444,3 +445,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+
+
+def attachTransmissionCurve(exposure, opticsTransmission=None, filterTransmission=None,
+                            sensorTransmission=None, atmosphereTransmission=None):
+    """Attach a TransmissionCurve to an Exposure, given separate curves for different components.
+
+    Parameters
+    ----------
+    exposure: `lsst.afw.image.Exposure`
+        Exposure object to modify by attaching a the product of all given
```
PaulPrice: "a the"

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -444,3 +445,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+        TransmissionCurves in post-assembly trimmed detector coordinates. Must
+        have a valid Detector attached that matches to the detector associated
```
PaulPrice: ``TransmissionCurve``s
    ``Detector``
?

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -444,3 +445,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
     ampImage -= offImage
+
+
+def attachTransmissionCurve(exposure, opticsTransmission=None, filterTransmission=None,
+                            sensorTransmission=None, atmosphereTransmission=None):
+    """Attach a TransmissionCurve to an Exposure, given separate curves for different components.
+
+    Parameters
+    ----------
+    exposure: `lsst.afw.image.Exposure`
```
PaulPrice: I believe we're supposed to have a space either side of the colon:

    exposure : `lsst.afw.image.Exposure`

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -444,3 +445,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+        evaluated in focal-plane coordinates.
+    filterTransmission: `lsst.afw.image.TransmissionCurve`
+        A TransmissionCurve that represents the throughput of the filter
+        itself, to be evaluated in focal-plane coordinates.
+    sensorTransmission: `lsst.afw.image.TransmissionCurve`
+        A TransmissionCurve that represents the throughput of the sensor
+        itself, to be evaluated in post-assembly trimmed detector coordinates.
+    atmosphereTransmission: `lsst.afw.image.TransmissionCurve`
+        A TransmissionCurve that represents the throughput of the atmosphere,
+        assumed to be spatially constant.
```
PaulPrice: Which of the above, if any, are optional?

python/lsst/ip/isr/isrFunctions.py
PaulPrice: There's no check in the below code that things match. So in what sense "must" it match?

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -444,3 +445,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+
+    Returns the TransmissionCurve attached to the exposure.
```
PaulPrice: Returns
    -------
    combined : `lsst.afw.image.TransmissionCurve`
        The combined transmission curve.

python/lsst/ip/isr/isrTask.py
```diff
@@ -249,6 +249,11 @@ class IsrTaskConfig(pexConfig.Config):
     )
     fallbackFilterName = pexConfig.Field(dtype=str,
                                          doc="Fallback default filter name for calibrations", optional=True)
+    doAttachTransmissionCurve = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Construct and attach a wavelength-dependent throughput curve for this CCD image."
```
PaulPrice: End doc with question mark.

python/lsst/ip/isr/isrTask.py
```diff
@@ -375,19 +380,36 @@ def readIsrData(self, dataRef, rawExposure):
         else:
             fringeStruct = pipeBase.Struct(fringes=None)
 
+        if self.config.doAttachTransmissionCurve:
+            opticsTransmission = dataRef.get("transmission_optics")
+            filterTransmission = dataRef.get("transmission_filter")
+            sensorTransmission = dataRef.get("transmission_sensor")
+            atmosphereTransmission = dataRef.get("transmission_atmosphere")
```
PaulPrice: Each of the above is supposed to be optional. Do you want to have toggles for each of them separately? Maybe `try` to read and set them to `None` if they don't exist?

python/lsst/ip/isr/isrTask.py
```diff
@@ -404,6 +426,10 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
         @param[in] fringes  a pipeBase.Struct with field fringes containing
                             exposure of fringe frame or list of fringe exposure
         @param[in] bfKernel  kernel for brighter-fatter correction
+        @param[in] opticsTransmission  a TransmissionCurve for the optics
+        @param[in] filterTransmission  a TransmissionCurve for the filter
+        @param[in] sensorTransmission  a TransmissionCurve for the sensors
```
PaulPrice: "sensor", singular?

python/lsst/ip/isr/isrTask.py
```diff
@@ -900,6 +932,36 @@ def brighterFatterCorrection(self, exposure, kernel, maxIter, threshold, applyGa
...
+            A TransmissionCurve that represents the throughput of the filter
+            itself, to be evaluated in focal-plane coordinates.
+        sensorTransmission: `lsst.afw.image.TransmissionCurve`
+            A TransmissionCurve that represents the throughput of the sensor
+            itself, to be evaluated in post-assembly trimmed detector coordinates.
+        atmosphereTransmission: `lsst.afw.image.TransmissionCurve`
+            A TransmissionCurve that represents the throughput of the atmosphere.
+            Assumed to be spatially constant.
+
+        Returns the TransmissionCurve attached to the exposure.
```
PaulPrice: Same comments as above.

python/lsst/ip/isr/isrFunctions.py
TallJimbo: There's no way for this code to check, but if the caller passes `Detector` for one CCD and a `TransmissionCurve` that's specifically for a different CCD, anyone using the result of this function won't get the right answers.

python/lsst/ip/isr/isrFunctions.py
TallJimbo: All of them; now document.

python/lsst/ip/isr/isrTask.py
TallJimbo: I like the `try` idea (too bad you cant' pass a default to `butler.get` like you can with `dict.get`).

TallJimbo: Hmm, actually, we should probably make it a bunch of config options so it's a hard failure rather than a silently incorrect curve when dataset is missing.

PR #47, opened 2018-02-01
Check that the gain is non-negative when setting the variance

RobertLuptonTheGood:


python/lsst/ip/isr/isrTask.py
```diff
@@ -661,10 +661,16 @@ def updateVariance(self, ampExposure, amp):
         @param[in,out]  ampExposure     exposure to process
         @param[in]      amp             amplifier detector information
         """
-        if not math.isnan(amp.getGain()):
+        gain = amp.getGain()
+        if not math.isnan(gain):
+            if gain <= 0:
```
parejkoj: Why not combine these lines as `if not math.isnan(gain) and gain <= 0:`?

RobertLuptonTheGood: Because the action on NaN and <= 0 is different.

parejkoj:
Um, I hadn't approved this yet. I was writing some comments about your unittest... How did you even manage to merge it with a red change marker in place?

RobertLuptonTheGood:
Because you asked for a unit test and I added one.  I merged and pushed from the command line, as normal.

parejkoj:
I had neither marked this PR as approved, nor marked the original ticket as Reviewed.

RobertLuptonTheGood:
Sorry.  It was a trivial ticket and I didn’t even think to check that you’d finished signing off.  My fault.

tests/test_flatAndIlluminationCorrection.py
```diff
@@ -98,6 +98,30 @@ def testIllum2(self):
     def testIllum3(self):
         self.doIllum(scaling=3.7)
 
+    def testGainAndReadnoise(self):
+        import lsst.afw.image as afwImage
+        from lsst.afw.cameraGeom.testUtils import DetectorWrapper
+        from lsst.ip.isr import IsrTask
```
parejkoj: Why are there imports inside the test method?

tests/test_flatAndIlluminationCorrection.py
```diff
@@ -98,6 +98,30 @@ def testIllum2(self):
...
+        amp = detector[0]
+        amp.setReadNoise(readNoise)
+
+        for gain in [-1, 0, 0.1, 1]:
+            amp.setGain(gain)
+            isrTask.updateVariance(raw, amp)
+            if gain <= 0:
+                gain = 1
+
+        self.assertEqual(raw.variance.get(0, 0), level/gain + readNoise**2)
```
parejkoj: I see three possible branches in the code you added: `gain is nan`, `gain <= 0`, and `gain > 0`. There is only one assert here, so it can only test one of those branches.

Separately, why are you doing `raw.variance.get(0, 0)`, instead of checking all values of `raw.variance`?

tests/test_flatAndIlluminationCorrection.py
RobertLuptonTheGood: There's no downside, but I'm happy to move them.  That's probably the rule anyway (I should know)

tests/test_flatAndIlluminationCorrection.py
RobertLuptonTheGood: The line's indentation is incorrect; it should be in the loop.  I'm not sure how this happened as I checked that the code failed when I reverted the fix.  User error.

I only need to check one pixel as they're all the same, so this is simpler and equivalent.

tests/test_flatAndIlluminationCorrection.py
parejkoj: Yes. As-is, this will fail flake8 (which you apparently aren't running?).

tests/test_flatAndIlluminationCorrection.py
parejkoj: 1. That still doesn't test the nan case.

2. How do you know the code is treating all the pixels correctly unless you actually test all the pixels?

tests/test_flatAndIlluminationCorrection.py
```diff
@@ -98,6 +98,30 @@ def testIllum2(self):
...
+        from lsst.afw.cameraGeom.testUtils import DetectorWrapper
+        from lsst.ip.isr import IsrTask
+
+        isrTask = IsrTask()
+
+        detector = DetectorWrapper().detector
+        raw = afwImage.ExposureF(detector.getBBox())
+
+        level = 10
+        readNoise = 1
```
parejkoj: I was just made aware that `readNoise=1` is not helpful, given the definition of variance below...

tests/test_flatAndIlluminationCorrection.py
RobertLuptonTheGood: What NaN case? 

I'm adding a test to check that we now handle gains <= 0, so I don't need to test all the pixels -- it wouldn't gain anything.

And I don't understand your question about the readNoise.  I'm adding a test that we handle an illegal value from the cameraGeom data, not a test to check that we handle variances correctly.  I would like better test coverage for the IsrTask, but that should be calling the IsrTask, not internals.

parejkoj: The code itself has an `if isnan(gain)`, which means that we need a test case for gain being `NaN`.

Again, how do you know that all of the pixels are treated correctly unless you test them?

The method you are testing is called `updateVariance`. There is a term related to `readNoise**2`. `1` doesn't give you much information when you square it.

tests/test_flatAndIlluminationCorrection.py
leannep: Imports have now been removed from the method and there are no flake8 violations

python/lsst/ip/isr/isrTask.py
leannep: is there anything further to do on this?

tests/test_flatAndIlluminationCorrection.py
leannep: A value 1.5 is now used for read noise.  A test for a NaN value of gain has been added.  Concerning the question about testing all pixels, I agree with Robert that this is outside the scope of this unit test.  This is a unit test for the handling of illegal values from the cameraGeom data. not the correct handling of variances.

PR #49, opened 2018-02-26
DM-12447: Remove TransformMap.transform, Camera.transform and Detector.transform

r-owen:
and update for Detector.getCenter returning a Point2D
instead of a CameraPoint


r-owen:
First two commits squashed

PR #50, opened 2018-03-13
DM-10299: Apply crosstalk correction in decam

mrawls:


python/lsst/ip/isr/isrTask.py
```diff
@@ -465,6 +466,9 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
 
         @return a pipeBase.Struct with field:
          - exposure
+        otherDataRef : `lsst.daf.persistence.butlerSubset.ButlerDataRef`
+            optional dataRef, e.g., needed for DECam crosstalk correction
```
PaulPrice: Isn't it possible to derive the exposure-based `dataRef` from the CCD-based `dataRef`?
And I believe the `run` method is forbidden from receiving a `dataRef`.

mrawls: You're correct that `run` methods generally shouldn't receive a `dataRef`, as per RFC-352. I've added this special case because I need to extract data from other CCDs within `DecamCrosstalkTask`. I use the `dataRef` to get a Butler and retrieve one or more different exposures with different ccdnums as needed. It's not clear to me how I could move this into a `runDataRef` method as suggested by RFC-352, but I'm open to suggestions.

I also thought I'd preserved old-style docstrings here for consistency (this was before the AHM hack day doc-venture), so I'm impressed you found this briefly-existing numpydoc version :)

PaulPrice: I think this will work:

```
def getVisitSensorRefs(sensorRef):
    """Return a data reference for all sensors in the visit

    Parameters
    ----------
    sensorRef : `lsst.daf.persistence.ButlerDataRef`
        Data reference for a single sensor.

    Returns
    -------
    visitSensorRefs : `list` of `lsst.daf.persistence.ButlerDataRef`
        Data references for all sensors in the visit.
    """
    butler = sensorRef.getButler()
    visitKeys = butler.getKeys("raw", "visit")
    dataId = sensorRef.dataId.copy()
    for kk in set(dataId.keys).difference(set(visitKeys)):
        dataId.pop(kk)
    visitRef = butler.dataRef("raw", "visit", dataId)
    return [rr for rr in visitRef.subItems("ccd")]
```

PaulPrice: The established pattern is that `runDataRef` loads everything that's required and passes that to the `run` method. I am concerned that it's naive (because loading the entire focal plane at once for every CCD is not good), but them's the rules. You'll need to get @RobertLuptonTheGood to sign off on your design.

parejkoj: Note that the decision on RFC-352 was "We will permit exceptions to that, by means of RFC — anybody who is (re-)designing or implementing a task, and who doesn't feel that run is appropriate, can file an RFC to explain why and get an exemption." We haven't fixed up our existing cases either.

Your suggested `getVisitSensorRefs()` seems to have its own problems of butler dataId manipulation.

python/lsst/ip/isr/isrTask.py
```diff
@@ -439,7 +439,8 @@ def readIsrData(self, dataRef, rawExposure):
     def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, defects=None,
             fringes=None, bfKernel=None, camera=None,
             opticsTransmission=None, filterTransmission=None,
-            sensorTransmission=None, atmosphereTransmission=None):
+            sensorTransmission=None, atmosphereTransmission=None,
+            otherDataRef=None):
```
SimonKrughoff: The way you've implemented the decam crosstalk correction, I don't think that modifications here are necessary.  I think this should be reverted except to add a `prepCrosstalk` method to the `CrosstalkTask` base class.

SimonKrughoff: That method would be called in the `RunDataRef` method on IsrTask.  You would still have to pass through the source data object, but then it wouldn't be opaque.

python/lsst/ip/isr/isrTask.py
RobertLuptonTheGood: What's the conclusion here?  I agree with John that there are cases when we might need to pass objects that look a bit like dataRefs around (reference catalogues are one), but I didn't see an RFC to permit it here.

 Why is this in `DecamCrosstalkTask` rather than generalising the cross-talk correction code?

jdswinbank: Hey folks — just to be clear, the aim behind this PR was for @mrawls to iterate with @SimonKrughoff on the approach she should take here. As of now, that discussion is still ongoing; let's see how it converges. Should any RFCs or other special measures be needed, rest assured that everything will be done by the book.

SimonKrughoff: I believe the conclusion is to add a (noop) method called `prepCrosstalk` to the default CrosstalkTask.  Then they will override that in the DECam version to assemble need information.   That's a generic solution that other crosstalk implementations can use.

python/lsst/ip/isr/isrTask.py
```diff
@@ -392,11 +402,13 @@ def readIsrData(self, dataRef, rawExposure):
...
 
-        biasExposure = self.getIsrExposure(dataRef, "bias") if self.config.doBias else None
+        biasExposure = self.getIsrExposure(dataRef, self.config.biasDataProductName) \
+            if self.config.doBias else None
         # immediate=True required for functors and linearizers are functors; see ticket DM-6515
         linearizer = dataRef.get("linearizer", immediate=True) if self.doLinearize(ccd) else None
         darkExposure = self.getIsrExposure(dataRef, "dark") if self.config.doDark else None
-        flatExposure = self.getIsrExposure(dataRef, "flat") if self.config.doFlat else None
+        flatExposure = self.getIsrExposure(dataRef, self.config.flatDataProductName) \
+            if self.config.doFlat else None
```
SimonKrughoff: I'm confused why this is in a ticket for adding crosstalk.  It seems to be complicating the issue.

Also, it seems odd to only generalize two of them.  Why not also generalize `dark`?

python/lsst/ip/isr/isrTask.py
```diff
@@ -510,9 +521,9 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
...
 
         if self.config.doCrosstalk:
-            # DECam requires a dataRef for crosstalk correction
-            if otherDataRef:
-                self.crosstalk.run(ccdExposure, otherDataRef)
+            if crosstalkSources is not None:
+                # crosstalkSources is used by DECam
+                self.crosstalk.run(ccdExposure, crosstalkSources)
             else:
                 self.crosstalk.run(ccdExposure)
```
SimonKrughoff: I think you should add the crosstalkSources to the default `CrosstalkTask`.  It can default to `None` in the base class.

python/lsst/ip/isr/isrTask.py
mrawls: This came up when I was writing a test which calls DecamCpIsrTask, which had its own copy-pasted readIsrData method for the specially-named cpBias and cpFlat issue alone. It was slowly diverging from ip_isr's readIsrData which is what I wanted to use anyway, so instead of perpetuating the copy-paste cycle I thought it best to just add this fix. (There aren't such a thing as cpDarks so I let it be.)

SimonKrughoff: I see.  Please add the generalization for `dark`.  I think it would be pretty confusing to someone who uses this to have access to only two of the three.

python/lsst/ip/isr/isrTask.py
mrawls: Think I've got this now.

python/lsst/ip/isr/isrTask.py
```diff
@@ -505,6 +531,13 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
                 self.suspectDetection(ccdExposure, amp)
                 self.overscanCorrection(ccdExposure, amp)
 
+        if self.config.doCrosstalk:
+            if crosstalkSources is not None:
+                # crosstalkSources is used by DECam
+                self.crosstalk.run(ccdExposure, crosstalkSources)
+            else:
+                self.crosstalk.run(ccdExposure)
```
SimonKrughoff: This can now be collapsed to:
```
if self.config.doCrosstalk:
    self.crosstalk.run(ccdExposure, crosstalkSources)
```

PR #53, opened 2018-04-24
DM-14182: subtract fit from overscan; add option to set variance empirically

PaulPrice:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -292,31 +291,57 @@ def illuminationCorrection(maskedImage, illumMaskedImage, illumScale):
...
-                        - 'NATURAL_SPLINE', 'CUBIC_SPLINE', 'AKIMA_SPLINE' (splines)
-    @param[in] order  polynomial order or spline knots (ignored unless fitType
-                      indicates a polynomial or spline)
-    @param[in] collapseRej  Rejection threshold (sigma) for collapsing dimension of overscan
-    @param[in] statControl  Statistics control object
+    """Apply overscan correction in-place
+
+    The ``ampMaskedImage`` and ``overscanImage`` are modified, with the fit
+    subtracted. Note that the ``overscanImage`` should not be a subimage of
+    the ``ampMaskedImage``, to avoid being subtracted twice.
```
mrawls: This is a good point. Is it possible to add a test to check that `overscanImage` is not a subimage of `ampMaskedImage`? You say in your commit message that it's not clear how to detect this case, but I think it's rather important if it's doable.

python/lsst/ip/isr/isrTask.py
```diff
@@ -295,6 +292,11 @@ class IsrTaskConfig(pexConfig.Config):
         default=True,
         doc="Load and use transmission_atmosphere (if doAttachTransmissionCurve is True)?"
     )
+    doEmpiricalReadNoise = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Calculate empirical read noise instead of value from AmpInfo data?"
```
mrawls: Clarify: "instead of using value from amp.getReadNoise?" This also makes me wonder what sets the read noise value in the amplifier info in the first place. Where does it usually come from?

python/lsst/ip/isr/isrTask.py
```diff
@@ -716,10 +734,18 @@ def updateVariance(self, ampExposure, amp):
                               (amp.getName(), gain, patchedGain))
                 gain = patchedGain
 
+            if self.config.doEmpiricalReadNoise and overscanImage is not None:
+                stats = afwMath.StatisticsControl()
+                stats.setAndMask(overscanImage.mask.getPlaneBitMask(maskPlanes))
+                readNoise = afwMath.makeStatistics(overscanImage, afwMath.STDEVCLIP, stats).getValue()
```
mrawls: I'm assuming a plain old `.getValue()` is adequate here even though earlier uses of `makeStatistics` to get, e.g., the mean specifies `.getValue(afwMath.MEAN)`.

python/lsst/ip/isr/isrTask.py
```diff
@@ -887,17 +930,17 @@ def overscanCorrection(self, exposure, amp):
 
         maskedImage = exposure.getMaskedImage()
         dataView = maskedImage.Factory(maskedImage, amp.getRawDataBBox())
+        overscanImage = maskedImage.Factory(maskedImage, amp.getRawHorizontalOverscanBBox())
```
mrawls: Does this assume overscan is always and forever horizontal? I'm thinking of DECam, which has both horizontal and vertical overscan regions (and its own `isr.py`, of course, so this code doesn't break anything, but I still think it should be as generic as possible).

tests/test_empiricalVariance.py
```diff
@@ -0,0 +1,149 @@
...
+    amp.setHasRawInfo(True)
+    amp.setGain(gain)
+    amp.setReadNoise(readNoise)
+    amp.setSaturation(saturation)
+    amp.setSuspectLevel(np.nan)
+    return amp
+
+class EmpiricalVarianceTestCast(lsst.utils.tests.TestCase):
+    def setUp(self):
+        """Constructs a CCD with two amplifiers and prepares for ISR"""
```
mrawls: Could you use a real image from some testdata rather than building a fake one here? I'd be more inclined to believe the test result for a real-life exposure than for this minimalistic fake one.

tests/test_empiricalVariance.py
```diff
@@ -0,0 +1,149 @@
...
+        self.config.doEmpiricalReadNoise = True
+
+        self.task = IsrTask(config=self.config)
+
+    def tearDown(self):
+        del self.exposure
+
+    def testEmpiricalVariance(self):
+        results = self.task.run(self.exposure)
+        exposure = results.exposure
```
mrawls: Mildly confusing variable name (since there is a `self.exposure` floating around already, but this is an exposure created from the result of processing `self.exposure`). Perhaps call one raw and one post-ISR.

tests/test_overscanCorrection.py
```diff
@@ -44,6 +44,9 @@ def testOverscanCorrectionY(self, **kwargs):
         maskedImage = afwImage.MaskedImageF(bbox)
         maskedImage.set(10, 0x0, 1)
 
+        dataBox = afwGeom.Box2I(afwGeom.Point2I(0, 0), afwGeom.Extent2I(10, 10))
+        dataImage = afwImage.MaskedImageF(maskedImage, dataBox)
```
mrawls: Is the point of `dataImage` just to be a small chunk of `maskedImage`? Does it save a lot of time to run the overscan correction test on this rather than the entire `maskedImage` throughout?

python/lsst/ip/isr/isrFunctions.py
PaulPrice: I don't believe it's possible to test, since we have no way of determining whether two images share the same physical pixels. I therefore consider that the only way to deal with this is to document the possibility.

python/lsst/ip/isr/isrTask.py
PaulPrice: The read noise is hard-wired into the camera configuration, often in the form of entries in FITS files.

python/lsst/ip/isr/isrTask.py
PaulPrice: My understanding is that the point of the `overscanCorrection` function is to correct for bias drift as the device is read out. It therefore uses the serial overclock ("horizontal overscan"), because that tracks the bias level as a function of time as the device is read out. I don't believe the parallel overclock ("vertical overscan") is useful for this because it's read out after all the science pixels; to my knowledge we've never used it (for anything). I'm not even sure what use it has unless you have zero bias drift and simply want more pixels to throw into your calculation.

tests/test_empiricalVariance.py
PaulPrice: obs_decam tests include running `processCcd.py` on some real DECam data, and ci_hsc does similarly with real HSC data. Both ran fine under Jenkins.

I think leaving the tests on real data to other packages is the right balance: it keeps the size of this code repo small, and using simulated data means we can isolate the features being tested, and we know the exact right answer.

tests/test_overscanCorrection.py
PaulPrice: `maskedImage` includes the overscan, so if I were to operate on that I would be subtracting the overscan twice, which breaks the checks.

PR #54, opened 2018-04-25
DM-14204: Modernize python in ip_* packages

r-owen:


tests/test_crosstalk.py
```diff
@@ -175,8 +175,8 @@ def testTaskAPI(self):
         """
 
         class NullIsrTask(IsrTask):
-            def runDataRef(_self, dataRef):
-                return Struct(exposure=self.exposure)
+            def runDataRef(self, dataRef, exposure=self.exposure):
+                return Struct(exposure=exposure)
```
natelust: I get here that self needs fixed. However, why did you move exposure to the arguments list? Even if this is needed in the arguments, I don't see how this syntax is correct (baring some odd variable scoping). The self variable is not bound until entering the function body, before that it is just part of the parameter pack. Therefore exposure=self.exposure should not even be able to be parsed by python. If it can be then the "self" in exposure=self.exposure must be part of some clouser or outer class level scope, and is not necessarily the "self" that is the first argument to the function. If exposure is not needed in the call signature the function behavior should go back to what it was. If it is needed the function should be something like:
```python
def runDataRef(self, dataRef, exposure=None):
    if exposure is None:
        exposure = self.exposure
    return Struct(exposure=exposure)
```

r-owen: The `self` in exposure=self.exposure` is the `self` of the class in which this class is being defined. This simply binds exposure. That works, but I agree it's confusing to read. I'll try to figure out something less confusing.

PR #56, opened 2018-07-17
Add support for the new brighter-fatter kernels

mfisherlevine:
bfKernelNew is a temporary name for this dataset
while the crossover between old and new is handled
This is largely irrelevant though, as this code has
never fired before, as only obs_subaru actually
brighter-fatter correction, and it has its own isr
task. This will be updated once this dataset is
renamed.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -248,6 +248,14 @@ class IsrTaskConfig(pexConfig.Config):
         default=False,
         doc="Apply the brighter fatter correction"
     )
+    brighterFatterLevel = pexConfig.ChoiceField(
+        doc="The level at which to correct for brighter-fatter",
+        dtype=str, default="DETECTOR",
+        allowed={
+            "AMP": "Every amplifier treated separately",
+            "DETECTOR": "One kernel per detector",
+        }
+    )
     brighterFatterKernelFile = pexConfig.Field(
         dtype=str,
         default='',
@@ -493,7 +501,8 @@ class IsrTask(pipeBase.CmdLineTask):
         @param[in] defects  list of detects
         @param[in] fringes  a pipeBase.Struct with field fringes containing
                             exposure of fringe frame or list of fringe exposure
-        @param[in] bfKernel  kernel for brighter-fatter correction
+        @param[in] bfKernel  kernel for brighter-fatter correction, an
+                             lsst.cp.pipe.makeBrighterFatterKernel.BrighterFatterKernel object
         @param[in] camera  camera geometry, an lsst.afw.cameraGeom.Camera;
                            used by addDistortionModel
         @param[in] opticsTransmission  a TransmissionCurve for the optics
@@ -585,7 +594,12 @@ class IsrTask(pipeBase.CmdLineTask):
                 self.maskAndInterpNan(ccdExposure)
                 interpolationDone = True
 
-            self.brighterFatterCorrection(ccdExposure, bfKernel,
+            if self.config.brighterFatterLevel == 'DETECTOR':
+                kernelElement = bfKernel.kernel[ccdExposure.getDetector().getId()]
+            else:
+                # TODO: DM-15631 for implementing this
+                raise NotImplementedError("per-amplifier brighter-fatter correction not yet implemented")
+            self.brighterFatterCorrection(ccdExposure, kernelElement,
                                           self.config.brighterFatterMaxIter,
                                           self.config.brighterFatterThreshold,
                                           self.config.brighterFatterApplyGain,


python/lsst/ip/isr/isrTask.py
```diff
@@ -417,7 +425,7 @@ def readIsrData(self, dataRef, rawExposure):
             if self.config.doDark else None
         flatExposure = self.getIsrExposure(dataRef, self.config.flatDataProductName) \
             if self.config.doFlat else None
-        brighterFatterKernel = dataRef.get("brighterFatterKernel") if self.config.doBrighterFatter else None
+        brighterFatterKernel = dataRef.get("bfKernelNew") if self.config.doBrighterFatter else None
```
r-owen: I thought you decided to use "brighterFatterKernel"

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,6 +582,10 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
                 self.maskAndInterpNan(ccdExposure)
                 interpolationDone = True
 
+            if self.config.brighterFatterLevel == 'CCD':
```
r-owen: `CCD` -> `DETECTOR`

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,6 +582,10 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
...
+                bfKernel = bfKernel[ccdExposure.getDetector().getId()]
+            else:
+                raise NotImplementedError("per-amplifier brighter-fatter correction not yet implemented")
```
r-owen: Is this outdated? Your new code in `cp_pipe` does support per-amplifier brighter-fatter correction.

python/lsst/ip/isr/isrTask.py
mfisherlevine: Indeed, this was just stale, thanks.

python/lsst/ip/isr/isrTask.py
mfisherlevine: Fixed here, and in the config option default and doc.

python/lsst/ip/isr/isrTask.py
mfisherlevine: It's actually not outdated, no. The code there generates kernels at the per-amplifier level, however this is the correction code, and nothing has yet been written to support doing the _correction_ at the per-amplifier level. If it looked trivial I would have chucked it in with this ticket, but because the detector is already assembled at this point in the code it's not quite that easy. I will file a ticket saying that this needs to be added at some point though.

python/lsst/ip/isr/isrTask.py
```diff
@@ -585,6 +594,11 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
                 self.maskAndInterpNan(ccdExposure)
                 interpolationDone = True
 
+            if self.config.brighterFatterLevel == 'DETECTOR':
+                bfKernel = bfKernel[ccdExposure.getDetector().getId()]
```
r-owen: You are replacing `bfKernel` with an element from the variable of the same name. I think this would be a lot less confusing if you used a different variable name for the sequence than for the element of the sequence, e.g. `kernelElement = brighterFatterKernel[...]`

It would also be less confusing if the dataset type name was not singular (e.g. `brighterFatterKernels`) since you are getting a kernel from it. But I suspect it's too late to reasonably do anything about that.

r-owen: Also, as a minor point: since "bf" is expanded almost everywhere else, the code would be a bit easier to read if you used "brighterFatter" for the kernel variable as well -- especially as it matches a dataset type.

Of course it gets long if you need two names for the kernel vs. the collection of kernels...perhaps just leave out "bf" for the former, e.g. `detectorKernel = brighterFatterKernel[...]`

python/lsst/ip/isr/isrTask.py
```diff
@@ -585,6 +594,11 @@ def run(self, ccdExposure, bias=None, linearizer=None, dark=None, flat=None, def
...
+            else:
+                # TODO: DM-15631 for implementing this
+                raise NotImplementedError("per-amplifier brighter-fatter correction not yet implemented")
```
r-owen: Do you anticipate being able to call `self.brighterFatterCorrection` in a single pass for the per-amplifier case? If not, I suggest putting the following into the "if" above.

python/lsst/ip/isr/isrTask.py
mfisherlevine: Oh good lord, that's a great spot there, overwriting the object with an element of it, that was totally unintended! I have changed that now.

Regarding the rest, I take your point, but sadly the name `bfKernel` is passed in as a kwarg so changing that is RFC territory (in reality I'm sure it's fine, but _technically_ I think it's required?). So as you say, I don't think it's worth changing now, especially as all of isr will need a cleanup at some point.

PR #59, opened 2018-07-26
DM-2639: {{Standardize primary method names, run/runDataRef, across PipeTasks}}

czwa:
Rename methods according to RFC-352.
Fifteen total repositories with changes.

### DIFF
--- a/python/lsst/ip/isr/measureCrosstalk.py
+++ b/python/lsst/ip/isr/measureCrosstalk.py
@@ -195,7 +195,7 @@ class MeasureCrosstalkTask(CmdLineTask):
             pickle.dump(resultList, open(results.parsedCmd.dumpRatios, "w"))
         return task.reduce(resultList)
 
-    def run(self, dataRef):
+    def runDataRef(self, dataRef):
         """Get crosstalk ratios for CCD
 
         Parameters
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -195,7 +195,7 @@ def parseAndRun(cls, *args, **kwargs):
             pickle.dump(resultList, open(results.parsedCmd.dumpRatios, "w"))
         return task.reduce(resultList)
 
-    def run(self, dataRef):
+    def runDataRef(self, dataRef):
```
r-owen: I think this is a case where `runDataRef` should call `run` since `run` can do useful work on the unpersisted exposure.

natelust: The work of converting tasks to have the work done in run, and the runDataRef be the entry point for command line task is planned for after this ticket is complete. We chose to do the big outward facing api change first, and save any refactoring work for another ticket.

r-owen: This would be a trivial change. Most of the other renames do not lend themselves to adding `run` methods (at least that I saw in the packages I looked at) but this one did.

czwa: In the interest in merging the changes for this ticket that refactoring depends on, @yalsayyad  has helped by opening a ticket on this refactorization.  DM-15311 covers this change, and I'll begin work on it next week.

PR #62, opened 2018-08-21
DM-15311: {{Refactor MeasureCrosstalkTask.runDataRef to new run.}}

czwa:
Small refactor of MeasureCrosstalkTask to move the calculations to a run method.


czwa:
Small change to remove the dataRef from the run() method entirely, with only the dataId being passed.

PR #63, opened 2018-08-29
Tickets/dm 15203

RobertLuptonTheGood:


python/lsst/ip/isr/isrTask.py
```diff
@@ -172,7 +173,7 @@ class IsrTaskConfig(pexConfig.Config):
              "or number of spline knots if overscan fit type is a spline."),
         default=1,
     )
-    overscanRej = pexConfig.Field(
+    overscanNumSigmaClip = pexConfig.Field(
```
PaulPrice: I think this change and the addition of `MEANCLIP` should be in their own commits, not hidden in a "modernisation" commit.

python/lsst/ip/isr/isrTask.py
```diff
@@ -937,7 +939,7 @@ def overscanCorrection(self, exposure, amp):
             overscanImage=overscanImage,
             fitType=self.config.overscanFitType,
             order=self.config.overscanOrder,
-            collapseRej=self.config.overscanRej,
+            statControl=sctrl,
```
PaulPrice: I think this change belongs in the next commit.

python/lsst/ip/isr/isrTask.py
```diff
@@ -934,6 +934,9 @@ def overscanCorrection(self, exposure, amp):
         dataView = maskedImage.Factory(maskedImage, amp.getRawDataBBox())
         overscanImage = maskedImage.Factory(maskedImage, amp.getRawHorizontalOverscanBBox())
 
+        sctrl = afwMath.StatisticsControl()
+        sctrl.setNumSigmaClip(self.config.overscanNumSigmaClip)
```
PaulPrice: It's not clear to me why passing a `StatisticsControl` object into `overscanCorrection` is better than passing in a rejection threshold: it doesn't add any information (or are you thinking to expand its use?), so the rejection threshold is more fundamental. It's like passing a `Struct` consisting of a single value.
Convince me?

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -336,10 +336,23 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
 
-    if fitType in ('MEAN', 'MEANCLIP', 'MEDIAN'):
+    if fitType in ('MEAN', 'MEANCLIP'):
         fitType = afwMath.stringToStatisticsProperty(fitType)
         offImage = afwMath.makeStatistics(overscanImage, fitType, statControl).getValue()
         overscanFit = offImage
+    if fitType in ('MEDIAN',):
+        # we need an image with integer pixels to handle ties properly
+        if hasattr(overscanImage, "image"):
+            imageI = overscanImage.image.convertI()
```
PaulPrice: Is it worth including a check that the data is indeed integer that has been converted to floating-point? Might someone want to run an overscan correction on an image that has floating-point values (maybe some other operation has been performed before the overscan correction)?

python/lsst/ip/isr/isrTask.py
RobertLuptonTheGood: I think that the API in `isrFunctions` should support the `StatisticsControl` as the user can set other options (e.g. masks to ignore);  indeed, we could then add them to `IsrTask` without modifying the API

python/lsst/ip/isr/isrFunctions.py
RobertLuptonTheGood: I'll add an argument (default True) to allow the conversion to int.  But I don't think anyone would/should ever want to run overscan estimation on processed data.

python/lsst/ip/isr/isrTask.py
RobertLuptonTheGood: I'll restructure the commits.

PR #64, opened 2018-08-29
Add config params to control which overscan columns to use

RobertLuptonTheGood:


python/lsst/ip/isr/isrTask.py
```diff
@@ -928,9 +939,25 @@ def overscanCorrection(self, exposure, amp):
...
+        prescanBBox = amp.getRawPrescanBBox()
+        if oscanBBox.getBeginX() > prescanBBox.getBeginX():  # amp is at the right
+            x0 += self.config.overscanNumLeadingColumnsToSkip
+            x1 -= self.config.overscanNumTrailingColumnsToSkip
+        else:
+            x0 += self.config.overscanNumTrailingColumnsToSkip
+            x1 -= self.config.overscanNumLeadingColumnsToSkip
+
+        oscanBBox = afwGeom.BoxI(afwGeom.PointI(x0, oscanBBox.getBeginY()),
+                                 afwGeom.PointI(x1 - 1, oscanBBox.getEndY() - 1))
```
mfisherlevine: I haven't played with this stuff to know if the `-1`s here are right. As long as you're sure that's correct then I think this looks good.

RobertLuptonTheGood: This is correct; integer BBoxes are inclusive (otherwise integer and float ones are too different).  It's a bit confusing, but Jim and I convinced ourselves that it was the least confusing option

PR #65, opened 2018-09-05
DM-6640: IsrTask is not a valid CmdLineTask

czwa:
This ticket adds a stand-alone runIsr.py script, as well as ensuring that runIsr.py and processCcd.py read the same ISR configuration files.


python/lsst/ip/isr/isrTask.py
```diff
@@ -1225,3 +1225,50 @@ def getSaturation(self):
...
+## @copybrief IsrWrapperTask
+## @}
+
+
+class IsrWrapperTask(pipeBase.CmdLineTask):
+    """!
+    @anchor IsrWrapperTask_
+
+    @brief Task to wrap the main IsrTask to allow camera specific retargeting.
+    """
```
parejkoj: Please make the new docstrings numpydoc, otherwise we'll just have to convert them later.

Also, can you please expand on this documentation a bit? I don't really understand what makes it different from the original `IsrTask`.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1225,3 +1225,50 @@ def getSaturation(self):
...
+    _DefaultName = "isrWrap"
+
+    def __init__(self, *args, **kwargs):
+        """!Constructor for IsrWrapperTask
+        @param[in] *args    a list of positional arguments passed on to the Task
+                            constructor
+        @param[in] *kwargs  a dictionary of keyword arguments passwed on to the
+                            Task constructor
+        Call the lsst.pipe.base.task.Task.__init__ method,
+        then setup the default IsrTask as a subtask
```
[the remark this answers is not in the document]
mrawls: Agree with earlier comments re. numpydoc, also please note our guide explicitly says not to have a docstring for __init__ methods. More details at https://developer.lsst.io/python/numpydoc.html#py-docstring-class-structure

bin.src/runIsr.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008, 2009, 2010 LSST Corporation.
```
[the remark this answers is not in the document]
yalsayyad: We do generally have copylefts in our scripts in bin.src/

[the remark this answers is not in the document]
timj: We should put the standard boiler plate (which changed recently so check in https://developer.lsst.io/stack/license-and-copyright.html?highlight=copyright ) in each source file. Copyrights should be in a COPYRIGHT file though, not each file.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1225,3 +1225,50 @@ def getSaturation(self):
 
     def getSuspectLevel(self):
         return float("NaN")
+
+
+class IsrWrapperTaskConfig(pexConfig.Config):
```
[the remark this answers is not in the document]
yalsayyad: And double check that you update the config file name to match this in obs_base too

bin.src/runIsr.py
```diff
@@ -0,0 +1,25 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.    See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+from lsst.ip.isr.isrTask import IsrWrapperTask
+
+IsrWrapperTask.parseAndRun()
```
yalsayyad: Might just be me, but I would expect this to tun `RunIsrTask` given the name of the script is `runIsr`.    Remember all the eye-rolling over  `bin/measureCoaddSources.py` calling 
`MeasureMergedCoaddSourcesTask.parseAndRun()` and the like in pipe_tasks

python/lsst/ip/isr/isrTask.py
czwa: Docstrings updated to numpydoc.  I didn't use a sufficiently new comparison to get the new format.

Docstring also amended to include information on what this change does, and why it's needed.

bin.src/runIsr.py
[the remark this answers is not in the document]
czwa: I found the updated boiler plate today while looking, so that has been updated as well.

bin.src/runIsr.py
czwa: I am open to different solutions, but I don't see one that doesn't involve either 1) making separate runIsr_obsX.py scripts for each obs_package; or 2) making the current `RunIsrTask` into `RunIsrRealTask` or something equivalent.

yalsayyad: Is there already a `RunIsrTask`???

yalsayyad: The pattern now should be `IsrTask` + camera-specific IsrTask's like `DecamNullIsrTask`. To my knowledge, there is no `RunIsrTask`.   I'm going offline for the night. I don't care enough for you to change it to `RunIsrTask`,  but be aware one of the PipelineTask precursor/refactoring jobs that @natelust going to work on is making the multiband script names match their Task names in pipe_tasks.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1225,3 +1225,50 @@ def getSaturation(self):
...
+    methods, this would remove those data-specific methods in the
+    output post-ISR images.  This wrapping class fixes the issue,
+    allowing identical post-ISR images to be generated by both the
+    processCcd and isrTask code.
+    """
+    ConfigClass = IsrWrapperConfig
+    _DefaultName = "isrWrap"
+
+    def __init__(self, *args, **kwargs):
+        pipeBase.Task.__init__(self, *args, **kwargs)
```
TallJimbo: I think @natelust is or shortly will be going through the `Tasks` and making them use `super()` in places like this, so we can use multiple inheritance if we need to during the `CmdLineTask`->`PipelineTask` transition.  Might be good to go ahead and do that in new code right now.

PR #69, opened 2018-11-29
DM-15862: Reduce ISR code duplication between ip_isr, obs_subaru, and obs_decam

czwa:


SimonKrughoff:
As a general comment, it is helpful if the summary line for each commit gives some info about what the commit is doing.  Having three commits with identical summaries makes it hard to tell which is which.  I really like the detail in the body of the commit BTW.

python/lsst/ip/isr/isrTask.py
```diff
@@ -189,71 +188,123 @@ class IsrTaskConfig(pexConfig.Config):
...
+        default=False,
+        doc="Calculate empirical read noise instead of value from AmpInfo data?"
+    )
+
     doLinearize = pexConfig.Field(
         dtype=bool,
         doc="Correct for nonlinearity of the detector's response?",
         default=True,
     )
+
```
SimonKrughoff: Same comment as above.  Looking at this, it seems some configs are being logically grouped, but if that's the case, it would be kind of nice to have a comment saying what the groups are.  This is optional, I guess.

python/lsst/ip/isr/isrTask.py
```diff
@@ -469,220 +642,502 @@ def readIsrData(self, dataRef, rawExposure):
...
 
         if self.config.doAssembleCcd:
+            self.log.info("Assembling CCD from amplifiers")
             ccdExposure = self.assembleCcd.assembleCcd(ccdExposure)
+
             if self.config.expectWcs and not ccdExposure.getWcs():
                 self.log.warn("No WCS found in input exposure")
+            self.debugView(ccdExposure, "doAssembleCcd")
+
+        # CZW: push these into output struct, add to docstring
```
SimonKrughoff: I don't understand this note.  If there is no config to do the Oss writing, shouldn't this just go away?

python/lsst/ip/isr/isrTask.py
```diff
@@ -469,220 +642,502 @@ def readIsrData(self, dataRef, rawExposure):
...
+
                 if self.config.doSaturationInterpolation:
                     self.saturationInterpolation(ccdExposure)
+
                 self.maskAndInterpNan(ccdExposure)
                 interpolationDone = True
 
             if self.config.brighterFatterLevel == 'DETECTOR':
-                kernelElement = bfKernel.kernel[ccdExposure.getDetector().getId()]
+                kernelElement = bfKernel  # [ccdExposure.getDetector().getId()]
```
SimonKrughoff: remove commented out code.

python/lsst/ip/isr/isrTask.py
```diff
@@ -469,220 +642,502 @@ def readIsrData(self, dataRef, rawExposure):
...
 
-        exposureTime = ccdExposure.getInfo().getVisitInfo().getExposureTime()
-        ccdExposure.getCalib().setFluxMag0(self.config.fluxMag0T1*exposureTime)
+        if self.config.doSetBadRegions:
+            badPixelCount, badPixelValue = isrFunctions.setBadRegions(ccdExposure)
+            self.log.info("Set %d BAD pixels to %f." % (badPixelCount, badPixelValue))
 
-        if self.config.doAddDistortionModel:
-            self.addDistortionModel(exposure=ccdExposure, camera=camera)
+        # CZW: add this to the result struct
```
SimonKrughoff: I don't understand this either.  Can't you just uncomment this?

python/lsst/ip/isr/isrTask.py
```diff
@@ -696,44 +1151,222 @@ def convertIntToFloat(self, exposure):
...
+        # Construct views
+        ampImage = afwImage.MaskedImageF(ccdExposure.getMaskedImage(), amp.getRawDataBBox(),
+                                         afwImage.PARENT)
+        overscanImage = afwImage.MaskedImageF(ccdExposure.getMaskedImage(),
+                                              amp.getRawHorizontalOverscanBBox(),
+                                              afwImage.PARENT)
+        overscanArray = overscanImage.getImage().getArray()
+
+        statControl = afwMath.StatisticsControl()
+        statControl.setAndMask(ccdExposure.getMaskedImage().getMask().getPlaneBitMask("SAT"))
```
SimonKrughoff: Is this just a rendering error?  It seems like this line is indented far too far.

SimonKrughoff: Oh never mind.  My browser window was too narrow.

python/lsst/ip/isr/isrQa.py
```diff
@@ -75,7 +75,7 @@ class IsrQaConfig(pexConfig.Config):
     doThumbnailOss = pexConfig.Field(
         dtype=bool,
         doc="Write overscan subtracted thumbnail?",
-        default=True,
+        default=False,
```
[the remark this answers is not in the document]
czwa: This was triggering obs_ctio0m9 issues.  The update to obs_base likely resolved this, so I will do a test with default=True, and see if the error goes away.

python/lsst/ip/isr/isrTask.py
czwa: I've added comments to explain the groups.  I needed some way to organize the configuration options for myself, given the large number of entries.  The previous ordering was not clear, and this will hopefully minimize future searching.

PR #70, opened 2018-12-20
tickets/DM-16813

plazas:
Check input in isrTask to raise if doCrosstalk is true and crosstalk coefficients are not found


czwa:
No issues with the code.  I suggest updating the commit message to match the "best practices" guide to keep the log tidy: https://developer.lsst.io/work/flow.html#git-commit-message-best-practices

laurenam:
It also looks like the two commits should be squashed.

PR #72, opened 2019-01-18
Convert isrTask to valid pipelineTask.

czwa:
Additional configuration options now exist to define the required
input/output dataset types, and appropriate handlers correctly remove
these datasets if the isr configuration will not require (for inputs)
or generate (for outputs) those products.  As the camera configuration
is not currently fully populated, the obs_subaru/hsc configuration is
read to set the correct parameters.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -38,9 +38,11 @@ from lsst.afw.display import getDisplay
 from lsst.afw.geom import Polygon
 from lsst.daf.persistence import ButlerDataRef
 from lsst.meas.algorithms.detection import SourceDetectionTask
+from lsst.meas.algorithms import Defect
 
 from . import isrFunctions
 from . import isrQa
+from . import linearize
 
 from .assembleCcdTask import AssembleCcdTask
 from .crosstalk import CrosstalkTask
@@ -59,16 +61,133 @@ class IsrTaskConfig(pexConfig.Config):
     Items are grouped in the order in which they are executed by the task.
     """
     # General ISR configuration
+
+    # gen3 options
+    isrName = pexConfig.Field(
+        dtype=str,
+        doc="Name of ISR",
+        default="ISR",
+    )
+
+    # input datasets
+    ccdExposure = pipeBase.InputDatasetField(
+        doc="Input exposure to process",
+        name="raw",
+        scalar=True,
+        storageClass="ExposureU",
+        dimensions=["Instrument", "Exposure", "Detector"],
+    )
+    camera = pipeBase.InputDatasetField(
+        doc="Input camera to construct complete exposures.",
+        name="camera",
+        scalar=True,
+        storageClass="TablePersistableCamera",
+        dimensions=["Instrument", "ExposureRange"],
+    )
+    bias = pipeBase.InputDatasetField(
+        doc="Input bias calibration.",
+        name="bias",
+        scalar=True,
+        storageClass="ImageF",
+        dimensions=["Instrument", "ExposureRange", "Detector"],
+    )
+    dark = pipeBase.InputDatasetField(
+        doc="Input dark calibration.",
+        name="dark",
+        scalar=True,
+        storageClass="ImageF",
+        dimensions=["Instrument", "ExposureRange", "Detector"],
+    )
+    flat = pipeBase.InputDatasetField(
+        doc="Input flat calibration.",
+        name="flat",
+        scalar=True,
+        storageClass="MaskedImageF",
+        dimensions=["Instrument", "PhysicalFilter", "ExposureRange", "Detector"],
+    )
+    bfKernel = pipeBase.InputDatasetField(
+        doc="Input brighter-fatter kernel.",
+        name="bfKernel",
+        scalar=True,
+        storageClass="NumpyArray",
+        dimensions=["Instrument", "ExposureRange"],
+    )
+    defects = pipeBase.InputDatasetField(
+        doc="Input defect tables.",
+        name="defects",
+        scalar=True,
+        storageClass="Catalog",
+        dimensions=["Instrument", "ExposureRange", "Detector"],
+    )
+    opticsTransmission = pipeBase.InputDatasetField(
+        doc="Transmission curve due to the optics.",
+        name="transmission_optics",
+        scalar=True,
+        storageClass="TablePersistableTransmissionCurve",
+        dimensions=["Instrument", "ExposureRange"],
+    )
+    filterTransmission = pipeBase.InputDatasetField(
+        doc="Transmission curve due to the filter.",
+        name="transmission_filter",
+        scalar=True,
+        storageClass="TablePersistableTransmissionCurve",
+        dimensions=["Instrument", "PhysicalFilter", "ExposureRange"],
+    )
+    sensorTransmission = pipeBase.InputDatasetField(
+        doc="Transmission curve due to the sensor.",
+        name="transmission_sensor",
+        scalar=True,
+        storageClass="TablePersistableTransmissionCurve",
+        dimensions=["Instrument", "ExposureRange", "Detector"],
+    )
+    atmosphereTransmission = pipeBase.InputDatasetField(
+        doc="Transmission curve due to the atmosphere.",
+        name="transmission_atmosphere",
+        scalar=True,
+        storageClass="TablePersistableTransmissionCurve",
+        dimensions=["Instrument"],
+    )
+
+    # output datasets
+    outputExposure = pipeBase.OutputDatasetField(
+        doc="Output ISR processed exposure.",
+        name="postISRCCD",
+        scalar=True,
+        storageClass="ExposureF",
+        dimensions=["Instrument", "Visit", "Detector"],
+    )
+    outputOssThumbnail = pipeBase.OutputDatasetField(
+        doc="Output Overscan-subtracted thumbnail image.",
+        name="OssThumb",
+        scalar=True,
+        storageClass="Thumbnail",
+        dimensions=["Instrument", "Visit", "Detector"],
+    )
+    outputFlattenedThumbnail = pipeBase.OutputDatasetField(
+        doc="Output flat-corrected thumbnail image.",
+        name="FlattenedThumb",
+        scalar=True,
+        storageClass="TextStorage",
+        dimensions=["Instrument", "Visit", "Detector"],
+    )
+
+    quantum = pipeBase.QuantumConfig(
+        dimensions=["Visit", "Detector", "Instrument"],
+    )
+
+    ## Original gen2 config
     datasetType = pexConfig.Field(
         dtype=str,
         doc="Dataset type for input data; users will typically leave this alone, "
         "but camera-specific ISR tasks will override it",
         default="raw",
     )
+
     fallbackFilterName = pexConfig.Field(
         dtype=str,
         doc="Fallback default filter name for calibrations.",
-        optional=True)
+        optional=True
+    )
     expectWcs = pexConfig.Field(
         dtype=bool,
         default=True,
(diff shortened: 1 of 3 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -574,16 +713,117 @@ class IsrTask(pipeBase.CmdLineTask):
...
+            del inputTypeDict["sensorTransmission"]
+        if config.doUseAtmosphereTransmission is not True:
+            del inputTypeDict["atmosphereTransmission"]
+
+        return inputTypeDict
+
+    @classmethod
+    def getOutputDatasetTypes(cls, config):
+        outputTypeDict = super().getOutputDatasetTypes(config)
+        obsConfigDir = os.path.join(getPackageDir("obs_subaru"), "config", "hsc")
```
SimonKrughoff: Same concern as above with hard coding config overrides

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,16 +713,117 @@ class IsrTask(pipeBase.CmdLineTask):
...
+        config.load(os.path.join(obsConfigDir, "isr.py"))
+
+        if config.qa.doThumbnailOss is not True:
+            del outputTypeDict["outputOssThumbnail"]
+        if config.qa.doThumbnailFlattened is not True:
+            del outputTypeDict["outputFlattenedThumbnail"]
```
SimonKrughoff: What about `postIsrCcd`?

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,16 +713,117 @@ class IsrTask(pipeBase.CmdLineTask):
...
+
+        return outputTypeDict
+
+    def adaptArgsAndRun(self, inputData, inputDataIds, outputDataIds, butler):
+        try:
+            inputData['detectorNum'] = int(inputDataIds['ccdExposure']['detector'])
+        except Exception as e:
+            raise ValueError("Failure to find valid detectorNum value for Dataset {}: {}".format(
+                inputDataIds, e))
```
SimonKrughoff: I just learned about `f""` formatted strings and they are really compact.  See a description [here](https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python).  I'm not requiring you change, but Tim introduced them to me and I thought they were so much better than previous formatting that I wanted to spread the word.

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,16 +713,117 @@ class IsrTask(pipeBase.CmdLineTask):
...
+
+        # Broken: DM-17152
+        # Fringes are not handled correctly by Gen3 butler.  We need to be able to return the
+        # fringe if the filter contains one, but not fail if the filter doesn't.
+        # inputData['fringes'] = (self.fringe.readFringes(inputDataIds['ccdExposure'],
+        #                                                 assembler=self.assembleCcd
+        #                                                 if self.config.doAssembleIsrExposures else None)
+        #                         if self.config.doFringe and
+        #                            self.fringe.checkFilter(inputData['ccdExposure'])
+        #                         else pipeBase.Struct(fringes=None))
```
SimonKrughoff: So you are saying there is some HSC specific way to do this in `obs_subaru` that will work for `ci_hsc`, but for now that won't work for any other cameras?

python/lsst/ip/isr/isrTask.py
```diff
@@ -747,13 +988,19 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
...
+            Flag this call to run() as using the Gen3 butler environment.
 
         Returns
         -------
         result : `lsst.pipe.base.Struct`
             Result struct with component:
             - ``exposure`` : `afw.image.Exposure`
                 The fully ISR corrected exposure.
+            - ``outputExposure`` : `afw.image.Exposure`
+                An alias for `exposure`
```
SimonKrughoff: Why add the alias?  Is it part of the sanitization of terminology that is coming with Gen3?

python/lsst/ip/isr/isrTask.py
```diff
@@ -788,11 +1035,28 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
...
-            return self.runDataRef(ccdExposure)
 
-        self.log.info("Performing ISR on exposure %s" %
-                      (ccdExposure.getInfo().getVisitInfo().getExposureId()))
+        if isGen3 is True:
+            # Gen3 currently cannot do configuration overrides.  Manually do so,
+            # supplying it with the obs_subaru configuration for HSC.
+            # DM-15257 looks to discuss this issue.
+
+            obsConfigDir = os.path.join(getPackageDir("obs_subaru"), "config", "hsc")
```
SimonKrughoff: Same concern as above that we are hard coding config overrides.  Thanks for including the ticket for discussion here.

python/lsst/ip/isr/isrTask.py
```diff
@@ -788,11 +1035,28 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
         exposure after all ISR processing has finished.
 
         """
-        if isinstance(ccdExposure, ButlerDataRef):
-            return self.runDataRef(ccdExposure)
 
-        self.log.info("Performing ISR on exposure %s" %
-                      (ccdExposure.getInfo().getVisitInfo().getExposureId()))
```
SimonKrughoff: This log message seems useful.  Is it being output in a different way now?

SimonKrughoff: Oh.  Never mind.  I see this is output later.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1136,6 +1405,50 @@ def getIsrExposure(self, dataRef, datasetType, immediate=True):
...
+        if isinstance(inputExp, afwImage.DecoratedImageU):
+            inputExp = afwImage.makeExposure(afwImage.makeMaskedImage(inputExp))
+        elif isinstance(inputExp, afwImage.ImageF):
+            inputExp = afwImage.makeExposure(afwImage.makeMaskedImage(inputExp))
+        elif isinstance(inputExp, afwImage.MaskedImageF):
+            inputExp = afwImage.makeExposure(inputExp)
+        elif isinstance(inputExp, afwImage.Exposure):
+            pass
+        else:
+            raise TypeError("Input Exposure is not known type in isrTask.ensureExposure!")
```
SimonKrughoff: Maybe add what type it is?

python/lsst/ip/isr/isrTask.py
```diff
@@ -1463,7 +1776,12 @@ def darkCorrection(self, exposure, darkExposure, invert=False):
         expScale = exposure.getInfo().getVisitInfo().getDarkTime()
         if math.isnan(expScale):
             raise RuntimeError("Exposure darktime is NAN")
-        darkScale = darkExposure.getInfo().getVisitInfo().getDarkTime()
+        if darkExposure.getInfo().getVisitInfo() is not None:
+            darkScale = darkExposure.getInfo().getVisitInfo().getDarkTime()
+        else:
+            # CZW: Hack
```
SimonKrughoff: What requires this hack?  Is it being fixed anywhere?

python/lsst/ip/isr/isrTask.py
```diff
@@ -59,16 +64,150 @@ class IsrTaskConfig(pexConfig.Config):
...
+        default="ISR",
+    )
+
+    # input datasets
+    ccdExposure = pipeBase.InputDatasetField(
+        doc="Input exposure to process",
+        name="raw",
+        scalar=True,
+        storageClass="ExposureU",
+        dimensions=["Instrument", "Exposure", "Detector"],
```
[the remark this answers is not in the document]
czwa: The `dimension`s can be extracted from the registry as well with a query like `select * from DatasetTypeDimensions WHERE dataset_type_name = 'raw';`.  The concept is "which dimensions are needed to uniquely specify a dataset to work on.  For raw, this set is `Instrument` (we work on one camera at a time), `Exposure` (`Visit` before; a single integration of the camera), and `Detector` (`ccd` before I think; a single chip in the camera).  Knowing all of these ahead of time isn't something that's obvious, which is why I've copied from the registry.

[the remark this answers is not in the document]
SimonKrughoff: OK, but these are now standardized (the switch from `Visit` to `Exposure` is what I was missing)?  If so, I'm less worried.

python/lsst/ip/isr/isrTask.py
```diff
@@ -59,16 +64,150 @@ class IsrTaskConfig(pexConfig.Config):
...
+
+    # gen3 options
+    isrName = pexConfig.Field(
+        dtype=str,
+        doc="Name of ISR",
+        default="ISR",
+    )
+
+    # input datasets
+    ccdExposure = pipeBase.InputDatasetField(
```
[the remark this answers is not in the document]
SimonKrughoff: And `dimensions`, of course, but that is discussed below.  I always get a little scared when someone says "There are no docs yet", but I don't think that's the responsibility of this ticket.

python/lsst/ip/isr/isrTask.py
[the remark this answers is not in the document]
czwa: I believe these are now standardized and stable for the foreseeable future.

python/lsst/ip/isr/isrTask.py
```diff
@@ -59,16 +64,150 @@ class IsrTaskConfig(pexConfig.Config):
...
+        dimensions=["Instrument", "ExposureRange"],
+    )
+    defects = pipeBase.InputDatasetField(
+        doc="Input defect tables.",
+        name="defects",
+        scalar=True,
+        storageClass="Catalog",
+        dimensions=["Instrument", "ExposureRange", "Detector"],
+    )
+    # # fringeExposure = pipeBase.InputDatasetField(
```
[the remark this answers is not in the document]
czwa: I suspect fringes actually would work now, but that isn't tested, and isn't needed on this ticket now, so I'm removing it.

python/lsst/ip/isr/isrTask.py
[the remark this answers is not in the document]
TallJimbo: Standardized: yes.  Stabilized: not to the extent that I'd like, but that's recognized as a high priority.

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,16 +713,117 @@ class IsrTask(pipeBase.CmdLineTask):
...
         self.makeSubtask("crosstalk")
         self.makeSubtask("strayLight")
         self.makeSubtask("fringe")
         self.makeSubtask("masking")
         self.makeSubtask("vignette")
 
+    @classmethod
+    def getInputDatasetTypes(cls, config):
+        inputTypeDict = super().getInputDatasetTypes(config)
+        obsConfigDir = os.path.join(getPackageDir("obs_subaru"), "config", "hsc")
```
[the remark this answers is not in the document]
TallJimbo: (as mention in local discussion) I would prefer to move this workaround to the command-line (i.e. `-C $OBS_SUBARU_DIR/config/hsc/isr.py`), even if that makes running the task a little less convenient right now.

python/lsst/ip/isr/isrTask.py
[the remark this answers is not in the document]
SimonKrughoff: Good enough.

python/lsst/ip/isr/isrTask.py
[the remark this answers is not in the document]
czwa: After the discussion here, I've tested that I can do that, and so these `obsConfigDir` snippets will be removed.

python/lsst/ip/isr/isrTask.py
czwa: My assumption was that if you didn't expect an `outputExposure`, you wouldn't get a valid graph, but that's not necessarily true.  For completeness, I've added the option.

python/lsst/ip/isr/isrTask.py
czwa: I've changed the comment a bi.  Gen3 is currently only testable with `ci_hsc`, but `ci_hsc` has no y-band data to test that fringes load correctly.  A block like that used for the defects may be needed, or this all will get wrapped into a new butler formatter that will handle repacking data structures.

python/lsst/ip/isr/isrTask.py
czwa: I aliased it to keep the gen2/gen3 elements of the struct independent.  Hopefully this keeps testing, debugging, and gen2 deprecation simpler.

PR #81, opened 2019-03-06
tickets/DM-1448: Move code for mock images so it reusable.

czwa:
The examples in the ip_isr/examples directory have been superseded.  The isr.py and runAssembleTask.py scripts produce results that can be obtained from the runIsr.py command line task.  Some of the exampleUtils.py code to generate fake image data will be included in the new isrMock.py code.

The added IsrMock code is based on the afw/cameraGeom/testUtils.py and afw/cameraGeom/utils.py code to provide a complete camera geometry/detector/amplifier set, and replaces the example code from ip_isr/examples.  The new mocks can be constructed to either represent a "raw" exposure with the inclusion of various instrument signals (bias, dark, flat, etc.), as well as the master calibration frame to allow the removal of those instrument signals.  A Gen-2 butler dataRef can also be constructed, to test that IsrTask properly retrieves the appropriate mock data.


python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+            for y in range(0, ampData.getDimensions().getY()):
+                ampData.getArray()[y][x] = (ampData.getArray()[y][x] +
+                                            scale * np.exp(-0.5 * ((x - 50)**2 + (y - 25)**2)**2 / 3.0**2))
+
+    def amplifierAddCT(self, ampDataSource, ampDataTarget, scale):
+        r"""Add a scaled copy of an amplifier to another, simulating crosstalk.
+
+         This method operates in the amplifier coordinate frame.
+
+        Parameters
```
mfisherlevine: I think both of these were copied & pasted from elsewhere 😄

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        ampData : `lsst.afw.image.ImageF`
+            Amplifier image to operate on.
+        scale : float
+            Peak flux of the source to add.
+
+        Notes
+        -----
+        This simulates simple crosstalk between amplifiers.
+        """
+        for x in range(0, ampDataSource.getDimensions().getX()):
```
mfisherlevine: Surely this is doable with a one-liner, _i.e._ not looping over pixels?

mfisherlevine: Something like
`ampDataTarget.array[:] +=  scale * ampDataSource.array[:]`
replacing the whole loop? Or am I not thinking about something here?

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+            ampData = exposure.image[bbox]
+
+            if self.config.doSky is True:
+                self.amplifierAddNoise(ampData, self.config.skyLevel, self.config.skySigma)
+            if self.config.doSource is True:
+                if idx == 0:
+                    self.amplifierAddSource(ampData, self.config.sourceFlux)
+
+            if self.config.doFringe is True:
+                self.amplifierAddFringe(amp, ampData, self.config.fringeScale)
```
mfisherlevine: I think you need to check here that it's is trimmed, because otherwise you'll add fringing to the overscan, and that's not right. Am I thinking correctly here?

Or perhaps, better than asserting that it's trimmed, add logic to only add the fringe to the data region regardless, that would be much better.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+
+        Notes
+        -----
+        This uses an offset sinc function to generate a ripple
+        pattern.  True fringes have much finer structure, but this
+        pattern should be visually identifiable.  The (x, y)
+        coordinates are in the frame of the amplifier, and (u, v) in
+        the frame of the full trimmed image.
+        """
+        for x in range(0, ampData.getDimensions().getX()):
```
mfisherlevine: I guess I'll let you off the pixel loop here, because I can't be bothered to work out the alternative because it's gone 1am here. However, there are some other problems. 

The sinc function rings down very fast, and the `/50` isn't really enough to keep it alive over much of the chip at all, given it goes 4k in both directions for LSST. Perhaps something a lot larger would be better? But I do realise that will look very silly near the center of the sinc though...

Also though, while I do see the similarity between `sinc` and fringes, perhaps there are better options for this. Might a simple 2d sin wave look better?

mfisherlevine: Also, again, center of the fringe is very close to chip edge. Moving it in might help things a bit too.

mfisherlevine: Oh, just noticed - it's moved in on one axis, but is literally on the edge on the other? Shouldn't something be being subtracted from `v` as well as `u`? Or did you really want the middle of the fringe on the literal edge?

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+            Amplifier to operate on.  Needed for amp<->exp coordinate transforms.
+        ampData : `lsst.afw.image.ImageF`
+            Amplifier image to operate on.
+        size : float
+            Size of the dome to use.  A larger value results in a
+            smaller center-to-edge fall off.
+
+        Notes
+        -----
+        This uses a centered 2-d Gaussian to simulate an illumination
```
mfisherlevine: I'm not sure how this is "centered" as you say. It seems like it would be 100 pixels from each edge, but there's no guarantee that that's the center.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        pattern that falls off towards the edge of the detector.  The
+        e(x, y) coordinates are in the frame of the amplifier, and (u, v)
+        in the frame of the full trimmed image.
+        """
+        for x in range(0, ampData.getDimensions().getX()):
+            for y in range(0, ampData.getDimensions().getY()):
+                (u, v) = self.localCoordToExpCoord(amp, x, y)
+                f = np.exp(-0.5 * ((u - 100)**2 + (v - 100)**2) / size**2)
+                ampData.getArray()[y][x] = (ampData.getArray()[y][x] * f)
```
mfisherlevine: The function name, and the docstring say "add", but this is a multiplicative effect. I see how this is adding an effect, so I don't disagree, but given that a lot of things in here are actually being added (all the noises, the dark current etc), I think it would be good to be clear.

Also, I think the function should probably be renamed to have `dome` in it somewhere, lest it get confused with normal flat stuff.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        self.config.isTrimmed = True
+        self.config.doOverscan = False
+        self.config.doBias = True
+        self.config.doDark = False
+        self.config.doFlat = False
+        self.config.doFringe = False
+        self.config.doCrosstalk = True
+        self.config.biasLevel = 0.0
+        self.config.biasSigma = 0.0
+        self.config.skySigma = 100.0
```
mfisherlevine: Right, this is exactly what I feared - changing the `skySigma` independent of the `skyLevel`. Does this makes sense?

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        self.config.doFringe = False
+
+
+class BiasMock(MasterMock):
+    r"""Simulated master bias calibration.
+    """
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.config.doBias = True
+        self.config.biasSigma = 30.0
```
mfisherlevine: Everything is relative, but this still seems like a very large number for a master calib. LSST sensors have a readnoise under 10e, and that's without beating that down with numbers, as with masters.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+    Currently only supports get and put operations, which are most
+    likely to be called for data in ISR processing.
+
+    """
+    dataId = "isrMock Fake Data"
+    darkval = 2.  # e-/sec
+    oscan = 250.  # DN
+    gradient = .10
+    exptime = 15  # seconds
+    darkexptime = 40.  # seconds
```
mfisherlevine: Not that it matters, but that's a super long darkTime for a 15s expTime!

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        elif dataType == 'defects':
+            return DefectMock(config=self.config).run()
+        elif dataType == 'bfKernel':
+            return BfKernelMock(config=self.config).run()
+        elif dataType == 'linearizer':
+            return None
+        elif dataType == 'crosstalkSources':
+            return None
+        else:
+            return None
```
mfisherlevine: OK, in this situation I support the switch-case, this seems fine. However, could you return something worse on the fall-through case? Either raise to let someone know they asked for a thing which couldn't be `get()`'ed, or if not that for some reason, even just a string saying `"You asked for %s and I don't know how to give you one of those" % dataType`?

mfisherlevine: OK, the latter suggestion is pretty dumb as it'll just end up raising one step down the line, but I do feel that falling through and silently giving `None` isn't great; it's already hard enough to debug things involving the butler.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+    )
+    darkTime = pexConfig.Field(
+        dtype=float,
+        default=5.0,
+        doc="Exposure time for the dark current contribution.",
+    )
+    flatScale = pexConfig.Field(
+        dtype=float,
+        default=320.0,
+        doc="Scale factor to apply to the flat dome.  Should not need to be so big.",
```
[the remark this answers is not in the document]
czwa: This was unclear, so I've switched to a hopefully more intuitive configuration option (flatDrop) with better description (Fractional flux drop due to flat from center to edge of detector along x-axis.), and updated the flat method to convert the option into the necessary math.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        self.crosstalkCoeffs = np.array([[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+                                         [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]])
+
+        self.bfKernel = np.array([[1., 4., 7., 4., 1.],
```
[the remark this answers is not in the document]
czwa: I haven't dealt with BF kernels enough to have a good handle on their shapes.  I wanted something flux conserving, as well as something small enough to not require loading from an external source.  This kernel does run through the ISR brighterFatterCorrection code, which was my other main constraint.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+    `lsst.afw.cameraGeom.testUtils` to avoid making the test data
+    dependent on any of the actual obs package formats.
+    """
+    ConfigClass = IsrMockConfig
+    _DefaultName = "isrMock"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.rng = np.random.RandomState(self.config.rngSeed)
+        self.crosstalkCoeffs = np.array([[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
```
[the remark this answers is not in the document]
czwa: I've made it a bit different, adding a few cross terms as well as making another amplifier problematic.  I made the original sparse out of a sense of optimism.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+
+                self.amplifierAddYGradient(ampData, -1.0 * self.config.overscanScale,
+                                           1.0 * self.config.overscanScale)
+                self.amplifierAddYGradient(oscanData, -1.0 * self.config.overscanScale,
+                                           1.0 * self.config.overscanScale)
+
+        if self.config.doGenerateDict is True:
+            expDict = dict()
+            for amp in exposure.getDetector():
+                expDict[amp.getName()] = exposure
```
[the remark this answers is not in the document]
czwa: I believe that's what the underlying use case expects.  AssembleCcdTask can accept a dict of exposures, which I think is the case in which each amplifier is written (and read in) as a single FITS file on disk.  I do not know when this situation arises, but it was not being adequately tested prior to this addition.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        ----------
+        ampData : `lsst.afw.image.ImageF`
+            Amplifier image to operate on.
+        scale : float
+            Peak flux of the source to add.
+        """
+        for x in range(0, ampData.getDimensions().getX()):
+            for y in range(0, ampData.getDimensions().getY()):
+                ampData.getArray()[y][x] = (ampData.getArray()[y][x] +
+                                            scale * np.exp(-0.5 * ((x - 50)**2 + (y - 25)**2)**2 / 3.0**2))
```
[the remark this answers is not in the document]
czwa: I've not noticed speed issues, but this could easily loop only over the region the source is significant.  The placement was arbitrary, but putting a real signal near the defect does (optionally) test that defect interpolation looks reasonable.  With the sharp star (a think-o more than a typo), the interpolated region only covers the sky noise, but I'll check when running the tests.

python/lsst/ip/isr/isrMock.py
czwa: This functional form is also arbitrary, and although many other forms would work, this met all of my initial criteria:

- Easily recognizable by eye (the arcs are obviously artificial, and are also not straight lines).
- Small number of free parameters (although DM-4441 breaks this to allow multiple fringe patterns; that will allow the center and flux scale to be arrays of floats).
- Varying intensity of signal (I don't think this is exercised the way I wanted, but I wanted the signal to be large in some amplifiers and small in others to check that the correction deals well with that).

python/lsst/ip/isr/isrMock.py
czwa: For this afw.cameraGeom detector, 100 pixels does center the Gaussian.  As part of making the scale factor more intuitive, I've added the actual center as a parameter that can be set.

python/lsst/ip/isr/isrMock.py
czwa: I've changed to `amplifierMultiplyFlat` to be mathematically consistent, and removed all mentions of "dome" for clarity.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+
+    # Signal parameters
+    skyLevel = pexConfig.Field(
+        dtype=float,
+        default=1000.0,
+        doc="Background contribution to be generated from 'the sky'.",
+    )
+    skySigma = pexConfig.Field(
+        dtype=float,
+        default=32.0,
```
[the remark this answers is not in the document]
czwa: Removed as part of the noise update.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        doc="Index for the detector to use.  The default value uses a standard 2x4 array of amps.",
+    )
+    rngSeed = pexConfig.Field(
+        dtype=int,
+        default=20000913,
+        doc="Seed for random number generator used to add noise.",
+    )
+    gain = pexConfig.Field(
+        dtype=float,
+        default=1.0,
```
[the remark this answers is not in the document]
czwa: I forgot to sort this out before, but the idea was to set the gain to 1.0, ignore it during development, and then fold it in once everything worked.  The order reversal between this and the standard ISR processing, as well as trying to keep units in reasonable units complicates things.  I think I've included the gain as needed to keep things consistent.
The `readNoise` (replaces `biasSigma`) and `darkRate` are now assumed to be the only units in electrons, with all others in DN.  The gain is used to convert these parameters, matching the units that seem more reasonable (i.e., the skyLevel is kept in DN, as I have never heard someone cite "the sky was 4000 electrons").

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        doc="Amplitude of the ramp function to add to overscan data.",
+    )
+    biasLevel = pexConfig.Field(
+        dtype=float,
+        default=8000.0,
+        doc="Background contribution to be generated from the bias offset.",
+    )
+    biasSigma = pexConfig.Field(
+        dtype=float,
+        default=90.0,
```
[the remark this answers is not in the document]
czwa: As above, changed to `readNoise` in electrons.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+        doc="Background noise contribution to be generated from the bias.",
+    )
+    darkRate = pexConfig.Field(
+        dtype=float,
+        default=5.0,
+        doc="Background level contribution to be generated from dark current.",
```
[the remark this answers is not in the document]
czwa: Now explicitly in electrons per second.

python/lsst/ip/isr/isrMock.py
```diff
@@ -0,0 +1,915 @@
...
+                for idxT, ampT in enumerate(exposure.getDetector()):
+                    if self.config.isTrimmed is True:
+                        ampDataS = exposure.image[ampS.getBBox()]
+                        ampDataT = exposure.image[ampT.getBBox()]
+                    else:
+                        ampDataS = exposure.image[ampS.getRawDataBBox()]
+                        ampDataT = exposure.image[ampT.getRawDataBBox()]
+                    self.amplifierAddCT(ampDataS, ampDataT, self.crosstalkCoeffs[idxT][idxS])
+
+        for idx, amp in enumerate(exposure.getDetector()):
```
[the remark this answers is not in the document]
czwa: The enumerate is needed so that the source can be added to amplifier 0.

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
mfisherlevine: OK, that all sounds great! Just make sure that the units/assumptions are clearly documented and this should be good to go 👍

Maybe also make a ticket for going back and making the gain `!=1` and stick it here as a `TODO:`

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
mfisherlevine: That sounds great. I'm a bit fan of really verbose names, so feel free to ignore this, but I'd probably call it something like `flatFieldDropOff` rather than `flatDrop`, but like I said, your call.

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
mfisherlevine: If it runs through and doesn't totally ruin everything then that's probably fine. Otherwise I can dig you out a more realistic one, they're not huge, usually 11x11 or 19x19 at worst I think.

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
mfisherlevine: That's true in the previous looping over the detector, but not for this one, as far as I can see. This loop has no reference to `idx` as the `doSource` is the only thing that tests that, and that's done before the crosstalk.

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
mfisherlevine: Hmmm. I don't know what the underlying use case is, but nothing should ever expect a dict that has 16 keys, but each of which has a value that's the entire object, that's madness.

If the target is `AssembleCcdTask` that will accept a raw, assembled exposure, as returned by `butler.get('raw')`, which indeed has a detector which has a dict of amps, but not a dict with each key containing the entire exposure.

Does this make any sense?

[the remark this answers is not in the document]
mfisherlevine: To help a little, to trim an as-buttled raw, do:

```from lsst.ip.isr import AssembleCcdTask
config = AssembleCcdTask.ConfigClass()
config.doTrim = True

rawExp = butler.get('raw', visit=someVisit)

assembleTask = AssembleCcdTask(config=config)
rawTrimmed = assembleTask.assembleCcd(rawExp)
```

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
mfisherlevine: Yeah, operating on 1Mpix instead of 16Mpix isn't too bad, especially if you're not doing it multiple times. I think the star sharpness was the main point here, so as long as that's fixed and it's not problematically slow it should be fine (though I don't really know where this stuff will be used, and that certainly drives how much we should care about speed).

python/lsst/ip/isr/isrMock.py
mfisherlevine: In case this helps, I wrote this yesterday:

```
def _isExposureTrimmed(exp):
    det = exp.getDetector()
    if exp.getDimensions() == det.getBBox().getDimensions():
        return True
    return False
```

python/lsst/ip/isr/isrMock.py
mfisherlevine: That all sounds very reasonable. I still think that it decays too quickly to meet your last spec though, and the signal will be undetectable in most amps without putting an absurdly large signal down at the centre. Maybe just tweak the 50 pixel length scale in this case?

python/lsst/ip/isr/isrMock.py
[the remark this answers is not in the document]
czwa: My intention was that this would be used for unit tests as well as for development when real data is either unavailable or has confounding signals.  I'm not expecting this to ever be called in regular processing, so I haven't been worried about speed.

python/lsst/ip/isr/isrMock.py
czwa: I was convinced this was right, looked into it, and discovered we were both wrong.
The `amplifierAddFringe` and `amplifierMultiplyFlat` both need to calculate in the trimmed coordinates, but write the results in the untrimmed coordinates (and, as you mention, not overwrite the overscan).  These methods iterate over the exposure.image[ampBBox].getDimensions() coordinates, which correctly only write to the data region.  However, the conversion method `localCoordToExpCoord` was improperly offsetting things in the untrimmed case (converting from local amp coordinates to untrimmed exposure coordinates, introducing discontinuities across amplifier edges).
This is fixed, and I've confirmed that a fringe made for an untrimmed exposure that is then trimmed by assembleCcd matches a fringe made for a trimmed exposure.

python/lsst/ip/isr/isrMock.py
czwa: I've increased this, and will think about making it a parameter in the multi-fringe testing ticket (DM-4441).

python/lsst/ip/isr/isrMock.py
mfisherlevine: OK, sounds good!

PR #83, opened 2019-03-19
DM-15683: Add tests for ip_isr/isrTask

czwa:
These tests have finally converged with the addition of the `ip_isr/isrMock.py` code to supply "realistic" simulated data to check.  Running these tests uncovered some bugs in the ISR code, which has now been corrected.

The code coverage is now much better than previously, with the only major untested code being in the methods to support Gen3 butler/processing.

### DIFF
--- a/python/lsst/ip/isr/isrMock.py
+++ b/python/lsst/ip/isr/isrMock.py
@@ -330,11 +330,8 @@ class IsrMock(pipeBase.Task):
         defectList : `lsst.meas.algorithms.Defects`
             Simulated defect list
         """
-        defectList = []
-        box = lsst.geom.Box2I(lsst.geom.Point2I(0, 0),
-                              lsst.geom.Extent2I(40, 50))
-        defectList.append(box)
-        return Defects(defectList)
+        return Defects([lsst.geom.Box2I(lsst.geom.Point2I(0, 0),
+                                        lsst.geom.Extent2I(40, 50))])
 
     def makeCrosstalkCoeff(self):
         """Generate the simulated crosstalk coefficients.
--- a/python/lsst/ip/isr/isrMock.py
+++ b/python/lsst/ip/isr/isrMock.py
@@ -40,7 +40,7 @@ __all__ = ["IsrMockConfig", "IsrMock", "RawMock", "TrimmedRawMock", "RawDictMock
 
 
 class IsrMockConfig(pexConfig.Config):
-    r"""Configuration parameters for isrMock.
+    """Configuration parameters for isrMock.
 
     These parameters produce generic fixed position signals from
     various sources, and combine them in a way that matches how those
@@ -229,7 +229,7 @@ class IsrMockConfig(pexConfig.Config):
 
 
 class IsrMock(pipeBase.Task):
-    r"""Class to generate consistent mock images for ISR testing.
+    """Class to generate consistent mock images for ISR testing.
 
     ISR testing currently relies on one-off fake images that do not
     accurately mimic the full set of detector effects. This class
@@ -259,7 +259,7 @@ class IsrMock(pipeBase.Task):
                                   [1., 4., 7., 4., 1.]]) / 273.0
 
     def run(self):
-        r"""Generate a mock ISR product, and return it.
+        """Generate a mock ISR product, and return it.
 
         Returns
         -------
@@ -285,7 +285,7 @@ class IsrMock(pipeBase.Task):
             return None
 
     def makeData(self):
-        r"""Generate simulated ISR data.
+        """Generate simulated ISR data.
 
         Currently, only the class defined crosstalk coefficient
         matrix, brighter-fatter kernel, a constant unity transmission
@@ -313,7 +313,7 @@ class IsrMock(pipeBase.Task):
             return None
 
     def makeBfKernel(self):
-        r"""Generate a simple Gaussian brighter-fatter kernel.
+        """Generate a simple Gaussian brighter-fatter kernel.
 
         Returns
         -------
@@ -323,7 +323,7 @@ class IsrMock(pipeBase.Task):
         return self.bfKernel
 
     def makeDefectList(self):
-        r"""Generate a simple single-entry defect list.
+        """Generate a simple single-entry defect list.
 
         Returns
         -------
@@ -334,7 +334,7 @@ class IsrMock(pipeBase.Task):
                                        lsst.geom.Extent2I(40, 50)))
 
     def makeCrosstalkCoeff(self):
-        r"""Generate the simulated crosstalk coefficients.
+        """Generate the simulated crosstalk coefficients.
 
         Returns
         -------
@@ -345,7 +345,7 @@ class IsrMock(pipeBase.Task):
         return self.crosstalkCoeffs
 
     def makeTransmissionCurve(self):
-        r"""Generate a simulated flat transmission curve.
+        """Generate a simulated flat transmission curve.
 
         Returns
         -------
@@ -356,7 +356,7 @@ class IsrMock(pipeBase.Task):
         return afwImage.TransmissionCurve.makeIdentity()
 
     def makeImage(self):
-        r"""Generate a simulated ISR image.
+        """Generate a simulated ISR image.
 
         Returns
         -------
@@ -474,7 +474,7 @@ class IsrMock(pipeBase.Task):
 
     # afw primatives to construct the image structure
     def getCamera(self):
-        r"""Construct a test camera object.
+        """Construct a test camera object.
 
         Returns
         -------
@@ -486,7 +486,7 @@ class IsrMock(pipeBase.Task):
         return camera
 
     def getExposure(self):
-        r"""Construct a test exposure.
+        """Construct a test exposure.
 
         The test exposure has a simple WCS set, as well as a list of
         unlikely header keywords that can be removed during ISR
@@ -536,7 +536,7 @@ class IsrMock(pipeBase.Task):
         return exposure
 
     def getWcs(self):
-        r"""Construct a dummy WCS object.
+        """Construct a dummy WCS object.
 
         Taken from the deprecated ip_isr/examples/exampleUtils.py.
 
@@ -553,7 +553,7 @@ class IsrMock(pipeBase.Task):
                                   cdMatrix=afwGeom.makeCdMatrix(scale=1.0*lsst.geom.degrees))
 
     def localCoordToExpCoord(self, ampData, x, y):
-        r"""Convert between a local amplifier coordinate and the full
+        """Convert between a local amplifier coordinate and the full
         exposure coordinate.
 
         Parameters
@@ -584,7 +584,7 @@ class IsrMock(pipeBase.Task):
 
     # Simple data values.
     def amplifierAddNoise(self, ampData, mean, sigma):
-        r"""Add Gaussian noise to an amplifier's image data.
+        """Add Gaussian noise to an amplifier's image data.
 
          This method operates in the amplifier coordinate frame.
 
@@ -602,7 +602,7 @@ class IsrMock(pipeBase.Task):
                                                 size=ampData.getDimensions()).transpose()
 
     def amplifierAddYGradient(self, ampData, start, end):
-        r"""Add a y-axis linear gradient to an amplifier's image data.
+        """Add a y-axis linear gradient to an amplifier's image data.
 
          This method operates in the amplifier coordinate frame.
 
@@ -621,7 +621,7 @@ class IsrMock(pipeBase.Task):
                                  np.zeros(ampData.getDimensions()).transpose())
 
     def amplifierAddSource(self, ampData, scale, x0, y0):
-        r"""Add a single Gaussian source to an amplifier.
+        """Add a single Gaussian source to an amplifier.
 
          This method operates in the amplifier coordinate frame.
 
@@ -642,7 +642,7 @@ class IsrMock(pipeBase.Task):
                                        scale * np.exp(-0.5 * ((x - x0)**2 + (y - y0)**2) / 3.0**2))
 
     def amplifierAddCT(self, ampDataSource, ampDataTarget, scale):
-        r"""Add a scaled copy of an amplifier to another, simulating crosstalk.
+        """Add a scaled copy of an amplifier to another, simulating crosstalk.
 
          This method operates in the amplifier coordinate frame.
 
@@ -664,7 +664,7 @@ class IsrMock(pipeBase.Task):
 
     # Functional form data values
     def amplifierAddFringe(self, amp, ampData, scale):
-        r"""Add a fringe-like ripple pattern to an amplifier's image data.
+        """Add a fringe-like ripple pattern to an amplifier's image data.
 
         Parameters
         ----------
@@ -690,7 +690,7 @@ class IsrMock(pipeBase.Task):
                                        scale * np.sinc(((u - 100)/150)**2 + (v / 150)**2))
 
     def amplifierMultiplyFlat(self, amp, ampData, fracDrop, u0=100.0, v0=100.0):
-        r"""Multiply an amplifier's image data by a flat-like pattern.
+        """Multiply an amplifier's image data by a flat-like pattern.
 
         Parameters
         ----------
@@ -725,7 +725,7 @@ class IsrMock(pipeBase.Task):
 
 
 class RawMock(IsrMock):
-    r"""Generate a raw exposure suitable for ISR.
+    """Generate a raw exposure suitable for ISR.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -741,7 +741,7 @@ class RawMock(IsrMock):
 
 
 class TrimmedRawMock(RawMock):
-    r"""Generate a trimmed raw exposure.
+    """Generate a trimmed raw exposure.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -750,7 +750,7 @@ class TrimmedRawMock(RawMock):
 
 
 class CalibratedRawMock(RawMock):
-    r"""Generate a trimmed raw exposure.
+    """Generate a trimmed raw exposure.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -766,7 +766,7 @@ class CalibratedRawMock(RawMock):
 
 
 class RawDictMock(RawMock):
-    r"""Generate a raw exposure dict suitable for ISR.
+    """Generate a raw exposure dict suitable for ISR.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -774,7 +774,7 @@ class RawDictMock(RawMock):
 
 
 class MasterMock(IsrMock):
-    r"""Parent class for those that make master calibrations.
+    """Parent class for those that make master calibrations.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -792,7 +792,7 @@ class MasterMock(IsrMock):
 
 
 class BiasMock(MasterMock):
-    r"""Simulated master bias calibration.
+    """Simulated master bias calibration.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -801,7 +801,7 @@ class BiasMock(MasterMock):
 
 
 class DarkMock(MasterMock):
-    r"""Simulated master dark calibration.
+    """Simulated master dark calibration.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -810,7 +810,7 @@ class DarkMock(MasterMock):
 
 
 class FlatMock(MasterMock):
-    r"""Simulated master flat calibration.
+    """Simulated master flat calibration.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -818,7 +818,7 @@ class FlatMock(MasterMock):
 
 
 class FringeMock(MasterMock):
-    r"""Simulated master fringe calibration.
+    """Simulated master fringe calibration.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -826,7 +826,7 @@ class FringeMock(MasterMock):
 
 
 class UntrimmedFringeMock(FringeMock):
-    r"""Simulated untrimmed master fringe calibration.
+    """Simulated untrimmed master fringe calibration.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -834,7 +834,7 @@ class UntrimmedFringeMock(FringeMock):
 
 
 class BfKernelMock(IsrMock):
-    r"""Simulated brighter-fatter kernel.
+    """Simulated brighter-fatter kernel.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -847,7 +847,7 @@ class BfKernelMock(IsrMock):
 
 
 class DefectMock(IsrMock):
-    r"""Simulated defect list.
+    """Simulated defect list.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -860,7 +860,7 @@ class DefectMock(IsrMock):
 
 
 class CrosstalkCoeffMock(IsrMock):
-    r"""Simulated crosstalk coefficient matrix.
+    """Simulated crosstalk coefficient matrix.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -873,7 +873,7 @@ class CrosstalkCoeffMock(IsrMock):
 
 
 class TransmissionMock(IsrMock):
-    r"""Simulated transmission curve.
+    """Simulated transmission curve.
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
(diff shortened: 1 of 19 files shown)


tests/test_fringes.py
```diff
@@ -252,6 +260,15 @@ def testRunDataRef(self, pedestal=0.0, stddevMax=1.0e-4):
...
         self.assertLess(afwMath.makeStatistics(mi, afwMath.STDEV).getValue(), stddevMax)
 
+    def test_readFringes(self):
+        r"""Test that fringes can be successfully accessed from the butler.
+        """
+        task = FringeTask()
+        dataRef = isrMock.DataRefMock()
+
+        result = task.readFringes(dataRef, assembler=None)
+        assert result
```
parejkoj: What are you checking? You have a `dataRefMock`, so I assume you can test that it returned the expected thing, instead of just not being None.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+    Returns
+    -------
+    median : `float`
+        Image median.
+    std : `float`
+        Image stddev.
+    """
+    median = np.nanmedian(image.getArray())
+    std = np.nanstd(image.getArray())
+    print("ms: %f %f" % (median, std))
```
parejkoj: No print statements in tests, please: they just get lost.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+    Returns
+    -------
+    nMask : `int`
+        Number of masked pixels.
+    """
+    bitMask = maskedImage.getMask().getPlaneBitMask(maskPlane)
+    isBit = maskedImage.getMask().getArray() & bitMask > 0
+    numBit = np.sum(isBit)
+    print("%s %d %d\n" % (maskPlane, bitMask, numBit))
```
parejkoj: No print statements in tests, please: they just get lost.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+import lsst.afw.image as afwImage
+import lsst.utils.tests
+import lsst.ip.isr as ipIsr
+import lsst.pex.exceptions as pexExcept
+import lsst.ip.isr.isrMock as isrMock
+
+
+def maskCountPixels(maskedImage, maskPlane):
+    r"""Function to count the number of masked pixels of a given type.
```
parejkoj: Wouldn't this be `countMaskedPixels` then, given the docstring?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    bitMask = maskedImage.getMask().getPlaneBitMask(maskPlane)
+    isBit = maskedImage.getMask().getArray() & bitMask > 0
+    numBit = np.sum(isBit)
+    print("%s %d %d\n" % (maskPlane, bitMask, numBit))
+    return numBit
+
+
+def imageMedianStd(image):
+    r"""Function to calculate median and std of image data.
```
parejkoj: Maybe name it  `computeImageMedianAndStd`?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+class IsrFunctionsCases(lsst.utils.tests.TestCase):
+    r"""Test functions for ISR produce expected outputs.
+    """
+    def setUp(self):
+        self.inputExp = isrMock.TrimmedRawMock().run()
+        self.mi = self.inputExp.getMaskedImage()
+
+    def tearDown(self):
+        pass
```
parejkoj: Unnecessary if it's not overloaded.

tests/test_assembleCcd.py
```diff
@@ -0,0 +1,74 @@
...
+#
+
+import unittest
+
+import lsst.utils.tests
+from lsst.ip.isr.assembleCcdTask import (AssembleCcdConfig, AssembleCcdTask)
+import lsst.ip.isr.isrMock as isrMock
+
+
+class AssembleCcdCases(lsst.utils.tests.TestCase):
```
parejkoj: We usually call our test classes `BlahTestCase`.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def test_transposeMaskedImage(self):
+        r"""Image transpose.
+
+        Expect height and width to be exchanged.
```
parejkoj: This is the only informative part of the docstring. The the first line is implied by the method name.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        transposed = ipIsr.transposeMaskedImage(self.mi)
+        self.assertEqual(transposed.getImage().getBBox().getHeight(),
+                         self.mi.getImage().getBBox().getWidth())
+        self.assertEqual(transposed.getImage().getBBox().getWidth(),
+                         self.mi.getImage().getBBox().getHeight())
```
parejkoj: It's worth also checking here that the image itself is transposed, no?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+        Expect number of interpolated pixels to be non-zero.
+        """
+        defectList = isrMock.DefectMock().run()
+        assert len(defectList) > 0
+
+        for fallbackValue in (None, -999.0):
+            for haveMask in (True, False):
+                with self.subTest(fallbackValue=fallbackValue, haveMask=haveMask):
+                    if haveMask is False:
```
parejkoj: Another block that should be in a separate method, or methods, to make it more obvious what is being tested.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+        transposed = ipIsr.transposeMaskedImage(self.mi)
+        self.assertEqual(transposed.getImage().getBBox().getHeight(),
+                         self.mi.getImage().getBBox().getWidth())
+        self.assertEqual(transposed.getImage().getBBox().getWidth(),
+                         self.mi.getImage().getBBox().getHeight())
+
+    def test_interpolateDefectList(self):
+        r"""Defect interpolation.
+
+        Expect number of interpolated pixels to be non-zero.
```
parejkoj: Again, this statement should be the entirety of the docstring: the first part doesn't add anything. Similarly for the other test docstrings below.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        defectList = isrMock.DefectMock().run()
+        assert len(defectList) > 0
```
parejkoj: What is this assert for?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+            for haveMask in (True, False):
+                with self.subTest(fallbackValue=fallbackValue, haveMask=haveMask):
+                    if haveMask is False:
+                        if 'INTRP' in self.mi.getMask().getMaskPlaneDict():
+                            self.mi.getMask().removeAndClearMaskPlane('INTRP')
+                    else:
+                        if 'INTRP' not in self.mi.getMask().getMaskPlaneDict():
+                            self.mi.getMask().addMaskPlane('INTRP')
+                    numBit = maskCountPixels(self.mi, "INTRP")
+                    self.assertGreaterEqual(numBit, 0)
```
parejkoj: You should know exactly the correct value of `numBit` for every case, no?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+            assert d.getBBox().getDimensions().getY() == t.getBBox().getDimensions().getX()
+
+    def test_makeThresholdMask(self):
+        r"""Thresholding.
+
+        Expect list of defects to have elements.
+        """
+        defectList = ipIsr.makeThresholdMask(self.mi, 200, growFootprints=2, maskName='SAT')
+
+        self.assertGreater(len(defectList), 0)
```
parejkoj: You should know the exact size of this `defectList` a priori, no? Similarly for the other `assertGreater` style tests below: Since you've constructed the test data, you should know the exact results.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def test_interpolateFromMask(self):
+        r"""Interpolate mask.
+
+        Expect number of interpolated pixels to be non-zero.
+        """
+        ipIsr.makeThresholdMask(self.mi, 200, growFootprints=2, maskName='SAT')
+        for growFootprints in (0, 2):
```
parejkoj: I would also test `1` here, to ensure that odd numbers also work (and 1 is sometimes a dangerous case).

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+                                                              growFootprints=growFootprints, maskName='SAT')
+                numBit = maskCountPixels(interpMaskedImage, "INTRP")
+                self.assertGreaterEqual(numBit, 0)
+
+    def test_saturationCorrection(self):
+        r"""Saturation correction.
+
+        Expect number of mask pixels with SAT marked to be non-zero.
+        """
+        for interpolate in (True, False):
```
parejkoj: Tests of exactly two case are often clearer as two separate tests: `test_saturationCorrectionInterpolate`, `test_saturationCorrectionNoInterpolate`.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        darkExp = isrMock.DarkMock().run()
+        darkMi = darkExp.getMaskedImage()
+
+        image = self.mi.getImage().getArray()
+        ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=True)
+        ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=True, invert=True)
+
+        diff = image - self.mi.getImage().getArray()
+        assert (diff.sum() == 0)
```
parejkoj: Why not `assertMaskedImagesEqual(mi, self.mi)`, with `mi = self.mi.clone()` above?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+        Expect round-trip application to be equal.
+        Expect RuntimeError if sizes are different.
+        """
+        darkExp = isrMock.DarkMock().run()
+        darkMi = darkExp.getMaskedImage()
+
+        image = self.mi.getImage().getArray()
+        ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=True)
+        ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=True, invert=True)
```
parejkoj: It also would be good to have a comment about what these two lines are doing: I had to look up the meaning of `invert` in this context.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+        diff = image - self.mi.getImage().getArray()
+        assert (diff.sum() == 0)
+
+        newDark = darkMi[1:-1, 1:-1, afwImage.LOCAL]
+        darkMi = newDark
```
parejkoj: Why not `darkMi = darkMi[1:-1, 1:-1, afwImage.LOCAL]`? Similarly in various places below.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+
+        Expect smaller median image value after.
+        Expect RuntimeError is sizes are different.
+        """
+        biasExp = isrMock.BiasMock().run()
+        biasMi = biasExp.getMaskedImage()
+
+        ipIsr.biasCorrection(self.mi, biasMi, trimToFit=True)
+
+        assert True
```
parejkoj: Did you mean to write an assertion test here?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+        newBias = biasMi[1:-1, 1:-1, afwImage.LOCAL]
+        biasMi = newBias
+
+        with self.assertRaises(RuntimeError):
+            ipIsr.biasCorrection(self.mi, biasMi, trimToFit=False)
+
+    def test_flatCorrection(self):
+        r"""Flat correction.
+
+        Expect round-trip application to be equal
```
parejkoj: Period after sentence.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+            with self.subTest(scaling=scaling):
+                if scaling == 'UNKNOWN':
+                    with self.assertRaises(pexExcept.Exception):
+                        ipIsr.flatCorrection(self.mi, flatMi, scaling, userScale=1.0, trimToFit=True)
+                        ipIsr.flatCorrection(self.mi, flatMi, scaling, userScale=1.0,
+                                             trimToFit=True, invert=True)
+                else:
+                    ipIsr.flatCorrection(self.mi, flatMi, scaling, userScale=1.0, trimToFit=True)
+                    ipIsr.flatCorrection(self.mi, flatMi, scaling, userScale=1.0,
+                                         trimToFit=True, invert=True)
```
parejkoj: This lower block doesn't test anything.

This is another loop that would be clearer as separate methods, or at least with `UNKNOWN` as a separate method.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,402 @@
...
+        Expect RuntimeError if sizes are different or if scaling type isn't known.
+        """
+        flatExp = isrMock.FlatMock().run()
+        flatMi = flatExp.getMaskedImage()
+
+        image = self.mi.getImage().getArray()
+        for scaling in ('USER', 'MEAN', 'MEDIAN', 'UNKNOWN'):
+            with self.subTest(scaling=scaling):
+                if scaling == 'UNKNOWN':
+                    with self.assertRaises(pexExcept.Exception):
```
parejkoj: Which specific exception is this supposed to raise?

czwa: This now raises a RuntimeError instead of leaving a generic exception.

tests/test_crosstalk.py
```diff
@@ -201,6 +213,23 @@ def runDataRef(self, dataRef):
...
+        r"""Test that prep crosstalk does not error when given a dataRef with no
+        crosstalkSources to find.
+        """
+        dataRef = Struct(dataId={'fake': 1})
+        task = CrosstalkTask()
+        result = task.prepCrosstalk(dataRef)
+        assert result is None
+
+    def test_nullCrosstalkTask(self):
+        r"""Test that the null crosstalk task does not create an error.
```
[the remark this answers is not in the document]
czwa: Crosstalk tests will be updated as part of an already existing crosstalk ticket (DM-19214).  I think actual tests will be added when that converges.

czwa:
The updated pull request resolves these issues (hopefully).  I had no issues with the style and pytest/unittest points.  I've punted on the crosstalk, as a future ticket (DM-19214) will implement changes that will allow more complete unit tests to be written.  I've also left a few tests with poor assertions simply because I'd rather have a test run through the code and be unhelpful than to leave that code entirely untested.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,416 @@
...
+        # The `invert` parameter controls the direction of the
+        # application.  This will apply, and un-apply the dark.
+        ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=True)
+        ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=True, invert=True)
+
+        self.assertMaskedImagesAlmostEqual(self.mi, mi, atol=1e-3)
+
+        darkMi = darkMi[1:-1, 1:-1, afwImage.LOCAL]
+
+        with self.assertRaises(RuntimeError):
```
parejkoj: Remove the newline here, so it's clear that the above line goes with this one.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,416 @@
...
+            ipIsr.darkCorrection(self.mi, darkMi, 1.0, 1.0, trimToFit=False)
+
+    def test_biasCorrection(self):
+        """Expect smaller median image value after.
+        Expect RuntimeError is sizes are different.
```
parejkoj: "is" -> "if"?

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,416 @@
...
+        biasMi = biasExp.getMaskedImage()
+
+        mi = self.mi.clone()
+        ipIsr.biasCorrection(self.mi, biasMi, trimToFit=True)
+        self.assertLess(computeImageMedianAndStd(self.mi.getImage())[0],
+                        computeImageMedianAndStd(mi.getImage())[0])
+
+        biasMi = biasMi[1:-1, 1:-1, afwImage.LOCAL]
+
+        with self.assertRaises(RuntimeError):
```
parejkoj: Please remove this newline to connect logical flow.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,416 @@
...
+            ipIsr.biasCorrection(self.mi, biasMi, trimToFit=False)
+
+    def test_flatCorrection(self):
+        """Expect round-trip application to be equal.
+        Expect RuntimeError if sizes are different or if scaling type isn't known.
```
parejkoj: How does this test test these two things? I only see the one `assertRaises` below.

tests/test_isrFunctions.py
```diff
@@ -0,0 +1,416 @@
...
+        flatMi = flatExp.getMaskedImage()
+
+        mi = self.mi.clone()
+        ipIsr.illuminationCorrection(self.mi, flatMi, 1.0)
+        self.assertGreater(computeImageMedianAndStd(self.mi.getImage())[0],
+                           computeImageMedianAndStd(mi.getImage())[0])
+
+        flatMi = flatMi[1:-1, 1:-1, afwImage.LOCAL]
+
+        with self.assertRaises(RuntimeError):
```
parejkoj: Please remove this newline to connect logical flow.

tests/test_isrMisc.py
```diff
@@ -0,0 +1,104 @@
...
+
+        self.assertIsInstance(result, afwGeom.Polygon)
+
+    def test_masking(self):
+        """Assert that the masking task does not error when given an exposure.
+        """
+        task = masking.MaskingTask()
+        result = task.run(self.inputExp)
+
+        self.assertIsNone(result)
```
parejkoj: Can you make this an explicit check, since you know exactly what pixels should be masked in the mock exposure.

tests/test_isrQa.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.assertTrue(np.nonzero(thumb))
+
+    def test_writeThumbnail(self):
+        """Test that a thumbnail can be "written" to disk by a mock dataRef.
+        """
+        dataRef = isrMock.DataRefMock()
+        thumb = isrQa.makeThumbnail(self.inputExp, self.config)
+
+        isrQa.writeThumbnail(dataRef, thumb, "thumbnail")
```
parejkoj: What is the test here? Does `dataRef` have a way to check that the appropriate `write` methods were called?

tests/test_measureCrosstalk.py
```diff
@@ -0,0 +1,124 @@
...
+        config = isrMock.IsrMockConfig()
+        config.rngSeed = 12345
+        config.doAddCrosstalk = True
+        config.doAddSky = True
+        config.doAddSource = True
+        config.skyLevel = 0.0
+        config.readNoise = 0.0
+        mcConfig = MeasureCrosstalkConfig()
+        mcConfig.threshold = 4000
+        mct = MeasureCrosstalkTask(config=mcConfig)
```
parejkoj: This test, and the one below it, have a lot of shared setup that should probably live in a `setUp` method. Similarly for the `for` loop below. As written, I can't really tell what's different between these two tests.

tests/test_isrTask.py
```diff
@@ -0,0 +1,529 @@
...
+        """Test that an exposure has a usable instance class.
+        """
+        self.assertIsInstance(self.task.ensureExposure(self.inputExp, self.camera, 0),
+                              afwImage.Exposure)
+
+    def test_convertItoF(self):
+        """Test conversion from integer to floating point pixels.
+        """
+        self.assertIsInstance(self.task.convertIntToFloat(self.inputExp).getImage()[1, 1],
+                              float)
```
parejkoj: I would think you could make this test explicit:

```
result = self.task.convertIntToFloat(self.inputExp)
self.assertEqual(result.getImage().getArray().dtype, float)
self.assertEqual(result, self.inputExp)
```

tests/test_isrTask.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def test_updateVariance(self):
+        """Expect The variance image should have a larger median value after
+        this operation.
+        """
+        statBefore = computeImageMedianAndStd(self.inputExp.variance[self.amp.getBBox()])
+        self.task.updateVariance(self.inputExp, self.amp)
+        statAfter = computeImageMedianAndStd(self.inputExp.variance[self.amp.getBBox()])
+        self.assertGreater(statAfter[0], statBefore[0])
```
parejkoj: I would make any test that is doing `assertGreater` or `assertLess` use the empirical values instead. That way we at least know if some code causes a change in the algorithm. So here, make the test:

```
self.assertFloatsEqual(statBefore == (123, 0.432))
self.assertFloatsEqual(statBefore == (567, 0.432))
```

or whatever the current output is.

tests/test_isrTask.py
```diff
@@ -0,0 +1,529 @@
...
+    def test_flatContext(self):
+        """Expect the flat context manager runs successfully and leaves the
+        image data the same.
+        """
+        darkExp = isrMock.DarkMock().run()
+        flatExp = isrMock.FlatMock().run()
+
+        mi = self.inputExp.getMaskedImage().clone()
+        with self.task.flatContext(self.inputExp, flatExp, darkExp):
+            self.assertTrue(True)
```
parejkoj: What is this doing?

tests/test_isrTask.py
```diff
@@ -0,0 +1,529 @@
...
+
+        self.assertMaskedImagesAlmostEqual(mi, self.inputExp.getMaskedImage(), -3)
```
parejkoj: Why `-3` here? That's the `doImage` boolean argument.

tests/test_isrFunctions.py
czwa: This message was leftover from before I split the tests in two.  Updated.

tests/test_isrQa.py
czwa: I've removed the test, as it most likely doesn't increase code coverage.  I had originally added this with a thought towards gen3, but that isn't helpful here.

tests/test_isrTask.py
czwa: I've included both, because even with the comment indicating which way this should go, I would like to have it explicitly tested for (if for no other reason than to prevent me from changing the values blindly in the future).

tests/test_isrTask.py
czwa: Nothing (but the goal is to test that the flat context manager doesn't alter the image), and so is now replaced with pass.

parejkoj: Wouldn't the test then be `self.assertEqual(self.inputExp, flatContext(self.inputExp)` or something like that?

czwa: I don't think so.  The flatContext() should be applying the flat and dark to the input exposure, passing (in the current test) instead of operating on that now-processed exposure, and then removing the flat and dark so the input exposure can be operated on in the "clean" state.  The goal here is to test that this context based round-trip leaves the exposure the same as before.

tests/test_isrTask.py
```diff
@@ -0,0 +1,550 @@
...
+        """Expect the median image value should decrease after this operation.
+        """
+        darkIm = isrMock.DarkMock().run()
+
+        statBefore = computeImageMedianAndStd(self.inputExp.image[self.amp.getBBox()])
+        self.task.darkCorrection(self.inputExp, darkIm)
+        statAfter = computeImageMedianAndStd(self.inputExp.image[self.amp.getBBox()])
+        self.assertLess(statAfter[0], statBefore[0])
+        self.assertAlmostEqual(statBefore[0], 8075.6885, places=2)
+        self.assertAlmostEqual(statAfter[0], 8051.6206, places=2)
```
parejkoj: I think we're better off using `assertFloatsAmostEqual` here, with `rtol` or `atol`: I find `assertAlmostEqual` with `places` to be an uncomfortable fit to our usage of numbers in LSST.

python/lsst/ip/isr/isrMock.py
```diff
@@ -323,18 +323,21 @@ def makeBfKernel(self):
...
+        """Generate a simple single-entry defect list.
 
         Returns
         -------
         defectList : `lsst.meas.algorithms.Defects`
             Simulated defect list
         """
-        return Defects(lsst.geom.Box2I(lsst.geom.Point2I(0, 0),
-                                       lsst.geom.Extent2I(40, 50)))
+        defectList = []
```
timj: Sorry about that.  The quick fix is to switch to:
```python
return Defects([lsst.geom.Box2I(lsst.geom.Point2I(0, 0),
	                                       lsst.geom.Extent2I(40, 50))])
```

tests/test_isrTask.py
parejkoj: Oh, I see: you're making that test just below (it was confusing without the test docstring). Is there any way to identify that inside the context manager the image has been modified correctly?

czwa: As the context manager calls the same flatCorrection/darkCorrection methods, I was hoping the tests on those methods would cover that code.  I have added value tests on the statistics to do ensure that unexpected code changes trigger an error.

PR #85, opened 2019-04-03
DM-18703 constructFlat.py --config isr.doCrosstalkBeforeAssemble=False raises a LengthError exception

czwa:
In addition to the length error, this ticket also includes crosstalk correction in ip_isr to not subtract background levels calculating the corrected pixels, but to include that background when calculating the mask of significant CT pixels.

### DIFF
--- a/python/lsst/ip/isr/crosstalk.py
+++ b/python/lsst/ip/isr/crosstalk.py
@@ -187,6 +187,10 @@ def subtractCrosstalk(exposure, badPixels=["BAD"], minPixelToMask=45000,
     enough if the crosstalk is small (e.g., coefficients < ~ 1e-3), but if it's
     larger you may want to iterate.
 
+    This method needs unittests (DM-18876), but such testing requires
+    DM-18610 to allow the test detector to have the crosstalk
+    parameters set.
+
     Parameters
     ----------
     exposure : `lsst.afw.image.Exposure`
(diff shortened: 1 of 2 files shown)


czwa:
> The code looks good as is, but there are a couple more issues I think that need to be addressed.
> Re: the lengthError problems: Are there associated tests? This should be tested going both ways, and that will ensure that when DM-15409 is resolved this doesn't break again.

Yes, this is true, but no, there are no associated tests.  I've filed ticket DM-18876 on this issue, but that depends on changes to afw to allow the "hasCrosstalk" and "getCrosstalk" methods to be set for the test detector to make testing feasible.

> Re: the background subtraction... I think you need to go into all the obs packages (minus obs_lsst, I assume) and change the default background subtraction to be AMP or DETECTOR as appropriate. Although this is a fix, if we suddenly turn this on there will be a mismatch between flats and images and that will create much confusion. I believe that fixing this bug and setting the default to "None" for the global requires an RFC.

I did not make any changes to the obs packages because both HSC and DECam subclass their own crosstalk corrections from the ip_isr version.  I think at this point only obs_lsst calls to the ip_isr version.

> Also re: the subtraction without the background, have we checked that this continues to work well on sources? I worry about 2nd order effects. I think that the background subtraction makes the first-order approximation of the crosstalk correction that we have implemented work okay (for sources!), but if we turn that off then we need to be more careful with the crosstalk numbers. I could be wrong about this, but I don't know if it's been tested on real data.

I have not done extensive tests due to the assortment of issues above (camera specific CT code, HSC not getting the coefficients from the detector object, etc).  I agree that the sources should be fine to first order, and my thought is that any higher order terms are too small to be an issue.  I suspect the simulation that found this issue may end up being the best place to find some of these issues, especially if we have to rely on our own measurements of the crosstalk coefficients.

erykoff:
Oh, I didn't realize that this particular code was only called by `obs_lsst`, and it's a totally different code path for `obs_subaru` and `obs_decam`.  In which case, that objection is lifted!

And that also makes sense to go forward with this, and we can look further if the DC2 sims find any residual problems.  

Finally, I look forward to the afw changes that will allow tests to be performed!

PR #89, opened 2019-04-30
Fix "backward-compatible" maskPixelsFromDefectList

RobertLuptonTheGood:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -149,7 +149,7 @@ def maskPixelsFromDefectList(maskedImage, defectList, maskName='BAD'):
     maskName : str, optional
         Mask plane name to use.
     """
-    return defectList.maskPixels(maskedImage, maskName=maskName)
+    return lsst.meas.algorithms.Defects(defectList).maskPixels(maskedImage, maskName=maskName)
```
timj: I think `measAlg.Defects(defectList)` is closer to how it's done elsewhere.

PR #94, opened 2019-05-21
DM-4441: Update isrMock for multi-component fringes.

czwa:
Adding multi-component fringes to the mocks allows full tests to confirm
FringeTask handles these fits.  Return fringe solution and RMS values
for validation of the measured values.

### DIFF
--- a/tests/test_isrTask.py
+++ b/tests/test_isrTask.py
@@ -147,7 +147,7 @@ class IsrTaskTestCases(lsst.utils.tests.TestCase):
         statAfter = computeImageMedianAndStd(self.inputExp.variance[self.amp.getBBox()])
         self.assertGreater(statAfter[0], statBefore[0])
         self.assertFloatsAlmostEqual(statBefore[0], 0.0, atol=1e-2)
-        self.assertFloatsAlmostEqual(statAfter[0], 8175.6885, atol=1e-2)
+        self.assertFloatsAlmostEqual(statAfter[0], 8170.0195, atol=1e-2)
 
     def test_darkCorrection(self):
         """Expect the median image value should decrease after this operation.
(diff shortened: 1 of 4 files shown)


tests/test_isrTask.py
```diff
@@ -171,8 +171,8 @@ def test_darkCorrection_noVisitInfo(self):
         self.task.darkCorrection(self.inputExp, darkIm)
         statAfter = computeImageMedianAndStd(self.inputExp.image[self.amp.getBBox()])
         self.assertLess(statAfter[0], statBefore[0])
-        self.assertFloatsAlmostEqual(statBefore[0], 8075.6885, atol=1e-2)
-        self.assertFloatsAlmostEqual(statAfter[0], 8051.6206, atol=1e-2)
+        self.assertFloatsAlmostEqual(statBefore[0], 8070.0195, atol=1e-2)
+        self.assertFloatsAlmostEqual(statAfter[0], 8045.7773, atol=1e-2)
```
mfisherlevine: Could you also comment on why these test values need changing?

tests/test_isrTask.py
```diff
@@ -183,8 +183,8 @@ def test_flatCorrection(self):
         self.task.flatCorrection(self.inputExp, flatIm)
         statAfter = computeImageMedianAndStd(self.inputExp.image[self.amp.getBBox()])
         self.assertGreater(statAfter[1], statBefore[1])
-        self.assertFloatsAlmostEqual(statAfter[1], 147417.36, atol=1e-2)
-        self.assertFloatsAlmostEqual(statBefore[1], 148.28436, atol=1e-2)
+        self.assertFloatsAlmostEqual(statAfter[1], 147407.02, atol=1e-2)
```
mfisherlevine: And here (and the ones below, etc)

czwa: The "science" images used for these tests are generated with `doAddFringe=True`.  To better test the multi-fringe code, the wavelength of the function was cut by a factor of 3, which results in the fringe component being significant over a smaller area.  This gives slightly smaller median and sigma values (as more area is not influenced by the fringe pattern).

tests/test_isrTask.py
czwa: As below, there was a change in the fringe size on the image, resulting in smaller median/sigma values.

PR #95, opened 2019-05-21
DM-19214: ip_isr crosstalk shouldn't depend on detector hasCrosstalk( )and getCrosstalk()

czwa:
Crosstalk application currently demands that the exposure detector have the crosstalk coefficient data included in the detector definition.  This makes measuring CT on new detectors challenging, and is explicitly not used for the most well-tested cameras (HSC/DECam).  This change allows the crosstalk coefficients to be set in the isr.crosstalk configuration.

To support this change, MeasureCrosstalkTask grabbed previously isolated functions into the task class, and additional debug visualization methods were added to help assess the measurement quality.

### DIFF
--- a/python/lsst/ip/isr/crosstalk.py
+++ b/python/lsst/ip/isr/crosstalk.py
@@ -20,7 +20,7 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 """
-Apply intra-CCD crosstalk corrections
+Apply intra-detector crosstalk corrections
 """
 import numpy as np
 
@@ -35,7 +35,7 @@ __all__ = ["CrosstalkConfig", "CrosstalkTask", "subtractCrosstalk", "writeCrosst
 
 
 class CrosstalkConfig(Config):
-    """Configuration for intra-CCD crosstalk removal"""
+    """Configuration for intra-detector crosstalk removal."""
     minPixelToMask = Field(
         dtype=float,
         doc="Set crosstalk mask plane for pixels over this value.",
@@ -63,7 +63,10 @@ class CrosstalkConfig(Config):
     )
     crosstalkValues = ListField(
         dtype=float,
-        doc="Amplifier-indexed crosstalk coefficients to use.",
+        doc=("Amplifier-indexed crosstalk coefficients to use.  This should be arranged as a 1 x nAmp**2 "
+             "list of coefficients, such that when reshaped by crosstalkShape, the result is nAmp x nAmp. "
+             "This matrix should be structured so CT * [amp0 amp1 amp2 ...]^T returns the column "
+             "vector [corr0 corr1 corr2 ...]^T."),
         default=[0.0],
     )
     crosstalkShape = ListField(
@@ -73,7 +76,7 @@ class CrosstalkConfig(Config):
     )
 
     def getCrosstalk(self, detector=None):
-        """Return a 2-D numpy array of crosstalk coefficients of the proper shape.
+        """Return a 2-D numpy array of crosstalk coefficients in the proper shape.
 
         Parameters
         ----------
@@ -95,8 +98,8 @@ class CrosstalkConfig(Config):
             if detector is not None:
                 nAmp = len(detector)
                 if coeffs.shape != (nAmp, nAmp):
-                    raise RuntimeError("Constructed crosstalk coeffients do not match detector shape. %s %s",
-                                       coeffs.shape, nAmp)
+                    raise RuntimeError("Constructed crosstalk coeffients do not match detector shape. " +
+                                       f"{coeffs.shape} {nAmp}")
             return coeffs
         elif detector is not None and detector.hasCrosstalk() is True:
             # Assume the detector defines itself consistently.
@@ -126,12 +129,12 @@ class CrosstalkConfig(Config):
 
 
 class CrosstalkTask(Task):
-    """Apply intra-CCD crosstalk correction"""
+    """Apply intra-detector crosstalk correction."""
     ConfigClass = CrosstalkConfig
     _DefaultName = 'isrCrosstalk'
 
     def prepCrosstalk(self, dataRef):
-        """Placeholder for crosstalk preparation method, e.g., for inter-CCD crosstalk.
+        """Placeholder for crosstalk preparation method, e.g., for inter-detector crosstalk.
 
         Parameters
         ----------
@@ -145,16 +148,16 @@ class CrosstalkTask(Task):
         return
 
     def run(self, exposure, crosstalkSources=None, isTrimmed=False):
-        """Apply intra-CCD crosstalk correction
+        """Apply intra-detector crosstalk correction
 
         Parameters
         ----------
         exposure : `lsst.afw.image.Exposure`
             Exposure for which to remove crosstalk.
         crosstalkSources : `defaultdict`, optional
-            Image data and crosstalk coefficients from other CCDs/amps that are
+            Image data and crosstalk coefficients from other detectors/amps that are
             sources of crosstalk in exposure.
-            The default for intra-CCD crosstalk here is None.
+            The default for intra-detector crosstalk here is None.
         isTrimmed : `bool`
             The image is already trimmed.
             This should no longer be needed once DM-15409 is resolved.
@@ -250,7 +253,7 @@ def subtractCrosstalk(exposure, crosstalkCoeffs=None,
                       badPixels=["BAD"], minPixelToMask=45000,
                       crosstalkStr="CROSSTALK", isTrimmed=False,
                       backgroundMethod="None"):
-    """Subtract the intra-CCD crosstalk from an exposure
+    """Subtract the intra-detector crosstalk from an exposure
 
     We set the mask plane indicated by ``crosstalkStr`` in a target amplifier
     for pixels in a source amplifier that exceed `minPixelToMask`. Note that
@@ -359,8 +362,8 @@ def writeCrosstalkCoeffs(outputFileName, coeff, det=None, crosstalkName="Unknown
         Used to provide the list of amplifier names;
         if None use ['0', '1', ...]
     ccdType : `str`
-        Name of CCD, used to index the yaml file
-        If all CCDs are identical could be the type (e.g. ITL)
+        Name of detector, used to index the yaml file
+        If all detectors are identical could be the type (e.g. ITL)
     indent : `int`
         Indent width to use when writing the yaml file
     """
--- a/python/lsst/ip/isr/crosstalk.py
+++ b/python/lsst/ip/isr/crosstalk.py
@@ -22,11 +22,12 @@
 """
 Apply intra-CCD crosstalk corrections
 """
+import numpy as np
 
 import lsst.afw.math
 import lsst.afw.table
 import lsst.afw.detection
-from lsst.pex.config import Config, Field, ChoiceField
+from lsst.pex.config import Config, Field, ChoiceField, ListField
 from lsst.pipe.base import Task
 
 __all__ = ["CrosstalkConfig", "CrosstalkTask", "subtractCrosstalk", "writeCrosstalkCoeffs",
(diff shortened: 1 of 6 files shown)


python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -266,13 +225,81 @@ def run(self, exposure, dataId=None):
...
+        for ii, iAmp in enumerate(ccd):
+            iImage = mi[iAmp.getBBox()]
+            iMask = iImage.mask.array
+            select = (iMask & detected > 0) & (iMask & bad == 0) & np.isfinite(iImage.image.array)
+            for jj, jAmp in enumerate(ccd):
+                if ii == jj:
+                    continue
+                jImage = extractAmp(mi.image, jAmp, iAmp.getReadoutCorner(), isTrimmed=True)
+                ratios[jj][ii] = (jImage.array[select] - bg)/iImage.image.array[select]
+                self.debugPixels('pixels', iImage.image.array[select], jImage.array[select] - bg, ii, jj)
```
mfisherlevine: Again just curious about the debug config checking here.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -266,13 +225,81 @@ def run(self, exposure, dataId=None):
...
+
+        for ii, iAmp in enumerate(ccd):
+            iImage = mi[iAmp.getBBox()]
+            iMask = iImage.mask.array
+            select = (iMask & detected > 0) & (iMask & bad == 0) & np.isfinite(iImage.image.array)
+            for jj, jAmp in enumerate(ccd):
+                if ii == jj:
+                    continue
+                jImage = extractAmp(mi.image, jAmp, iAmp.getReadoutCorner(), isTrimmed=True)
+                ratios[jj][ii] = (jImage.array[select] - bg)/iImage.image.array[select]
```
mfisherlevine: I haven't really thought about this, but just checking you're sure you don't want to subtract the background from the other image. No need for a lengthy explanation, just wanted to check this was thought about/deliberate.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -316,17 +358,193 @@ def reduce(self, ratioList):
                 else:
                     result = np.concatenate([vv for vv in values if len(vv) > 0])
             ratios[ii][jj] = result
-        coeff, coeffErr, coeffNum = measureCrosstalkCoefficients(ratios, self.config.rejIter,
-                                                                 self.config.rejSigma)
+            self.debugRatios('reduce', ratios, ii, jj)
```
mfisherlevine: And again. I'm sure I must be missing what makes this work though, so sorry about all these.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -316,17 +358,193 @@ def reduce(self, ratioList):
...
+        coeff = np.zeros((numAmps, numAmps))
+        coeffErr = np.zeros((numAmps, numAmps))
+        coeffNum = np.zeros((numAmps, numAmps), dtype=int)
+
+        for ii, jj in itertools.product(range(numAmps), range(numAmps)):
+            if ii == jj:
+                values = [0.0]
+            else:
+                values = np.array(ratios[ii][jj])
+                values = values[np.abs(values) < 1.0]  # Discard unreasonable values
```
mfisherlevine: Interesting. I wonder if this is worth logging when it happens. Maybe not, just a thought. Also, while 1 is an obvious amount, I wonder if it might be better to have it as a config option? Again, feel free to ignore.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -316,17 +358,193 @@ def reduce(self, ratioList):
...
+
+            coeffNum[ii][jj] = len(values)
+
+            if len(values) == 0:
+                coeff[ii][jj] = np.nan
+                coeffErr[ii][jj] = np.nan
```
mfisherlevine: Huh... Again, these are interesting. I get why you don't just want to set it to zero, but is the crosstalk correction code nan-safe, _i.e._ is it expecting this? I think I like this, but again, think this really might be a place to chuck in a log, probably even at WARN level, that no values remained here, so crosstalk element will be NaN.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -316,17 +358,193 @@ def reduce(self, ratioList):
...
+            else:
+                if ii != jj:
+                    for rej in range(rejIter):
```
mfisherlevine: This is an interesting approach. Is there a reason for not using afwMath stats and a clipped mean there? That would automatically do the number of iterations you want, and I believe would use the actual standard deviation, rather than the approximation from the IQR as is being done here.

If that's not desirable, could you just update the docstring to note that the sigma-clipping is using the IQR (times a const) rather than the std dev itself please?

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -316,17 +358,193 @@ def reduce(self, ratioList):
...
+                        lo, med, hi = np.percentile(values, [25.0, 50.0, 75.0])
+                        sigma = 0.741*(hi - lo)
+                        good = np.abs(values - med) < rejSigma*sigma
+                        if good.sum() == len(good):
+                            break
+                        values = values[good]
+
+            coeff[ii][jj] = np.mean(values)
+            coeffErr[ii][jj] = np.nan if coeffNum[ii][jj] == 1 else np.std(values)
```
mfisherlevine: Confused by a couple of things here. You've tossed all the 1's already, so I don't think `coeffNum[ii][jj] == 1` can ever be True. Also, don't you mean zero anyway here?

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -316,17 +358,193 @@ def reduce(self, ratioList):
...
+            self.debugRatios('measure', ratios, ii, jj)
```
mfisherlevine: As always, my usual (probably wrong) comment about debug. Sorry.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -55,6 +56,76 @@ class CrosstalkConfig(Config):
...
+        RuntimeError
+            Raised if no coefficients could be generated from this detector/configuration.
+        """
+        if self.useConfigCoefficients is True:
+            coeffs = np.array(self.crosstalkValues).reshape(self.crosstalkShape)
+            if detector is not None:
+                nAmp = len(detector)
+                if coeffs.shape != (nAmp, nAmp):
+                    raise RuntimeError("Constructed crosstalk coeffients do not match detector shape. %s %s",
+                                       coeffs.shape, nAmp)
```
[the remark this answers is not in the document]
timj: or use f-strings...

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -225,13 +176,21 @@ def parseAndRun(cls, *args, **kwargs):
...
+        """Disable config output"""
+        return None
+
+    def _getMetadataName(self):
+        """Disable metdata output"""
+        return None
+
     def runDataRef(self, dataRef):
-        """Get crosstalk ratios for CCD
+        """Get crosstalk ratios for CCD.
```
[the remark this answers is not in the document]
czwa: I've changed all crosstalk related "CCD" to "detector".  I'd rather do it here while I'm thinking about it.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -266,13 +225,81 @@ def run(self, exposure, dataId=None):
...
+        if badPixels is None:
+            badPixels = list(self.config.badMask)
+
+        mi = exposure.getMaskedImage()
+        FootprintSet(mi, Threshold(threshold), "DETECTED")
+        detected = mi.getMask().getPlaneBitMask("DETECTED")
+        bad = mi.getMask().getPlaneBitMask(badPixels)
+        bg = calculateBackground(mi, badPixels + ["DETECTED"])
+
+        self.debugView('extract', exposure)
```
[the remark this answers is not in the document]
czwa: It should only fire if the script is run with `--debug`, and if the `lsstDebug.Info` object has been overridden to return true for `lsst.ip.isr.measureCrosstalk` and the `extract` frame/stepname.  I think this is the correct way to trigger debug code in the stack, as I based this code on the debug statements in the bias correction.  It seems to only trigger when I have `~czw/dev/DM-19214/debug.py` in my python path, so it should be safe.

python/lsst/ip/isr/measureCrosstalk.py
czwa: Yes, it's deliberate.

python/lsst/ip/isr/measureCrosstalk.py
czwa: I'm ignoring because I think this would be a case that could come up a lot.  If a bright star in amp A ends up on an even brighter star in amp B, we want to kick out that sample from amp A.

python/lsst/ip/isr/measureCrosstalk.py
czwa: I've copied a similar warning from elsewhere.  I didn't originally flag this as the measured values aren't automatically used anywhere (someone needs to add them to a detector/configuration).

python/lsst/ip/isr/measureCrosstalk.py
czwa: This is inherited code, but I believe the logic was that this is a noisy measurement, so using the Gaussian sigma approximation from the IQR gives an initial robust sigma estimate, allowing the clipping to be more aggressive.  I'll test the afwMath version to see if that's significantly different.

python/lsst/ip/isr/measureCrosstalk.py
czwa: `coeffNum[ii][jj]` is the number of samples for this entry in the CT matrix.  Setting the error to `np.nan` shows that this isn't well estimated, instead of `np.std() = 0.0`, which would indicate it's perfectly estimated.

python/lsst/ip/isr/measureCrosstalk.py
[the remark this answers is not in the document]
mfisherlevine: Ah, I think I get it now. Is it that those debug functions (`debugRatios()`, `debugView()` etc) always fire, but the `frame = getDebugFrame(self._display, stepname)` will return `None` if the debug hasn't been run with, and that all the subsequent code in those methods is inside an `if frame:` block? If so then I think that's fine.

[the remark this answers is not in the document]
mfisherlevine: I've seen other patterns for this which I find a little more intuitive (and also don't call the methods at all if debugs not being run with) but it's probably not worth changing.

python/lsst/ip/isr/measureCrosstalk.py
mfisherlevine: Meaning this is now logging a warning? If so then mark this as resolved. I do think that logging here is appropriate, as this is the task to measure the coefficients, so it failing for any elements should be announced loudly I think.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -219,7 +296,10 @@ def subtractCrosstalk(exposure, badPixels=["BAD"], minPixelToMask=45000,
 
     ccd = exposure.getDetector()
     numAmps = len(ccd)
-    coeffs = ccd.getCrosstalk()
+    if crosstalkCoeffs is None:
```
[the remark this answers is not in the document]
czwa: It's a legacy thing I don't want to break if I don't need to.  The previous version assumed coefficients could only be obtained by accessing the `ccd.getCrosstalk()` method.  This version allows it to be less restrictive, but since `subtractCrosstalk()` is independent, I didn't want that kind of call to cause problems.  I will admit that I can't find any issues with the new version from a scan of github.

python/lsst/ip/isr/isrMock.py
```diff
@@ -251,6 +253,11 @@ class IsrMock(pipeBase.Task):
     ConfigClass = IsrMockConfig
     _DefaultName = "isrMock"
 
+    XFLIP = {afwTable.LL: False, afwTable.LR: True,
```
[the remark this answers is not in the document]
czwa: I've directly imported them, as they should be uniquely defined.  The goal of the dicts is to allow horizontal/vertical flips to be determined by simply XORing the amplifier corner position from potential source/victim amplifiers.

python/lsst/ip/isr/measureCrosstalk.py
czwa: I've decided against switching to `afwMath` simply because there is enough statistics code in `ip_isr` that this should probably be a full ticket itself (to identify what is in `afwMath`, and see what, if anything, would need to be added).

mfisherlevine: OK, sounds good 👍

PR #98, opened 2019-05-31
DM-19874: Add support for decam illumcor calibration products

mrawls:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1356,6 +1403,12 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
 
         self.roughZeroPoint(ccdExposure)
 
+        if self.config.doIlluminationCorrection and filterName in self.config.illumFilters:
```
czwa: I'm not sure I agree with the placement here.  Since the illumination correction is "flat-like", I would have thought that it would go there.  Given that the filter sets for this and the fringe do not overlap, this likely doesn't matter for DECam.  It would still be good to at least move this prior to the interpolation step, as that will ensure that the pre- and post- interpolated image products have the same operations (again, likely not an issue for DECam processing).

mrawls: OK, I can move it to before interpolation. I put it near the end since that's where the DECam CP pipeline put it, but I'll move it and verify that doesn't change anything substantially. Thanks!

PR #99, opened 2019-06-12
DM-20109: Improve ip_isr log messages to be more explicit

czwa:
Clarify ISR log messages, comments, and docstrings.

Change isrFunctions.brighterFatterCorrection to return final
difference and iteration count, so an old log message can be restored.

Remove old out-of-date documentation that was causing build warnings.

### DIFF
--- a/python/lsst/ip/isr/fringe.py
+++ b/python/lsst/ip/isr/fringe.py
@@ -74,7 +74,7 @@ class FringeTask(Task):
     _DefaultName = 'isrFringe'
 
     def readFringes(self, dataRef, assembler=None):
-        """Read the fringe frame(s)
+        """Read the fringe frame(s), and pack data into a Struct
 
         The current implementation assumes only a single fringe frame and
         will have to be updated to support multi-mode fringe subtraction.
@@ -82,10 +82,23 @@ class FringeTask(Task):
         This implementation could be optimised by persisting the fringe
         positions and fluxes.
 
-        @param dataRef     Data reference for the science exposure
-        @param assembler   An instance of AssembleCcdTask (for assembling fringe frames)
-        @return Struct(fringes: fringe exposure or list of fringe exposures;
-                       seed: 32-bit uint derived from ccdExposureId for random number generator
+        Parameters
+        ----------
+        dataRef : `daf.butler.butlerSubset.ButlerDataRef`
+            Butler reference for the exposure that will have fringing
+            removed.
+        assembler : `lsst.ip.isr.AssembleCcdTask`, optional
+            An instance of AssembleCcdTask (for assembling fringe
+            frames).
+
+        Returns
+        -------
+        fringeData : `pipeBase.Struct`
+            Struct containing fringe data:
+            - ``fringes`` : `lsst.afw.image.Exposure` or `list` thereof
+                Calibration fringe files containing master fringe frames.
+            - ``seed`` : `int`, optional
+                Seed for random number generation.
         """
         try:
             fringe = dataRef.get("fringe", immediate=True)
@@ -108,14 +121,27 @@ class FringeTask(Task):
         Primary method of FringeTask.  Fringes are only subtracted if the
         science exposure has a filter listed in the configuration.
 
-        @param exposure    Science exposure from which to remove fringes
-        @param fringes     Exposure or list of Exposures
-        @param seed        32-bit unsigned integer for random number generator
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Science exposure from which to remove fringes.
+        fringes : `lsst.afw.image.Exposure` or `list` thereof
+            Calibration fringe files containing master fringe frames.
+        seed : `int`, optional
+            Seed for random number generation.
+
+        Returns
+        -------
+        solution : `np.array`
+            Fringe solution amplitudes for each input fringe frame.
+        rms : `float`
+            RMS error for the fit solution for this exposure.
         """
         import lsstDebug
         display = lsstDebug.Info(__name__).display
 
         if not self.checkFilter(exposure):
+            self.log.info("Filter not found in FringeTaskConfig.filters. Skipping fringe correction.")
             return
 
         if seed is None:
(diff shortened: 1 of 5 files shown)


python/lsst/ip/isr/fringe.py
```diff
@@ -74,18 +74,31 @@ class FringeTask(Task):
...
+            frames)
+
+        Returns
+        -------
+        fringeData : `pipeBase.Struct`
+            Struct containing fringe data:
+            - ``fringes`` : `lsst.afw.image.Exposure` or `list` thereof
+                Calibration fringe files containing master fringe frames.
+            - ``seed`` : `int`, optional
+                Seed for random number generation
```
taranu: Missing period.

python/lsst/ip/isr/fringe.py
```diff
@@ -74,18 +74,31 @@ class FringeTask(Task):
...
-        @return Struct(fringes: fringe exposure or list of fringe exposures;
-                       seed: 32-bit uint derived from ccdExposureId for random number generator
+        Parameters
+        ----------
+        dataRef : `daf.butler.butlerSubset.ButlerDataRef`
+            Butler reference for the exposure that will have fringing
+            removed.
+        assembler : `lsst.ip.isr.AssembleCcdTask`, optional
+            An instance of AssembleCcdTask (for assembling fringe
+            frames)
```
taranu: Missing period.

python/lsst/ip/isr/fringe.py
```diff
@@ -108,14 +121,27 @@ def run(self, exposure, fringes, seed=None):
...
-        @param fringes     Exposure or list of Exposures
-        @param seed        32-bit unsigned integer for random number generator
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Science exposure from which to remove fringes.
+        fringes : `lsst.afw.image.Exposure` or `list` thereof
+            Calibration fringe files containing master fringe frames.
+        seed : `int`, optional
+            Seed for random number generation
```
taranu: Missing period.

python/lsst/ip/isr/fringe.py
```diff
@@ -108,14 +121,27 @@ def run(self, exposure, fringes, seed=None):
...
         science exposure has a filter listed in the configuration.
 
-        @param exposure    Science exposure from which to remove fringes
-        @param fringes     Exposure or list of Exposures
-        @param seed        32-bit unsigned integer for random number generator
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Science exposure from which to remove fringes.
+        fringes : `lsst.afw.image.Exposure` or `list` thereof
```
taranu: Does it have to be a list or just an iterable? I didn't see iterable anywhere in the diffs.

python/lsst/ip/isr/fringe.py
```diff
@@ -148,25 +174,57 @@ def runDataRef(self, exposure, dataRef, assembler=None):
...
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Science exposure from which to remove fringes.
+        dataRef : `daf.persistence.butlerSubset.ButlerDataRef`
+            Butler reference to the exposure.  Used to find
+            appropriate fringe data.
+        assembler : `lsst.ip.isr.AssembleCcdTask`, optional
+            An instance of AssembleCcdTask (for assembling fringe
+            frames)
```
taranu: Period.

python/lsst/ip/isr/fringe.py
```diff
@@ -227,11 +310,23 @@ def measureExposure(self, exposure, positions, title="Fringe"):
 
     @timeMethod
     def solve(self, science, fringes):
-        """Solve (with iterative clipping) for the scale factors
-
-        @param science     Array of science exposure fringe amplitudes
-        @param fringes     Array of arrays of fringe frame fringe amplitudes
-        @return Array of scale factors for the fringe frames
+        """Solve (with iterative clipping) for the scale factors.
```
taranu: Not necessary but "Solve for the scale factors using iterative clipping" would be easier to read.

python/lsst/ip/isr/isrTask.py
```diff
@@ -904,9 +904,9 @@ def readIsrData(self, dataRef, rawExposure):
             - ``bfKernel``: Brighter-Fatter kernel (`numpy.ndarray`)
             - ``defects``: list of defects (`lsst.meas.algorithms.Defects`)
             - ``fringes``: `lsst.pipe.base.Struct` with components:
-                - ``fringes``: fringe calibration frame (`afw.image.Exposure`)
-                - ``seed``: random seed derived from the ccdExposureId for random
-                    number generator (`uint32`)
+              - ``fringes``: fringe calibration frame (`afw.image.Exposure`)
```
taranu: ``fringes``: `afw.image.Exposure`
Fringe calibration frame.
... would be more consistent, I think.

python/lsst/ip/isr/isrTask.py
```diff
@@ -904,9 +904,9 @@ def readIsrData(self, dataRef, rawExposure):
...
+              - ``seed``: random seed derived from the ccdExposureId for random
+                  number generator (`uint32`)
```
taranu: As above:
``seed``: `uint32`
Random seed derived from the ccdExposureId for number generator.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1211,31 +1214,31 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
                             qaMedian = qaStats.getValue(afwMath.MEDIAN)
                             qaStdev = qaStats.getValue(afwMath.STDEVCLIP)
 
-                        self.metadata.set("ISR OSCAN {} MEDIAN".format(amp.getName()), qaMedian)
-                        self.metadata.set("ISR OSCAN {} STDEV".format(amp.getName()), qaStdev)
+                        self.metadata.set(f"ISR OSCAN {amp.getName()} MEDIAN", qaMedian)
+                        self.metadata.set(f"ISR OSCAN {amp.getName()} STDEV", qaStdev)
```
taranu: I presume there's a good reason for the all caps but it is a bit loud.

python/lsst/ip/isr/fringe.py
czwa: It can probably be any iterable.  This value comes from readFringes, which in turn pulls it from the butler, and I've naively assumed that that would return a list of entries if multiple items were found.

python/lsst/ip/isr/isrTask.py
czwa: That's what I found copying other `metadata.set` calls.  I suspect there's an assumption that these may end up as fits header keywords (or `HIERARCH`s), and so upper case?

python/lsst/ip/isr/isrTask.py
czwa: This seems to be the way we document structs and dict-like objects.

PR #103, opened 2019-06-28
DM-20287: Use '%' strings instead of "f" strings for log messages.

czwa:


python/lsst/ip/isr/isrTask.py
```diff
@@ -839,7 +839,8 @@ def adaptArgsAndRun(self, inputData, inputDataIds, outputDataIds, butler):
         try:
             inputData['detectorNum'] = int(inputDataIds['ccdExposure']['detector'])
         except Exception as e:
-            raise ValueError(f"Failure to find valid detectorNum value for Dataset {inputDataIds}: {e}")
+            raise ValueError("Failure to find valid detectorNum value for Dataset %s: %s." %
+                             (inputDataIds, e))
```
PaulPrice: I think this is one instance where the `f`-string formatting would be OK. Don't mind either way.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1203,7 +1204,7 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
                 if self.config.doOverscan and not badAmp:
                     # Overscan correction on amp-by-amp basis.
                     overscanResults = self.overscanCorrection(ccdExposure, amp)
-                    self.log.debug(f"Corrected overscan for amplifier {amp.getName()}.")
+                    self.log.debug("Corrected overscan for amplifier %s." % (amp.getName()))
```
PaulPrice: This should be:

    self.log.debug("Corrected overscan for amplifier %s.", amp.getName())

This way, the stringification only occurs if the log threshold is sufficient, rather than being done always.

Here and for all other changes involving `self.log.whatever`.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1594,7 +1597,8 @@ def ensureExposure(self, inputExp, camera, detectorNum):
             # Assume this will be caught by the setup if it is a problem.
             return inputExp
         else:
-            raise TypeError(f"Input Exposure is not known type in isrTask.ensureExposure: {type(inputExp)}.")
+            raise TypeError("Input Exposure is not known type in isrTask.ensureExposure: %s." %
+                            (type(inputExp)))
```
PaulPrice: Another one that can stay as it was if you want.

python/lsst/ip/isr/isrTask.py
czwa: Do we have recommendations for when various methods are appropriate?  I have very little intuition on this topic.

PaulPrice: Perhaps @timj or @ktlim can make a suggestion?

timj: I don't understand why this needs to be moved back to `%` formatting as f-strings seem ideal for this one. I do have a preference for f-strings over `%` and `%` style is only required for logging.

jdswinbank: ...and as I understand it, the “`%` style” which is required for logging is not the same as the above (I assume `"some string %s: %s" % (foo, bar)` suffers from the same defect as f-strings, in that it will always be interpolated by Python regardless of the log level).

I think the message here is, per https://developer.lsst.io/stack/logging.html#basic-usage-in-python, that when calling a logger you don't use Python string interpolation *at all*, but rather leave the logger to do the interpolation for you based on arguments you provide.

timj: Yes, you always defer the formatting when logging.

python/lsst/ip/isr/isrTask.py
PaulPrice: Either drop the brackets around `type(inputExp)` or add a comma, `(type(inputExp),)`: as it is, the brackets don't add anything.

PR #104, opened 2019-06-28
DM-5081: Mark amplifier edges SUSPECT if brighter-fatter is applied.

czwa:
The edges of amplifiers that have had the brighter-fatter correction
applied may not have had the appropriate correction applied.  To
reflect this fact, ensure that the number of edge pixels masked is the
larger of IsrTask.config.numEdgeSuspect and the brighter-fatter kernel
size.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -252,11 +252,6 @@ class IsrTaskConfig(pexConfig.Config):
         doc="Name of mask plane to use for suspect pixels",
         default="SUSPECT",
     )
-    numEdgeSuspect = pexConfig.Field(
-        dtype=int,
-        doc="Number of edge pixels to be flagged as untrustworthy.",
-        default=0,
-    )
 
     # Initial masking options.
     doSetBadRegions = pexConfig.Field(
@@ -1277,6 +1272,11 @@ class IsrTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
             self.log.info("Masking defects.")
             self.maskDefect(ccdExposure, defects)
 
+            if self.config.numEdgeSuspect > 0:
+                self.log.info("Masking edges as SUSPECT.")
+                self.maskEdges(ccdExposure, numEdgePixels=self.config.numEdgeSuspect,
+                               maskPlane="SUSPECT")
+
         if self.config.doNanMasking:
             self.log.info("Masking NAN value pixels.")
             self.maskNan(ccdExposure)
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -1187,6 +1187,17 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
...
 
+        minEdgeSuspectMask = 0
+        if self.config.doBrighterFatter is True:
+            # Applying the brighter-fatter correction applies a
+            # convolution to the science image. At the edges this
+            # convolution may not have sufficient valid pixels to
+            # produce a valid correction. Mark pixels within the size
+            # of the brighter-fatter kernel as SUSPECT to warn of this
+            # fact.
+            minEdgeSuspectMask = numpy.max(bfKernel.shape)
```
jdswinbank: Should this be halved? Ref https://github.com/lsst/ip_isr/blob/master/python/lsst/ip/isr/isrFunctions.py#L799

python/lsst/ip/isr/isrTask.py
```diff
@@ -2068,7 +2080,7 @@ def suspectDetection(self, exposure, amp):
             maskName=self.config.suspectMaskName,
         )
 
-    def maskDefect(self, exposure, defectBaseList):
+    def maskDefect(self, exposure, defectBaseList, numEdgeSuspect=0):
```
jdswinbank: The docstring on this method is wrong. It was wrong before you started (since it was already setting `SUSPECT` as well as `BAD`), but it's even more wrong now. Could you take this opportunity to fix it?

jdswinbank: Also consider renaming the method, since it's clearly doing more than masking defects.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2091,10 +2105,11 @@ def maskDefect(self, exposure, defectBaseList):
             defectList = defectBaseList
         defectList.maskPixels(maskedImage, maskName="BAD")
 
-        if self.config.numEdgeSuspect > 0:
+        numEdgeSuspect = numpy.max([numEdgeSuspect, self.config.numEdgeSuspect])
```
jdswinbank: Do we want to do this calculation here as well as at line 1301, above?

At line 1301, it seems appropriate. Here it's saying that even if I call this particular method with a very specific value of `numEdgeSuspect` for my own reasons, the code might just ignore it and use the task config instead. That seems like it could be unfortunate.

I'd suggest that if you want to leave this behaviour in, you might want to change the docstring above to indicate that `numEdgeSuspect` can be ignored in some circumstances.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2112,7 +2127,7 @@ def maskAndInterpolateDefects(self, exposure, defectBaseList):
         defectBaseList : `List` of `Defects`
 
         """
-        self.maskDefects(exposure, defectBaseList)
+        self.maskDefect(exposure, defectBaseList)
```
jdswinbank: ...how did this work before? *Did* this work before?

python/lsst/ip/isr/isrTask.py
czwa: I've split edge masking into a separate method, and made sure that IsrTask is consistent about calling it.  This also simplifies the main goal of the ticket, as there's a clear way to apply edge masks.

python/lsst/ip/isr/isrTask.py
czwa: Moving to a stand-alone method removes this multiple calculation.

python/lsst/ip/isr/isrTask.py
czwa: I don't have a clear answer.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2105,7 +2130,9 @@ def maskAndInterpolateDefects(self, exposure, defectBaseList):
         defectBaseList : `List` of `Defects`
```
timj: Pedantic point but I think this is:
```
`list` of `~lsst.afw.image.DefectBase`
```
since `Defects` is going to be `~lsst.meas.algorithms.Defects` which is itself a collection of `lsst.meas.algorithms.Defect`.  In general it's better to use `Defects` right from the start if at all possible.

czwa: Fixed to match the `maskAndInterpolate` method, and also fixed to remove interpolation mention in the not-interpolating method.

PR #105, opened 2019-06-28
DM-20154: Implement new initial WCS design

parejkoj:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1377,10 +1377,6 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
                                                  sensorTransmission=sensorTransmission,
                                                  atmosphereTransmission=atmosphereTransmission)
 
-        if self.config.doAddDistortionModel:
```
mrawls: Would it be reasonable to leave this "if" in here and log a deprecation warning? Otherwise I worry WCSs will subtly change immediately and nobody will see the deprecation warning until the functionality silently disappears in v19.

parejkoj: The problem is that if somebody tries to use it, it would result in applying the distortion terms twice (once when the raw is read in, and again here), which would result in nonsense. I could make it an explicit failure instead, if that would help people catch the necessary change to their `Config`s?

mrawls: I don't think having just `if self.config.doAddDistortionModel: log.loglog.logface.warn('What are you doing stop it')` would apply anything? That's all I'm suggesting. If I'm wrong, an explicit failure would be a good alternative. I tend to assume people add things to their configs and then forget about them indefinitely unless something breaks.

parejkoj: Note that `doAddDistortionModel` already will issue a warning if it is set in a Config because it is using the new `Config` deprecation system: https://developer.lsst.io/stack/deprecating-interfaces.html#config-deprecation

PR #111, opened 2019-09-13
Tickets/dm 13440

leannep:
I believe I have addressed all issues in the original PR/review request from 18 months ago


tests/test_flatAndIlluminationCorrection.py
```diff
@@ -97,29 +102,27 @@ def testIllum3(self):
...
         level = 10
-        readNoise = 1
+        readNoise = 1.5
         raw.image.set(level)
 
         amp = detector[0]
         amp.setReadNoise(readNoise)
 
-        for gain in [-1, 0, 0.1, 1]:
+        for gain in [-1, 0, 0.1, 1, float("NaN")]:
```
RobertLuptonTheGood: Might be better to use `math.nan` (or use numpy not math)

leannep: Fixed.  I followed the other examples in this package that do this, but you are right.

PR #115, opened 2019-11-05
DM-22108: Fix gen3 handling of defects.

czwa:
This should resolve KeyError exceptions when the inputs hash does not
contain defects.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -801,16 +801,17 @@ class IsrTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
         inputs['isGen3'] = True
 
         if self.config.doLinearize is True:
-            if 'linearizer' not in inputs.keys():
+            if 'linearizer' not in inputs:
                 detector = inputs['ccdExposure'].getDetector()
                 linearityName = detector.getAmplifiers()[0].getLinearityType()
                 inputs['linearizer'] = linearize.getLinearityTypeByName(linearityName)()
 
-        if inputs['defects'] is not None:
-            # defects is loaded as a BaseCatalog with columns x0, y0, width, height.
-            # masking expects a list of defects defined by their bounding box
-            if not isinstance(inputs["defects"], Defects):
-                inputs["defects"] = Defects.fromTable(inputs["defects"])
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -806,11 +806,14 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
                 linearityName = detector.getAmplifiers()[0].getLinearityType()
                 inputs['linearizer'] = linearize.getLinearityTypeByName(linearityName)()
 
-        if inputs['defects'] is not None:
-            # defects is loaded as a BaseCatalog with columns x0, y0, width, height.
-            # masking expects a list of defects defined by their bounding box
-            if not isinstance(inputs["defects"], Defects):
-                inputs["defects"] = Defects.fromTable(inputs["defects"])
+        if self.config.doDefects is True:
+            if "defects" not in inputs.keys():
```
timj: I don't think `.keys()` is needed here.

czwa:
Jenkins noted that there is no `doDefects` and that it is `doDefect`.

python/lsst/ip/isr/isrTask.py
```diff
@@ -801,16 +801,19 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
 
-        if inputs['defects'] is not None:
-            # defects is loaded as a BaseCatalog with columns x0, y0, width, height.
-            # masking expects a list of defects defined by their bounding box
-            if not isinstance(inputs["defects"], Defects):
-                inputs["defects"] = Defects.fromTable(inputs["defects"])
+        if self.config.doDefect is True:
+            if "defects" not in inputs:
+                inputs['defects'] = None
+            if inputs['defects'] is not None:
```
timj: Are both `if` checks needed?
```python
if "defects" in inputs and inputs["defect"] is not None:
```
saves us from doing an extra assignment?

plazas:
The changes look good. The only question I had is, if the user sets self.config.doDefects=True but there are no defect list, then inputs['defects'] = None, should it complain or raise or warn or say something like “you told me to do defects, but I can’t find any defects list”? Or is this something that happens later?. Chris has checked that indeed this is handled later along the way.

PR #116, opened 2019-11-08
DM-21739: Enable and test fringe/strayLight in gen3

czwa:


python/lsst/ip/isr/isrTask.py
```diff
@@ -846,12 +845,12 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
             # However, these are packed oddly, so replace the data
             # reference with a URI that can be handled by the
             # straylight correction.  This breaks IO purity.
-            straylightRef = inputs['strayLightData']
-            diskUri = straylightRef.getUri()
-            inputs['strayLightData'] = diskUri
+            # straylightRef = inputs['strayLightData']
+            # diskUri = straylightRef.getUri()
+            # inputs['strayLightData'] = diskUri
+            print("Recieved straylightData: ", inputs['strayLightData'])
```
natelust: Probably remove this debug print and put back in the comments?

python/lsst/ip/isr/fringe.py
```diff
@@ -104,14 +104,41 @@ def readFringes(self, dataRef, assembler=None):
...
+            - ``fringes`` : `lsst.afw.image.Exposure` or `list` thereof
+                Calibration fringe files containing master fringe frames.
+            - ``seed`` : `int`, optional
+                Seed for random number generation.
+        """
         if assembler is not None:
-            fringe = assembler.assembleCcd(fringe)
+            fringeExp = assembler.assembleCcd(fringeExp)
+
+        seed = self.config.stats.rngSeedOffset + 12345678
```
natelust: why this magic number, if no reason add a comment. If there is a reason, add a comment

python/lsst/ip/isr/isrTask.py
```diff
@@ -821,12 +833,25 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+
+        if self.config.doStrayLight is True and self.strayLight.checkFilter(inputs['ccdExposure']):
+            # This holds a deferred reference to the straylight file.
+            # However, these are packed oddly, so replace the data
+            # reference with a URI that can be handled by the
+            # straylight correction.  This breaks IO purity.
+            # straylightRef = inputs['strayLightData']
+            # diskUri = straylightRef.getUri()
+            # inputs['strayLightData'] = diskUri
+            print("Recieved straylightData: ", inputs['strayLightData'])
```
natelust: I think you dont want to print here, but add it to inputs?

tests/test_fringes.py
```diff
@@ -287,8 +287,8 @@ def test_multiFringes(self):
         stdAfter = np.nanstd(exp.getImage().getArray())
 
         self.assertLess(medianAfter, medianBefore)
-        self.assertFloatsAlmostEqual(medianAfter, 3002.233, atol=1e-4)
-        self.assertFloatsAlmostEqual(stdAfter, 3549.9375, atol=1e-4)
+        self.assertFloatsAlmostEqual(medianAfter, 3001.312, atol=1e-4)
+        self.assertFloatsAlmostEqual(stdAfter, 3549.949, atol=1e-4)
```
natelust: What did these values change?

python/lsst/ip/isr/isrTask.py
czwa: I'm rewriting this, as butler now handles StrayLightData objects.  This mitigates the need for the getUri, as inputs['strayLightData'] now contains the object itself.

python/lsst/ip/isr/isrTask.py
czwa: As above, rewritten.  inputs['strayLightData'] should get what it says from the butler, and if not, rely on the straylight task to raise if it can't handle things.

tests/test_fringes.py
czwa: These values are related to how the fringe seed value is set.  The tolerance likely should be smaller to more accurately represent the error in this measurement due to changes in seed, but I'm not going to worry about it on this ticket.

python/lsst/ip/isr/fringe.py
czwa: Updated to try to use the exposureId as in gen2 instead of a magic number.

PR #118, opened 2019-12-17
DM-22647: Fix ip_isr.isrMock bug with CameraWrapper() args.

czwa:
Pass arguments to CameraWrapper correctly.

This changeset reproduces the current unit test values, while ensuring that the camera is generated correctly.

### DIFF
--- a/python/lsst/ip/isr/isrMock.py
+++ b/python/lsst/ip/isr/isrMock.py
@@ -52,8 +52,18 @@ class IsrMockConfig(pexConfig.Config):
     # Detector parameters. "Exposure" parameters.
     isLsstLike = pexConfig.Field(
         dtype=bool,
-        default=True,
+        default=False,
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrMock.py
```diff
@@ -52,9 +52,19 @@ class IsrMockConfig(pexConfig.Config):
     # Detector parameters. "Exposure" parameters.
     isLsstLike = pexConfig.Field(
         dtype=bool,
-        default=True,
+        default=False,
```
parejkoj: Why change the default to False here? Doesn't this mean that places that use isrMock will have to set this back to True if they wan't the behavior to be unchanged?

czwa: This change is to match the previous behavior.  As you mentioned, the isLsstLike option was being eaten by the plateScale, so in all previous calls to IsrMockConfig, the object returned had isLsstLike=False.

PR #120, opened 2019-12-18
Fix bitrot for using new style brighter-fatter kernels

mfisherlevine:
Remove unused config option for specifying files

### DIFF
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -647,7 +647,7 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
     return Struct(imageFit=offImage, overscanFit=overscanFit, overscanImage=overscanImage)
 
 
-def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain):
+def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain, gains=None):
     """Apply brighter fatter correction in place for the image.
 
     Parameters
@@ -665,6 +665,9 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain):
     applyGain : `Bool`
         If True, then the exposure values are scaled by the gain prior
         to correction.
+    gains : `dict` [`str`, `float`]
+        A dictionary, keyed by amplifier name, of the gains to use.
+        If gains is None, the nominal gains in the amplifier object are used.
 
     Returns
     -------
@@ -705,7 +708,7 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain):
     image = exposure.getMaskedImage().getImage()
 
     # The image needs to be units of electrons/holes
-    with gainContext(exposure, image, applyGain):
+    with gainContext(exposure, image, applyGain, gains):
 
         kLx = numpy.shape(kernel)[0]
         kLy = numpy.shape(kernel)[1]
@@ -767,7 +770,7 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain):
 
 
 @contextmanager
-def gainContext(exp, image, apply):
+def gainContext(exp, image, apply, gains=None):
     """Context manager that applies and removes gain.
 
     Parameters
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -665,6 +665,9 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain):
     applyGain : `Bool`
         If True, then the exposure values are scaled by the gain prior
         to correction.
+    gains : `dict` of `float`
```
czwa: I think the recommendation for dict documentation has changed:
https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-dict

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -778,17 +781,32 @@ def gainContext(exp, image, apply):
         Image to apply/remove gain.
     apply : `Bool`
         If True, apply and remove the amplifier gain.
+    gains : `dict` of `float`
```
czwa: Ditto.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -778,17 +781,32 @@ def gainContext(exp, image, apply):
...
+        If gains is None, the nominal gains in the amplifier object are used.
 
     Yields
     ------
     exp : `lsst.afw.image.Exposure`
         Exposure with the gain applied.
     """
+    # check we have all of them if provided because mixing and matching would
+    # be a real mess
+    if gains:
```
czwa: Should this check if that both `apply is True and gains`?  That might be excessively pedantic.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -797,6 +815,10 @@ def gainContext(exp, image, apply):
             ccd = exp.getDetector()
             for amp in ccd:
                 sim = image.Factory(image, amp.getBBox())
+                if gains:
+                    gain = gains[amp.getName()]
+                else:
+                    gain = amp.getGain()
                 sim /= amp.getGain()
```
czwa: `sim /= gain`

python/lsst/ip/isr/isrTask.py
```diff
@@ -918,22 +913,31 @@ def readIsrData(self, dataRef, rawExposure):
                         if self.config.doFlat else None)
 
         brighterFatterKernel = None
+        brighterFatterGains = None
         if self.config.doBrighterFatter is True:
-
-            # Use the new-style cp_pipe version of the kernel is it exists.
             try:
+                # Use the new-style cp_pipe version of the kernel is it exists
```
czwa: Typo.

python/lsst/ip/isr/isrTask.py
```diff
@@ -918,22 +913,31 @@ def readIsrData(self, dataRef, rawExposure):
...
                 except NoResults:
                     brighterFatterKernel = None
             if brighterFatterKernel is not None and not isinstance(brighterFatterKernel, numpy.ndarray):
-                # If the kernel is not an ndarray, it's the cp_pipe version, so extract the kernel for
-                # this detector, or raise an error.
+                # If the kernel is not an ndarray, it's the cp_pipe version
+                # so extract the kernel for this detector, or raise an error
                 if self.config.brighterFatterLevel == 'DETECTOR':
-                    brighterFatterKernel = brighterFatterKernel.kernel[ccd.getId()]
+                    if brighterFatterKernel.detectorKernel:
```
czwa: Have the format of the new-style kernels changed?  This seems to be using different attributes.

python/lsst/ip/isr/isrTask.py
```diff
@@ -918,22 +913,31 @@ def readIsrData(self, dataRef, rawExposure):
...
+                        brighterFatterKernel = brighterFatterKernel.detectorKernel[ccd.getId()]
+                    elif brighterFatterKernel.detectorKernelFromAmpKernels:
+                        brighterFatterKernel = brighterFatterKernel.detectorKernelFromAmpKernels[ccd.getId()]
+                    else:
+                        raise RuntimeError("Failed to extract kernel from new-style BF kernel.")
                 else:
                     # TODO DM-15631 for implementing this
                     raise NotImplementedError("Per-amplifier brighter-fatter correction not implemented")
```
czwa: Can you duplicate this logic into the gen3 runQuantum() method (line isrTask.py:827)?  The kernel object should be supplied as inputs['bfKernel'], and can be overwritten in place by the appropriate detectorKernel (the linearizer gets and uses a detector as well).  Adding the inputs['bfGains'] entry should ensure this works cleanly in that case.

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Very nice, I didn't know that, thanks!

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Yeah, fair enough, I can't really see a downside to that.

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Awesome 👏 _This_ is what reviews are all about. That would have been a nightmare to find if you hadn't caught it!

python/lsst/ip/isr/isrTask.py
mfisherlevine: Yup, exactly - that's the bitrot referred to in the ticket title (everything else is just scope-creep while I was in there in fact 😳)

python/lsst/ip/isr/isrTask.py
mfisherlevine: I was about to try, but I have no way of testing what I've done as I don't have any Gen3 repos (or even knowledge yet), and combined with no experience in such things, I think it's best left. I'll ticket, leave a `TODO:` and assign it to you. Sorry and thanks!

mfisherlevine: DM-22776

PR #123, opened 2020-01-17
DM-23083: Update large masks for BF convolution issues

czwa:
Grow large masks after BF correction.
Large masks should be treated like edges in the mask plane.  To do so,
convolve the important problem masks (those that are interpolated) by
a top-hat the size of the BF kernel, and add pixels to the mask if
they are calculated using a significant fraction of masked pixels.

### DIFF
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -134,6 +134,27 @@ def makeThresholdMask(maskedImage, threshold, growFootprints=1, maskName='SAT'):
     return measAlg.Defects.fromFootprintList(fpList)
 
 
+def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
+    """Grow a mask by an amount and add to the requested plane.
+
+    Parameters
+    ----------
+    mask : `lsst.afw.image.Mask`
+        Mask image to process.
+    radius : scalar
+        Amount to grow the mask.
+    maskNameList : `str` or `list` [`str`]
+        Mask names that should be grown.
+    maskValue : `str`
+        Mask plane to assign the newly masked pixels to.
+    """
+    if radius > 0:
+        thresh = afwDetection.Threshold(mask.getPlaneBitMask(maskNameList), afwDetection.Threshold.BITMASK)
+        fpSet = afwDetection.FootprintSet(mask, thresh)
+        fpSet = afwDetection.FootprintSet(fpSet, rGrow=radius, isotropic=False)
+        fpSet.setMask(mask, maskValue)
+
+
 def interpolateFromMask(maskedImage, fwhm, growSaturatedFootprints=1,
                         maskNameList=['SAT'], fallbackValue=None):
     """Interpolate over defects identified by a particular set of mask planes.
@@ -154,12 +175,9 @@ def interpolateFromMask(maskedImage, fwhm, growSaturatedFootprints=1,
     mask = maskedImage.getMask()
 
     if growSaturatedFootprints > 0 and "SAT" in maskNameList:
-        thresh = afwDetection.Threshold(mask.getPlaneBitMask("SAT"), afwDetection.Threshold.BITMASK)
-        fpSet = afwDetection.FootprintSet(mask, thresh)
         # If we are interpolating over an area larger than the original masked region, we need
         # to expand the original mask bit to the full area to explain why we interpolated there.
-        fpSet = afwDetection.FootprintSet(fpSet, rGrow=growSaturatedFootprints, isotropic=False)
-        fpSet.setMask(mask, "SAT")
+        growMasks(mask, radius=growSaturatedFootprints, maskNameList=['SAT'], maskValue="SAT")
 
     thresh = afwDetection.Threshold(mask.getPlaneBitMask(maskNameList), afwDetection.Threshold.BITMASK)
     fpSet = afwDetection.FootprintSet(mask, thresh)
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -521,6 +521,12 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         default=True,
         doc="Should the gain be applied when applying the brighter fatter correction?"
     )
+    brighterFatterMaskGrowSize = pexConfig.Field(
+        dtype=int,
+        default=0,
+        doc="Number of pixels to grow the masks listed in config.maskListToInterpolate "
+        " when brighter-fatter correction is applied."
+    )
 
     # Dark subtraction.
     doDark = pexConfig.Field(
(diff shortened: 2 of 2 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -2127,6 +2133,42 @@ def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT"):
...
 
+    def maskGrowTophat(self, exposure, boxRadius=0, boxThreshold=0.25, maskPlane="BAD", maskLabel="BAD"):
+        """Mask regions around large masked regions.
+
+        Parameters
+        ----------
+        exposure: `lsst.afw.image.Exposure`
+            Exposure to process.
+        boxRadius : `int`, optional
+            Half-size of the box to use to grow masks.
```
yalsayyad: How does this manifest in terms of # of pixels the original mask is "grown" by?

python/lsst/ip/isr/isrTask.py
```diff
@@ -2127,6 +2133,42 @@ def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT"):
...
+        boxThreshold : `float`, optional
```
yalsayyad: Do you think `boxThreshold` seems like it should be a new IsrTask config parameter?

python/lsst/ip/isr/isrTask.py
```diff
@@ -2127,6 +2133,42 @@ def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT"):
...
+        """
+        if boxRadius > 0 and boxThreshold > 0.0:
+            maskedImage = exposure.getMaskedImage()
+            maskBitMask = maskedImage.getMask().getPlaneBitMask(maskPlane)
+
+            thresholdImage = afwImage.MaskedImageF(maskedImage.getDimensions())
+            thresholdImage.getImage().getArray()[:, :] = maskedImage.getMask().getArray()[:, :] & maskBitMask
+
+            kernelImage = afwImage.ImageD(2 * boxRadius, 2 * boxRadius)
+            kernelImage.getArray()[:, :] = 1
```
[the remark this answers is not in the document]
yalsayyad: You could go with `kernelImage = afwImage.ImageD(2*boxRadius + 1, 2*boxRadius + 1 , 1)` and get back a kernel with the same shape as the BF kernel, which will be odd.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2127,6 +2133,42 @@ def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT"):
...
+
+            convControl = afwMath.ConvolutionControl(False, True, 0)
+            convolvedImage = afwImage.MaskedImageF(maskedImage.getDimensions())
+            afwMath.convolve(convolvedImage, thresholdImage, afwMath.FixedKernel(kernelImage), convControl)
```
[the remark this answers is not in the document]
czwa: I've taken the above suggestion to use (2*radius + 1) instead of my initial thought to simply accept the input kernel shape.  Although all the examples I can find have odd dimensions, this guarantees that for any future kernel.

python/lsst/ip/isr/isrTask.py
czwa: In a somewhat complicated way that is probably unintuitive.  Because of this, I've taken the advice from the ticket comments, and switched to simply allowing a list of mask plane names to be grown and assigned to a new value.  This duplicates code that was used for saturation mask growing, so I've split it out.  This is faster, and provides a better expectation for the output.

python/lsst/ip/isr/isrTask.py
czwa: The new mask growth radius is used, IsrConfig.brighterFatterGrowMask.

czwa: IsrConfig.brighterFatterMaskGrowSize is the new config option, to be more explicit what it does.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -154,12 +175,13 @@ def interpolateFromMask(maskedImage, fwhm, growSaturatedFootprints=1,
...
 
     if growSaturatedFootprints > 0 and "SAT" in maskNameList:
-        thresh = afwDetection.Threshold(mask.getPlaneBitMask("SAT"), afwDetection.Threshold.BITMASK)
-        fpSet = afwDetection.FootprintSet(mask, thresh)
-        # If we are interpolating over an area larger than the original masked region, we need
-        # to expand the original mask bit to the full area to explain why we interpolated there.
-        fpSet = afwDetection.FootprintSet(fpSet, rGrow=growSaturatedFootprints, isotropic=False)
-        fpSet.setMask(mask, "SAT")
+        growMasks(mask, radius=growSaturatedFootprints, maskNameList=['SAT'], maskValue="SAT")
+        # thresh = afwDetection.Threshold(mask.getPlaneBitMask("SAT"), afwDetection.Threshold.BITMASK)
```
yalsayyad: Leave the original inline comment, but remove the commented out code.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2127,6 +2141,43 @@ def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT"):
                 maskBitMask
             )
 
+    def maskGrowTophat(self, exposure, boxRadius=0, boxThreshold=0.25, maskPlane="BAD", maskLabel="BAD"):
```
yalsayyad: Delete the definition of  `maskGrowTophat`. It'll bitrot without a unit test or use anywhere

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -134,6 +134,27 @@ def makeThresholdMask(maskedImage, threshold, growFootprints=1, maskName='SAT'):
...
+def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
+    """Grow a mask by an amount and add to the BAD plane.
+
+    Parameters
+    ----------
+    mask : `lsst.afw.image.Mask`
+        Mask image to process.
+    radius : scalar
+        Amount to grow the mask.
+    maskNameList : `list` [`str`]
```
yalsayyad: This isn't correct anymore since you no longer pass in a list here:
```
                    isrFunctions.growMasks(ccdExposure.getMask(),
                                           radius=self.config.brighterFatterMaskGrowSize,
                                           maskNameList=maskPlane,
                                           maskValue=maskPlane)
```
 You're passing in a single string.  The current interface isn't bad, but worth thinking about whether growing a maskNameList with a single maskValue is the most reusable.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -134,6 +134,27 @@ def makeThresholdMask(maskedImage, threshold, growFootprints=1, maskName='SAT'):
     return measAlg.Defects.fromFootprintList(fpList)
 
 
+def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
+    """Grow a mask by an amount and add to the BAD plane.
```
yalsayyad: This docstring isn't correct after the behavior change. BAD is just your default value.

python/lsst/ip/isr/isrFunctions.py
czwa: growMasks uses the mask.getPlaneBitMask() method, which accepts vectors of strings or single strings, and returns the appropriate value.  I've updated the docstring to note both `str` and `list` [`str`] are valid types.

PR #124, opened 2020-01-28
DM-22776: Update Gen3 isrTask for BF corrections

czwa:
Update isrTask runQuantum for new-style BF correction.  This requires examining the object returned from the butler and reworking it to match the expectation of IsrTask().run().

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -840,6 +840,23 @@ class IsrTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
         # fixed for non-HSC cameras in the future.
         # inputs['crosstalkSources'] = (self.crosstalk.prepCrosstalk(inputsIds['ccdExposure'])
         #                        if self.config.doCrosstalk else None)
+        if self.config.doBrighterFatter:
+            brighterFatterKernel = inputs.get('bfKernel', None)
+            if brighterFatterKernel is not None and not isinstance(brighterFatterKernel, numpy.ndarray):
+                inputs['bfGains'] = brighterFatterKernel.gain
+                # If the kernel is not an ndarray, it's the cp_pipe version
+                # so extract the kernel for this detector, or raise an error
+                if self.config.brighterFatterLevel == 'DETECTOR':
+                    ccd = inputs['ccdExposure']
+                    if brighterFatterKernel.detectorKernel:
+                        inputs['bfKernel'] = brighterFatterKernel.detectorKernel[ccd.getId()]
+                    elif brighterFatterKernel.detectorKernelFromAmpKernels:
+                        inputs['bfKernel'] = brighterFatterKernel.detectorKernelFromAmpKernels[ccd.getId()]
+                    else:
+                        raise RuntimeError("Failed to extract kernel from new-style BF kernel.")
+                else:
+                    # TODO DM-15631 for implementing this
+                    raise NotImplementedError("Per-amplifier brighter-fatter correction not implemented")
 
         if self.config.doFringe is True and self.fringe.checkFilter(inputs['ccdExposure']):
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -831,9 +838,28 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
                     inputs["defects"] = Defects.fromTable(inputs["defects"])
 
-        # TODO: DM-22776 add retrieval for brighter-fatter kernel for Gen3
-        # if we can get HSC to use a new kernel, or just translate the old
-        # one to the new format we could drop the whole new/old style support
+        # Load the correct style of brighter fatter kernel, and repack
+        # the information as a numpy array.
+        if self.config.doBrighterFatter:
+            brighterFatterKernel = inputs.pop('newBFKernel', None)
+            if brighterFatterKernel is None:
```
natelust: why do this line, the if statement below will never be triggered by the bfKernel as it is an array

python/lsst/ip/isr/isrTask.py
```diff
@@ -831,9 +838,28 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+                brighterFatterKernel = inputs.get('bfKernel', None)
+
+            if brighterFatterKernel is not None and not isinstance(brighterFatterKernel, numpy.ndarray):
+                detId = detector.getId()
+                inputs['bfGains'] = brighterFatterKernel.gain
```
natelust: bfGains does not seem to be populated if not newBFKernel, is that what you want?

python/lsst/ip/isr/isrTask.py
```diff
@@ -943,7 +969,7 @@ def readIsrData(self, dataRef, rawExposure):
                 # If using a new-style kernel, always use the self-consistent
                 # gains, i.e. the ones inside the kernel object itself
                 brighterFatterKernel = dataRef.get("brighterFatterKernel")
-                brighterFatterGains = brighterFatterKernel.gain
+                brighterFatterGains = getattr(brighterFatterKernel, 'gain', None)
```
natelust: not that its bad, but why getattr here? if it is a new style then gain will be an attribute, if its an old style an exception will be thrown (likely by the dataRef.get) and this block wont be executed. I would be more worried about getting this far if it is an old style. If you wanted to be safe, then maybe a different except block? This seems like it will just hide a problem

python/lsst/ip/isr/isrTask.py
czwa: If a newBFKernel exists in the inputs, then we should use that (and do the block beginning on line 848 to pull the kernel for the detector we have).  Otherwise, we should use the old style, which does not need to do the L848 block (and inputs['bfKernel'] remains what we received from the butler).

python/lsst/ip/isr/isrTask.py
czwa: That is correct.  bfGains will not be in inputs, will then be None inside run(), and will be passed as such to the correction code (which handles the None case by using camera gain values).

python/lsst/ip/isr/isrTask.py
czwa: There is a possibility that there will be no .gains attribute, depending on how the kernel was generated.  I will ask @mfisherlevine if this is still possible.

mfisherlevine: In short, yes, Chris is right, that _is_ technically possible. However, I think that's a bug. The generating code always has them, and downstream code should be able to rely upon them being stored in the kernel so the gains can be applied in a consistent manner, so I think the best course of action here is to allow yourselves to rely up this, and ticket fixing up the BF code so it always stores the gain. The patch should be trivial. Does that sound OK?

czwa: DM-23273 requests that gain existence be enforced.  Marking this done, and will merge after testing.

PR #125, opened 2020-02-05
DM-23238: Add per row overscan removal

plazas:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -546,10 +560,30 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+            for row in biasMaskedArr:
+                if row.mask is False:
+                    row.mask = numpy.repeat(False, len(row))
+                medianRow = afwMath.makeStatistics(row, row.mask, fitTypeStats, statControl).getValue()
+                collapsed.append(medianRow)
+            collapsed = numpy.array(collapsed)
+            # Still need mask: use numpy to obtain the collapsed mask from biasMaskedArr
+            a = numpy.median(biasMaskedArr, axis=1)
+            collapsed = numpy.ma.masked_array(collapsed, mask=a.mask)
+            fitBiasArr = collapsed
```
mfisherlevine: I'm not quite sure why this whole block doesn't like in the big block above, but maybe I'm missing something.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -546,10 +560,30 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
-        collapsed = numpy.mean(biasMaskedArr, axis=1)
-        if collapsed.mask.sum() > 0:
-            collapsed.data[collapsed.mask] = numpy.mean(biasArray.data[collapsed.mask], axis=1)
+
+        if fitType in ('MEDIAN_PER_ROW',):
+            collapsed = []
+            fitTypeStats = afwMath.stringToStatisticsProperty('MEDIAN')
+            # Loop over rows because there is no afwMath.Median(array, axis=1)
+            for row in biasMaskedArr:
+                if row.mask is False:
```
mfisherlevine: I need to go in with a debugger, but is that _ever_ `True`? Isn't `row.mask` the mask array for a whole row? If so, that never `is` False.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -546,10 +560,30 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+                    row.mask = numpy.repeat(False, len(row))
```
mfisherlevine: And even then, I'm not quite sure what this is doing really.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -527,13 +527,27 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+            if overscanIsInt:
+                if hasattr(overscanImage, "image"):
+                    imageI = overscanImage.image.convertI()
+                    overscanImageI = afwImage.MaskedImageI(imageI, overscanImage.mask, overscanImage.variance)
+                else:
+                    overscanImageI = overscanImage.convertI()
+            else:
+                overscanImageI = overscanImage
+        else:
+            overscanImageI = overscanImage
```
mfisherlevine: This giant cascade of `elif if if if / else else else` is horrendous! It's not your fault at all, this code was a mess when you arrived, but I think it desperately needs refactoring. Can you file a ticket to refactor this whole function please? (You don't have to do the ticket, of course)

tests/test_isrTask.py
```diff
@@ -340,6 +339,25 @@ def test_overscanCorrection(self):
         statAfter = computeImageMedianAndStd(self.inputExp.image[self.amp.getRawDataBBox()])
         self.assertLess(statAfter[0], statBefore[0])
 
+    def test_overscanCorrectionMedianPerRow(self):
```
mfisherlevine: Like we discussed on Slack, this is a great start. Could you extend it to putting in the sine wave and checked that after doing the correction things are reasonably flat though? I think that would be a great check.

mfisherlevine: (and especially important if this is all going to be refactored soon)

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -546,10 +560,30 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
-            collapsed.data[collapsed.mask] = numpy.mean(biasArray.data[collapsed.mask], axis=1)
+
+        if fitType in ('MEDIAN_PER_ROW',):
+            collapsed = []
+            fitTypeStats = afwMath.stringToStatisticsProperty('MEDIAN')
+            # Loop over rows because there is no afwMath.Median(array, axis=1)
+            for row in biasMaskedArr:
+                if row.mask is False:
+                    row.mask = numpy.repeat(False, len(row))
+                medianRow = afwMath.makeStatistics(row, row.mask, fitTypeStats, statControl).getValue()
```
mfisherlevine: Also, I just realised, doing this here means you're actually doing a single-round-sigma-clipped row-wise median, right? Not saying that's bad _per se_ but trying to understand what's happening here.

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: EDIT: I tried moving it and it was a bad bad time. It should live here, and this code *really* needs a refactor!

python/lsst/ip/isr/isrFunctions.py
plazas: Done: https://jira.lsstcorp.org/browse/DM-23396

tests/test_overscanCorrection.py
```diff
@@ -106,7 +111,6 @@ def checkPolyOverscanCorrectionX(self, **kwargs):
 
         dataBox = lsst.geom.Box2I(lsst.geom.Point2I(0, 0), lsst.geom.Extent2I(10, 10))
         dataImage = afwImage.MaskedImageF(maskedImage, dataBox)
-
```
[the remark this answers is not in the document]
plazas: I don't remember why that's there, probably I was testing some code, or put a "print" statement and then deleted it. I don't care either; why is it important?

python/lsst/ip/isr/isrFunctions.py
plazas: At first, I thought about including the new code where you say, but then I saw that the other part (the part that collapses the overscan for the polynomial corrections) had stuff that included dealing with long vs short axis, clipping, and masking, so I thought we wanted to include that. I had to fiddle with the code to make it pass the tests (and probably made it uglier by adding more control statements; I'm sorry). However---and as you have corroborated independently too---when you move things around then you end up breaking other parts.

python/lsst/ip/isr/isrFunctions.py
plazas: Actually, one of the unit tests found that it could be `False` if this condition is not satisfied: https://github.com/lsst/ip_isr/blob/tickets/DM-23238/python/lsst/ip/isr/isrFunctions.py#L545 (a masked numpy array is not created).

python/lsst/ip/isr/isrFunctions.py
plazas: The second argument of `medianRow = afwMath.makeStatistics(row, row.mask, fitTypeStats, statControl).getValue()`---one line below---is expecting a list. It fails if it just `False`, so I'm creating a list of `False` to pass a valid argument.

python/lsst/ip/isr/isrFunctions.py
plazas: (and yes, I agree that it is not elegant)

tests/test_isrTask.py
plazas: We discussed that we could add a fake sinusoidal signal when we thought that this function was the only test for overscan correction. But then I saw that there are more tests in `test_overscanCorrection.py` that create mock images with mock overscans (10x10 imaging area full of 10's, with a 10x3 overscan region full of 2's), and I thought that creating the sine wave was not necessary anymore (these tests pass).  Would you like to see another test that, instead of the 2's in the overscan, has values that follow a sine wave?

python/lsst/ip/isr/isrFunctions.py
plazas: There are a few lines before the loop that do some clipping, and then, in the loop, the median is taken, as you point out. This section of the code was originally intended for the polynomial cases of the overscan correction, so it might be redundant in the case of the median per row.  But then if I move the loop to the other section, things might break as we pointed out in another comment.

python/lsst/ip/isr/isrFunctions.py
plazas: If you want, I could try to move the loop to that other part and see if I can make things work.

tests/test_overscanCorrection.py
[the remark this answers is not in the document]
mfisherlevine: I think it's just generally untidy and making diffing more confusing. @timj are there reasons to care deeply about such things?

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Yeah, this whole function is incredibly unwieldy and in desperate need of refactoring. Let's leave this here for now (as discussed OOB).

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: It might `== False` but surely it can't _be_ `False`, right?

mfisherlevine: _i.e._ remember that `0 == False` is True, but `0 is False` is not.

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: I don't think you have to pass in a mask at all. Can't you just remove the argument and call with a different signature?

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Great, thanks!

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Also, as discussed OOB, I don't think we _ever_ want to mask a row - it makes sense for things where a spline or poly is being fitted up the column, but for a row-wise median we _have_ to have a number for every row.

tests/test_isrTask.py
mfisherlevine: Yeah, I think that would be good. Just checking the image median and std-dev are lower is a _really_ low bar; doesn't just doing `image = image/2` fulfill that?! (Median and stddev both go down there)

python/lsst/ip/isr/isrFunctions.py
mfisherlevine: Can you check with @RobertLuptonTheGood (or he can confirm here, now that he's tagged) that he's OK with doing a single round of clipping before doing the per-row median please? I don't it's necessarily bad, but maybe there is some statistical subtlety that will cause him to dislike this.

python/lsst/ip/isr/isrFunctions.py
timj: You have to be careful if a mask is in theory a numpy array here. There are many definitions of "truth" available depending on whether there is any numpy mask attached and whether it's filled with zeroes. If mask is meant to be a numpy array I would not expect it to ever be truly the boolean `False`.  Is this really testing that `mask` is `None` rather than `False`?

tests/test_overscanCorrection.py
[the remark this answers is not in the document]
timj: Usually removing or adding lines in parts of the code not being changed is not seen as worthwhile unless it's making the code compliant with some standard. I tend to avoid these sorts of changes because they distract people reading the diff into wondering why the line disappeared. If removing/adding the blank line is wanted it should be done as a distinct commit making clear this is a formatting cleanup.

mfisherlevine:
For any onlookers, @czwa is taking over the review now.

plazas:
I moved the loop that calculates the median per row to the part of the code where it says `elif fitType in ('MEDIAN',): if overscanIsInt`: (before, it was in the `fitType in ('POLY', 'CHEB', 'LEG', 'NATURAL_SPLINE', 'CUBIC_SPLINE', 'AKIMA_SPLINE')` part), as initially suggested by Merlin. This part does not have masking or clipping.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -520,14 +520,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+            offImage = afwImage.ImageF(ampImage.getDimensions())
+            overscanFit = afwImage.ImageF(overscanImageI.getDimensions())
+            if shortInd == 0:
+                offArray = numpy.transpose(offImage.getArray())
+                ovFitArray = numpy.transpose(overscanFit.getArray())
+            else:
+                offArray = offImage.getArray()
+                ovFitArray = overscanFit.getArray()
+            offArray[:, :] = collapsedReshaped
+            ovFitArray[:, :] = numpy.repeat(collapsed, biasArray.shape[1]).reshape(biasArray.shape)
```
czwa: Is it possible to refactor this to use the simpler `offArray[:, :] = vector[:, numpy.newaxis] or vector[numpy.newaxis, :]` as is done for the spline/polynomial fits below (L629 on master)?  I think all these transposes and repeats are correct, but I find the newaxis notation easier to follow.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -520,14 +520,46 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+            del collapsedReshaped, collapsed, biasArray
 
         if overscanIsInt:
             del overscanImageI
-    elif fitType in ('POLY', 'CHEB', 'LEG', 'NATURAL_SPLINE', 'CUBIC_SPLINE', 'AKIMA_SPLINE'):
+    elif fitType in ('POLY', 'CHEB', 'LEG', 'NATURAL_SPLINE', 'CUBIC_SPLINE',
```
czwa: Is this change needed? Same with whitespace below on L561.

python/lsst/ip/isr/isrFunctions.py
plazas: Yes. I followed this suggestion (thanks), and have followed what is done in the spline/polynomial part.

python/lsst/ip/isr/isrFunctions.py
plazas: I fixed these.

PR #126, opened 2020-02-11
DM-23023: Simplify linearity corrections

czwa:
Implement RFC-665 changes:
 * allow linearity to be persisted in a yaml file (giving linearity measurement code a target).
 * allow cameraGeom linearity to be overridden.
 * pull all duplicated code into a single Linearizer() class.
 * demote the current linearize functors to operate on an image view
 * Define LinearityType=Proportional as a no-op, and add generic polynomial correction code.

### DIFF
--- a/python/lsst/ip/isr/linearize.py
+++ b/python/lsst/ip/isr/linearize.py
@@ -26,226 +26,510 @@ import numpy as np
 from lsst.pipe.base import Struct
 from .applyLookupTable import applyLookupTable
 
-__all__ = ["LinearizeBase", "LinearizeLookupTable", "LinearizeSquared"]
+__all__ = ["Linearizer",
+           "LinearizeBase", "LinearizeLookupTable", "LinearizeSquared",
+           "LinearizeProportional", "LinearizePolynomial", "LinearizeNone"]
 
 
-def getLinearityTypeByName(linearityTypeName):
-    """Determine the linearity class to use if the type is known.
+class Linearizer(abc.ABC):
+    """Parameter set for linearization.
+
+    These parameters are included in cameraGeom.Amplifier, but
+    should be accessible externally to allow for testing.
 
     Parameters
     ----------
-    linearityTypeName : str
-        String name of the linearity type that is needed.
-
-    Returns
-    -------
-    linearityType : `~lsst.ip.isr.linearize.LinearizeSquared`
-        The appropriate linearity class to use.  If no matching class
-        is found, `None` is returned.
+    table : `numpy.array`, optional
+        Optional lookup table to use for correction.
+    override : `bool`, optional
+        Override the parameters defined in the detector/amplifier.
+    log : `lsst.log.Log`, optional
+        Logger to handle messages.
     """
-    for t in [LinearizeLookupTable, LinearizeSquared]:
-        if t.LinearityType == linearityTypeName:
-            return t
-    return None
+    def __init__(self, table=None, detector=None, override=False, log=None):
+        self._detectorName = None
+        self._detectorSerial = None
+
+        self.linearityCoeffs = dict()
+        self.linearityType = dict()
+        self.linearityThreshold = dict()
+        self.linearityMaximum = dict()
+        self.linearityUnits = dict()
+        self.linearityBBox = dict()
+
+        self.override = override
+        self.populated = False
+        self.log = log
+
+        self.tableData = None
+        if table is not None:
+            if len(table.shape) != 2:
+                raise RuntimeError("table shape = %s; must have two dimensions" % (table.shape,))
+            if table.shape[1] < table.shape[0]:
+                raise RuntimeError("table shape = %s; indices are switched" % (table.shape,))
+            self.tableData = np.array(table, order="C")
+
+        if detector:
+            self.fromDetector(detector)
+
+    def __call__(self, exposure):
+        """Apply linearity, setting parameters if necessary.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure to correct.
+
+        Returns
+        -------
+        output : `lsst.pipe.base.Struct`
+            Linearization results:
+            ``"numAmps"``
+                Number of amplifiers considered.
+            ``"numLinearized"``
+                Number of amplifiers linearized.
+        """
+
+    def fromDetector(self, detector):
+        """Read linearity parameters from a detector.
+
+        Parameters
+        ----------
+        detector : `lsst.afw.cameraGeom.detector`
+            Input detector with parameters to use.
+        """
+        self._detectorName = detector.getName()
+        self._detectorSerial = detector.getSerial()
+        self.populated = True
+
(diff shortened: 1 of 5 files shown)


python/lsst/ip/isr/linearize.py
```diff
@@ -26,226 +26,510 @@
...
+            before truncation; this supports tables that can handle
+            negative image values; also, if the c1 ends with .5 then
+            the nearest index is used instead of truncating to the
+            next smaller index
 
-    In order to keep related data together, the coefficients are persisted along with the table.
-    """
-    LinearityType = "LookupTable"
+    In order to keep related data together, the coefficients are
+    persisted along with the table.
```
plazas: This docstring seems to come from the old code, correct? Does it still describe properly what your new code does?  I have some questions (some of them actually come from t things I quite did not understand the first time I read the docstring in the old code): 

-  "for each i,j of image"? which image? The amplifier image? (I believe so, but I think it's better to be explicit in the docstring). 
-  "c0, c1 are collimation coefficients": what does "collimation" exactly mean in this context? 
-  are c0 and c1, in general, not necessarily integers?
- "c1" is a columns index or the correction from the LUT?
- "In order to keep related data together, the coefficients are persisted along with the table.": is this still true?

python/lsst/ip/isr/linearize.py
```diff
@@ -26,226 +26,510 @@
...
+    Parameters
+    ----------
+    table : `numpy.array`
+        Lookup table; a 2-dimensional array of floats:
+            - one row for each row index (value of coef[0] in the amplifier)
             - one column for each image value
-            To avoid copying the table the last index should vary fastest (numpy default "C" order)
-        @param[in] detector  detector information (an instance of lsst::afw::cameraGeom::Detector);
-            the name, serial, and amplifier linearization type and coefficients are saved
+        To avoid copying the table the last index should vary fastest
```
plazas: What does this mean?

python/lsst/ip/isr/linearize.py
```diff
@@ -26,226 +26,510 @@
...
 
-    def checkDetector(self, detector):
-        """Check detector name and serial number, ampInfo table length and linearity type
 
-        @param[in] detector  detector info about image (an lsst.afw.cameraGeom.Detector);
+class LinearizePolynomial(LinearizeBase):
+    """Correct non-linearity with a polynomial mode.
 
-        @throw RuntimeError if anything doesn't match
+    corrImage = uncorrImage + sum_i c_i uncorrImage^(2 + i)
```
plazas: Should we reproduce here some of the explanation that out wrote in the RFC, e.g, that is is assumed that the k0 coefficient is zero (bias subtracted) and that c1=1 as it it degenerate with variations in gain?

tests/test_isrMisc.py
```diff
@@ -88,7 +88,7 @@ def test_linearize(self):
         """Assert that the linearize task does not error when a linearity is requested.
         """
         for linearityTypeName in ('LookupTable', 'Squared', 'Unknown'):
```
plazas: Do you need to check for the other types of LinearityTypeName's such as polynomial?

python/lsst/ip/isr/linearize.py
```diff
@@ -26,226 +26,510 @@
...
-            numOutOfRange += applyLookupTable(ampView, tableRow, colIndOffset)
+        rowInd = int(rowInd)
+        if rowInd < 0 or rowInd > numTableRows:
+            raise RuntimeError("LinearizeLookupTable rowInd=%s not in range[0, %s)" %
+                               (rowInd, numTableRows))
+
+        tableRow = table[rowInd, :]
+        numOutOfRange += applyLookupTable(image, tableRow, colIndOffset)
 
         if numOutOfRange > 0 and log is not None:
```
plazas: You are already raising if the requested rowInd is out of range. What is this "numOutOfRange" counter referring to?

python/lsst/ip/isr/linearize.py
```diff
@@ -26,226 +26,510 @@
...
-        if t.LinearityType == linearityTypeName:
-            return t
-    return None
+    def __init__(self, table=None, detector=None, override=False, log=None):
+        self._detectorName = None
+        self._detectorSerial = None
+
+        self.linearityCoeffs = dict()
+        self.linearityType = dict()
+        self.linearityThreshold = dict()
```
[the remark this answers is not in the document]
czwa: The Threshold and Maximum are values that are defined in the amplifiers in afw.cameraGeom.  The original idea was to have the Threshold be a minimum value below which linearity wasn't needed, and the Maximum to be the maximum value above which the correction was unreliable.  However, no Threshold value has ever been set to anything but 0.0, and the SUSPECT flag is now the default way to mark pixels that are high but not saturated.

The BBox is unfortunately needed to avoid passing an amplifier object around.  The box is still loaded generally loaded from the amplifier.

python/lsst/ip/isr/linearize.py
```diff
@@ -26,226 +26,510 @@
...
+            String name of the linearity type that is needed.
+
+        Returns
+        -------
+        linearityType : `~lsst.ip.isr.linearize.LinearizeBase`
+            The appropriate linearity class to use.  If no matching class
+            is found, `None` is returned.
+        """
+        for t in [LinearizeLookupTable,
+                  LinearizeSquared,
```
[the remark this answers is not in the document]
czwa: Probably not, but I suspect it would require a second RFC to remove it.  This ensures that currently working code doesn't break, and no camera definitions need changing either.

python/lsst/ip/isr/linearize.py
czwa: I've updated the docstring, and moved relevant portions to the Linearizer init docstring.  For your questions:
* The LinearizeBase (of which LinearizeLookupTable is a subclass) operate on a complete afw Image.  The Linearizer code handles iteration over amplifier BBoxes, and passes a view onto the amplifiers' bbox to the LinearizeBase math code.  This allows each of these functions to simply do the math, and not need to worry about amplifiers/bboxes/etc.
* "collimation" is used here as "selecting a column from the table."  c0 is used to select a row in the lookup table for this amplifier, and c1 as an offset in the lookup table vector pointed to by c0.
* c0 and c1 are integers, and the image value is also an integer for lookup tables.  c0 gives the row to use, and (imVal + c1) gives the column.  That entry in the table yields the amount to add to the image.
* Line removed as I don't know that it was ever true.

python/lsst/ip/isr/linearize.py
czwa: The implementation works best if each row vector is contiguous in memory.  This is indexed by  (imVal + c1) in the previous comment.

python/lsst/ip/isr/linearize.py
czwa: That raise docstring is now updated.  We did not previously raise if out of range (despite the docstring), and should not now.  `numOutOfRange` denotes the number of pixels that have (imVal + c1) > len(table[c0]) (the required lookup table entry did not exist).

PR #127, opened 2020-02-24
DM-23396: Function "overscanCorrection" in "isrFunctions.py" needs refactoring

czwa:
Split isrFunctions.overscanCorrection to new class.

This split is necessary to reduce the confusion in overscanCorrection,
reduce the duplication, and limit the number of nested if/else
statements used.  The functionality should be identical, just in
better named and more reusable components.

### DIFF
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -28,16 +28,14 @@ import lsst.afw.image as afwImage
 import lsst.afw.detection as afwDetection
 import lsst.afw.math as afwMath
 import lsst.meas.algorithms as measAlg
-import lsst.pex.exceptions as pexExcept
 import lsst.afw.cameraGeom as camGeom
 
 from lsst.afw.geom.wcsUtils import makeDistortedTanWcs
 from lsst.meas.algorithms.detection import SourceDetectionTask
-from lsst.pipe.base import Struct
 
 from contextlib import contextmanager
 
-from .overscan import OverscanCorrector
+from .overscan import OverscanCorrectionTask, OverscanCorrectionTaskConfig
 
 
 def createPsf(fwhm):
@@ -504,55 +502,19 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
     (normalized between +/-1).
     """
     ampImage = ampMaskedImage.getImage()
-    corrector = OverscanCorrector(fitType, statControl=statControl, sigma=collapseRej)
 
-    if fitType in ('MEAN', 'MEANCLIP', 'MEDIAN'):
-        overscanValue, maskArray, isTransposed = corrector.overscanConstant(overscanImage)
-    elif fitType in ('MEDIAN_PER_ROW', 'POLY', 'CHEB', 'LEG',
-                     'NATURAL_SPLINE', 'CUBIC_SPLINE', 'AKIMA_SPLINE'):
-        overscanValue, maskArray, isTransposed = corrector.overscanVector(overscanImage, order)
-    else:
-        raise pexExcept.Exception('%s : %s an invalid overscan type' % ("overscanCorrection", fitType))
+    config = OverscanCorrectionTaskConfig()
+    if fitType:
+        config.fitType = fitType
+    if order:
+        config.order = order
+    if collapseRej:
+        config.numSigmaClip = collapseRej
+    if overscanIsInt:
+        config.overscanIsInt = True
 
-    if isinstance(overscanValue, float):
-        offImage = overscanValue
-        overscanModel = overscanValue
-        maskSuspect = None
-    else:
-        offImage = ampImage.Factory(ampImage.getDimensions())
-        offArray = offImage.getArray()
-
-        overscanModel = afwImage.ImageF(overscanImage.getDimensions())
-        overscanArray = overscanModel.getArray()
-
-        if hasattr(ampImage, 'getMask'):
-            maskSuspect = afwImage.Mask(ampImage.getDimensions())
-        else:
-            maskSuspect = None
-
-        if isTransposed:
-            offArray[:, :] = overscanValue[numpy.newaxis, :]
-            overscanArray[:, :] = overscanValue[numpy.newaxis, :]
-            if maskSuspect:
-                maskSuspect.getArray()[:, maskArray] |= ampImage.getMask().getPlaneBitMask("SUSPECT")
-        else:
-            offArray[:, :] = overscanValue[:, numpy.newaxis]
-            overscanArray[:, :] = overscanValue[:, numpy.newaxis]
-            if maskSuspect:
-                maskSuspect.getArray()[maskArray, :] |= ampImage.getMask().getPlaneBitMask("SUSPECT")
-
-        import lsstDebug
-        if lsstDebug.Info(__name__).display:
-            corrector.debugView(overscanImage, overscanValue)
-
-    ampImage -= offImage
-    if maskSuspect:
-        ampImage.getMask().getArray()[:, :] |= maskSuspect.getArray()[:, :]
-    overscanImage -= overscanModel
-    return Struct(imageFit=offImage,
-                  overscanFit=overscanModel,
-                  overscanImage=overscanImage,
-                  edgeMask=maskSuspect)
+    overscanTask = OverscanCorrectionTask(config=config)
+    return overscanTask.run(ampImage, overscanImage)
 
 
 def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain, gains=None):
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -37,6 +37,8 @@ from lsst.pipe.base import Struct
 
 from contextlib import contextmanager
 
+from .overscan import OverscanCorrector
+
 
 def createPsf(fwhm):
     """Make a double Gaussian PSF.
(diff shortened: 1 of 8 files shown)


python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,404 @@
...
+                'LEG': (poly.legendre.legfit, poly.legendre.legval),
+                'NATURAL_SPLINE': (self.splineFit, self.splineEval),
+                'CUBIC_SPLINE': (self.splineFit, self.splineEval),
+                'AKIMA_SPLINE': (self.splineFit, self.splineEval)
+            }[self.fitType]
+
+            coeffs = fitter(indices, collapsed, order)
+            overscanVector = evaler(indices, coeffs)
+            maskArray = self.extrapolationMask(collapsed)
+        return overscanVector, maskArray, isTransposed
```
PaulPrice: Use a `Struct`.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,404 @@
...
+
+        Parameters
+        ----------
+        image : `lsst.afw.image.Image`
+            Input image the overscan solution was determined from.
+        model : `numpy.ndarray` or `float`
+            Overscan model determined for the image.
+        """
+        import lsstDebug
+        if lsstDebug.Info(__name__).display or lsstDebug.Info("lsst.ip.isr.isrFunctions").display:
```
PaulPrice: Generally I prefer something like:

    if not lsstDebug.Info(__name__).display...:
        return

And then everything can be unindented below.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,404 @@
...
+                'CUBIC_SPLINE': (self.splineFit, self.splineEval),
+                'AKIMA_SPLINE': (self.splineFit, self.splineEval)
+            }[self.fitType]
+
+            coeffs = fitter(indices, collapsed, order)
+            overscanVector = evaler(indices, coeffs)
+            maskArray = self.extrapolationMask(collapsed)
+        return overscanVector, maskArray, isTransposed
+
+    def debugView(self, image, model):
```
PaulPrice: I think the plotting should be called from `OverscanCorrectionTask.run`, rather than as an optional extra method.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,404 @@
...
+            self.statControl = afwMath.StatisticsControl()
+        if sigma:
+            self.statControl.setNumSigmaClip(sigma)
+
+    def integerConvert(self, image):
+        """Return an integer version of the input image.
+
+        Parameters
+        ----------
+        image : `numpy.ndarray`, `lsst.afw.image.Image` or `lsst.afw.image.MaskedImage`
```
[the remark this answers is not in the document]
czwa: I could not find any details, and have abbreviated types to fit.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,404 @@
...
+        """Create mask if edges are extrapolated.
+
+        Parameters
+        ----------
+        collapsed : `numpy.ma.masked_array`
+            Masked array to check the edges of.
+
+        Returns
+        -------
+        maskArray : `list` [`bool`]
```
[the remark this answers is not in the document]
czwa: It simply needed to be something that would work.  Promoted to numpy array now.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -502,197 +502,19 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
...
+        config.fitType = fitType
+    if order:
+        config.order = order
+    if collapseRej:
+        config.numSigmaClip = collapseRej
+    if overscanIsInt:
+        config.overscanIsInt = True
+
+    overscanTask = OverscanCorrectionTask(config=config)
+    return overscanTask.run(ampImage, overscanImage)
```
PaulPrice: You should deprecate this function, and modify `IsrTask` to call the `OverscanCorrectionTask` directly. Otherwise, there's not really any advantage to having the overscan correction in a `Task`.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+            "POLY": "Fit ordinary polynomial to the longest axis of the overscan region",
+            "CHEB": "Fit Chebyshev polynomial to the longest axis of the overscan region",
+            "LEG": "Fit Legendre polynomial to the longest axis of the overscan region",
+            "NATURAL_SPLINE": "Fit natural spline to the longest axis of the overscan region",
+            "CUBIC_SPLINE": "Fit cubic spline to the longest axis of the overscan region",
+            "AKIMA_SPLINE": "Fit Akima spline to the longest axis of the overscan region",
+            "MEAN": "Correct using the mean of the overscan region",
+            "MEANCLIP": "Correct using a clipped mean of the overscan region",
+            "MEDIAN": "Correct using the median of the overscan region",
+            "MEDIAN_PER_ROW": "Correct using the median per row of the overscan region",
```
PaulPrice: Add `MEANCLIP_PER_ROW`?

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+        },
+    )
+    order = pexConfig.Field(
+        dtype=int,
+        doc=("Order of polynomial or to fit if overscan fit type is a polynomial, " +
```
PaulPrice: I think the `or` is excess.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+        default='MEDIAN',
+        allowed={
+            "POLY": "Fit ordinary polynomial to the longest axis of the overscan region",
+            "CHEB": "Fit Chebyshev polynomial to the longest axis of the overscan region",
+            "LEG": "Fit Legendre polynomial to the longest axis of the overscan region",
+            "NATURAL_SPLINE": "Fit natural spline to the longest axis of the overscan region",
+            "CUBIC_SPLINE": "Fit cubic spline to the longest axis of the overscan region",
+            "AKIMA_SPLINE": "Fit Akima spline to the longest axis of the overscan region",
+            "MEAN": "Correct using the mean of the overscan region",
+            "MEANCLIP": "Correct using a clipped mean of the overscan region",
```
PaulPrice: Where are the clipping parameters set?

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+    )
+    numSigmaClip = pexConfig.Field(
+        dtype=float,
+        doc="Rejection threshold (sigma) for collapsing overscan before fit",
+        default=3.0,
+    )
+    overscanIsInt = pexConfig.Field(
+        dtype=bool,
+        doc="Treat overscan as an integer image for purposes of overscan.FitType=MEDIAN" +
+            " and overscan.FitType=MEDIAN_PER_ROW.",
```
PaulPrice: What is the leading `overscan.` for?

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+    _DefaultName = "overscan"
+
+    def __init__(self, statControl=None, **kwargs):
+        super().__init__(**kwargs)
+        if statControl:
+            self.statControl = statControl
+        else:
+            self.statControl = afwMath.StatisticsControl()
+
+    def run(self, ampImage, overscanImage):
```
PaulPrice: Needs a docstring.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+        self.statControl.setNumSigmaClip(self.config.numSigmaClip)
```
PaulPrice: I don't think you should be modifying something that's been given to you. Your API seems to want things both ways, but you should choose one: use a `statControl` that you're given, or configure a new `statControl` from the configuration.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+                                     'NATURAL_SPLINE', 'CUBIC_SPLINE', 'AKIMA_SPLINE'):
+            overscanResult = self.measureVectorOverscan(overscanImage, self.config.order)
+        else:
+            raise pexExcept.Exception('%s : %s an invalid overscan type' %
+                                      ("overscanCorrection", self.config.fitType))
+        overscanValue = overscanResult.overscanValue
+        maskArray = overscanResult.maskArray
+        isTransposed = overscanResult.isTransposed
+
+        if isinstance(overscanValue, float):
```
PaulPrice: This belongs in the `measureConstantOverscan` method, or in the above block that calls it. Similarly with the `else` block below.

Try to keep the `run` method as simple as possible.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+        """Return an integer version of the input image.
+
+        Parameters
+        ----------
+        image : `numpy.ndarray`, `lsst.afw.image.Image` or `MaskedImage`
+            Image to convert to integers.
+
+        Returns
+        -------
+        outI : `numpy.ndarray`, `lsst.afw.image.Image` or `MaskedImage`
```
PaulPrice: `outI` feels Hungarian...

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+            imageI = image.image.convertI()
+            outI = afwImage.MaskedImageI(imageI, image.mask, image.variance)
+        elif hasattr(image, "convertI"):
+            # Is an Image:
+            outI = image.convertI()
+        elif hasattr(image, "astype"):
+            # Is a numpy array:
+            outI = image.astype(int)
+        else:
+            raise RuntimeError("Could not convert this to integers: %s %s %s\n",
```
PaulPrice: No newline on the error string.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+            calcImage = image.getArray()
+        return calcImage
+
+    @staticmethod
+    def transpose(image):
+        """Transpose input numpy array if necessary.
+
+        Parameters
+        ----------
+        image : `numpy.ndarray`
```
PaulPrice: I'm not sure it's a good idea to call a 2D numpy array an `image`, for fear of confusing the types. I usually use `array`.

python/lsst/ip/isr/overscan.py
```diff
@@ -0,0 +1,504 @@
...
+
+        return interp.interpolate(indices.astype(float))
+
+    @staticmethod
+    def maskExtrapolated(collapsed):
+        """Create mask if edges are extrapolated.
+
+        Parameters
+        ----------
+        collapsed : `numpy.ma.masked_array`
```
PaulPrice: Why the name `collapsed`? Isn't this just a 1D array?

python/lsst/ip/isr/isrFunctions.py
czwa: The newest commit does this, although it still calls it from an IsrTask.overscanCorrection call.  The overscan chopping is still in IsrTask, allowing OverscanCorrectionTask to concentrate on the image-level math.

This has necessitated config changes on obs_ packages to point to the updated overscan config item.

python/lsst/ip/isr/overscan.py
czwa: OverscanCorrectionTaskConfig.numSigmaClip controls the clipping.

python/lsst/ip/isr/overscan.py
czwa: Not on this ticket.  I think after the next latiss run there will be more work on sorting out the peculiar overscan structure that prompted the MEDIAN_PER_ROW (which in turn prompted this rewrite so new methods can be added easier in the future).

python/lsst/ip/isr/overscan.py
czwa: Yes, but it's the 1d array after the second dimension has been removed.

python/lsst/ip/isr/overscan.py
```diff
@@ -131,6 +152,9 @@ def run(self, ampImage, overscanImage):
                 overscanArray[:, :] = overscanValue[:, np.newaxis]
                 if maskSuspect:
                     maskSuspect.getArray()[maskArray, :] |= ampImage.getMask().getPlaneBitMask("SUSPECT")
+        else:
+            raise pexExcept.Exception('%s : %s an invalid overscan type' %
```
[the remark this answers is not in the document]
timj: Also in python we really want to raise python exceptions not pex_exception exceptions. Those exist so that C++ code can raise python-style exceptions.

python/lsst/ip/isr/overscan.py
```diff
@@ -89,30 +89,51 @@ def __init__(self, statControl=None, **kwargs):
...
+
+        Returns
+        -------
+        overscanResults : `lsst.pipe.base.Struct`
+            Result struct with components:
+            - ``imageFit`` : scalar or `lsst.afw.image.Image`
+                Value or fit subtracted from the amplifier image data.
+            - ``overscanFit`` : scalar or `lsst.afw.image.Image`
+                Value or fit subtracted from the overscan image data.
+            - ``overscanImage`` : `lsst.afw.image.Image`
```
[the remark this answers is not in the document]
czwa: This is used (optionally) in the empirical variance calculation.  That happens after the trim and assembly, and I suspect saving that image segment as part of the overscan results struct was chosen instead of deciding whether to save it during the trim step.

python/lsst/ip/isr/overscan.py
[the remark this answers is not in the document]
czwa: Changed to RuntimeError.

PR #128, opened 2020-03-02
DM-22655: Genericize gen2to3.py to be useable with any gen2 repo

parejkoj:


python/lsst/ip/isr/isrTask.py
```diff
@@ -60,7 +60,7 @@
 class IsrTaskConnections(pipeBase.PipelineTaskConnections,
                          dimensions={"instrument", "visit", "detector"},
                          defaultTemplates={}):
-    ccdExposure = cT.PrerequisiteInput(
+    ccdExposure = cT.Input(
```
timj: Is this change related to DM-23702? Should we be making that change here or deferring to that ticket?

czwa: I am currently implementing this change on DM-23702.

parejkoj: Yes, I put this in here to help with the pipeline testing in DM-2361. I'll take it out.

parejkoj:
@timj : thoughts on my last two commits on this ticket? (I'll squash them together before merging.) It gets me a clean Jenkins, and I think it may in fact be the correct behavior, given how we currently mismanage master calibrations.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1981,16 +1981,15 @@ def darkCorrection(self, exposure, darkExposure, invert=False):
         expScale = exposure.getInfo().getVisitInfo().getDarkTime()
         if math.isnan(expScale):
             raise RuntimeError("Exposure darktime is NAN.")
-        if darkExposure.getInfo().getVisitInfo() is not None:
+        if darkExposure.getInfo().getVisitInfo() is not None \
+                and not math.isnan(darkExposure.getInfo().getVisitInfo().getDarkTime()):
```
timj: Do we get NaN because of a bug? Or because pipe_drivers/cp_pipe isn't putting that information anywhere?

parejkoj: I'm pretty sure it's the latter: we've been treating calibrations as `Image` and/or `MaskedImage` mostly, so their VisitInfo was always null. Now we're reading them as Exposure, so they get a default-constructed VisitInfo.

PR #132, opened 2020-03-28
DM-23044: PTC task should persist usable linearity models 

plazas:


python/lsst/ip/isr/linearize.py
```diff
@@ -56,9 +62,14 @@ class Linearizer(abc.ABC):
         Raised if the supplied table is not 2D, or if the table has fewer
         columns than rows (indicating that the indices are swapped).
     """
+
+    _OBSTYPE = "linearizer"
+    """The calibration type used for ingest."""
```
timj: Might be more accurate to call this the "dataset type name used for this class".

python/lsst/ip/isr/linearize.py
```diff
@@ -56,9 +62,14 @@ class Linearizer(abc.ABC):
...
+
     def __init__(self, table=None, detector=None, override=False, log=None):
```
timj: The docstring has not been updated to include the detector argument.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+            Name of the file containing the linearity definition.
+        Returns
+        -------
+        linearity : `~lsst.ip.isr.linearize.Linearizer``
+            Linearity parameters.
+        """
+        data = ''
+        try:
+            with open(filename, 'r') as f:
+                data = yaml.load(f, Loader=yaml.FullLoader)
```
timj: Is the FullLoader needed because you are writing out numpy arrays natively into the YAML file (and I see PropertyList is not being turned into a dict first)? It still might be worth trying yaml.CFullLoader instead. It always gets a bit tricky to write special python classes to yaml.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+        except Exception:
```
timj: What's going on here? What are the scenarios where the exception will be raised? If you are always going to this exception handler why is the first attempt happening?

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+            def yamlLoad(x):
+                return yaml.load(x, Loader=yaml.Loader)
+            with open(filename, 'r') as f:
+                data = yamlLoad(f)
```
timj: Why is this calling `yamlLoad` and not `yaml.load`? The `yamlLoad` only has one line in it anyhow.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+            The name of the file used to write the data.
+
+        Notes
+        -----
+        The file is written to YAML format and will include any metadata
+        associated with the `Linearity`.
+        """
+        outDict = self.toDict()
+
+        with open(filename, 'w') as f:
```
timj: You need to ensure that the file extension is `.yaml` since you can't rely on the caller of `writeText` knowing what the text format will be.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+        linDict['detectorId'] = metadata['DETECTOR_ID']
+        linDict['detectorName'] = metadata['DETECTOR_NAME']
+        try:
+            linDict['detectorSerial'] = metadata['DETECTOR_SERIAL']
+        except Exception:
+            linDict['detectorSerial'] = 'NOT SET'
+        linDict['amplifiers'] = dict()
+
+        for row in table:
+            record = row.extract("*")
```
timj: See DM-24055 to see why this is going to be slow. My fault for getting it wrong in `Defects`.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+        Parameters
+        ----------
+        filename : `str`
+            Name of the file containing the linearity definition.
+        Returns
+        -------
+        linearity : `~lsst.ip.isr.linearize.Linearizer``
+            Linearity parameters.
+        """
+        table = afwTable.BaseCatalog.readFits(filename)
```
timj: Please refactor this routine such that the code that understands the afw table is a distinct method. I called it `fromTable` in Defects. `readFits` should then be "read the first table from FITS file and then call fromAmpTable, read the second table from FITS and call fromLookupTable". That way other code can reuse it. It's a trivial reorganization.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+            ampDict['linearityBBox'] = Box2I(Point2I(record['BBOX_X0'], record['BBOX_Y0']),
+                                             Extent2I(record['BBOX_DX'], record['BBOX_DY']))
+
+            linDict['amplifiers'][ampName] = ampDict
+
+        try:
+            table = afwTable.BaseCatalog.readFits(filename, 2)
+            tableData = []
+            for row in table:
+                record = row.extract('*')
```
timj: Again. My fault but this has to be changed for performance. It could now probably look like:
```python
linDict["tableData"] = [record[lookupValuesKey] for record in table]
```

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+
+        if self.tableData is not None:
+            schema = afwTable.Schema()
+            dimensions = self.tableData.shape
+            lut = schema.addField("LOOKUP_VALUES", type='ArrayI', size=dimensions[1],
+                                  doc="linearity lookup data")
+            catalog = afwTable.BaseCatalog(schema)
+            catalog.resize(dimensions[0])
+
+            for ii in range(dimensions[0]):
```
timj: afw allows you to assign numpy arrays directly to a column so given you have one column you might be able to assign the entire tableData array to the afwTable in one assignment. @TallJimbo will this work?

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+            for row in table:
+                record = row.extract('*')
+                tableData.append(record['LOOKUP_VALUES'])
+            linDict['tableData'] = tableData
+        except Exception:
+            pass
+
+        return cls().fromYaml(linDict)
+
+    def writeFits(self, filename):
```
timj: Please have writeFits call `toAmpTable` and `toTableDataTable` so that we can call to and from table methods independently of writing a FITS file.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+        """
+        return self._metadata
+
+    def setMetadata(self, metadata=None):
+        """Store a copy of the supplied metadata with the `Linearizer`.
+
+        Parameters
+        ----------
+        metadata : `lsst.daf.base.PropertyList`, optional
+            Metadata to associate with the defects.  Will be copied and
```
timj: Not defects.

python/lsst/ip/isr/linearize.py
```diff
@@ -127,32 +139,280 @@ def fromYaml(self, yamlObject):
...
+        """
+        mdOriginal = self.getMetadata()
+        mdSupplemental = dict()
+
+        if date:
+            mdSupplemental['CALIBDATE'] = date.isoformat()
+            mdSupplemental['CALIB_CREATION_DATE'] = date.date().isoformat(),
+            mdSupplemental['CALIB_CREATION_TIME'] = date.time().isoformat(),
+        if detectorId:
+            mdSupplemental['DETECTOR_ID'] = f"{detectorId}"
```
timj: DETECTOR I think for compatibility.

python/lsst/ip/isr/linearize.py
TallJimbo: Can't assign multiple columns at once, I'm afraid.

But to back up a bit - why are we using `afw.table` at all here (or in defects, for that matter)?  Do we use this data structure in C++ at some point?

timj: No. We used it to so that we can call Catalog.writeFits with a PropertyList and afwTable. If you want us to only use astropy.table to write and read fits tables we can switch but that won't be immediate.

Note that we are assigning a single column here but it's a 1D cell.

TallJimbo: I think I would recommend astropy in the longer term; I'm trying to reduce our usage of afw.table when there's something third-party that does the job, and while that has focused on Pandas/Parquet for catalogs, it probably should apply to astropy for general FITS I/O as well.

(Or something else - I know several DM members who prefer Erin Sheldon's fitsio package, as it's a lot faster, but for a lot of stuff even astropy is faster than afw.table.)

python/lsst/ip/isr/linearize.py
plazas: For now, I'm changing this to [[yaml.CLoader]], but there is an ongoing discussion in https://jira.lsstcorp.org/browse/DM-24225

timj: Good. Yes. That will help a lot and yes DM-24225 is where we are discussing the broader issue.

python/lsst/ip/isr/linearize.py
plazas: I was having some loading error when I was developing this part of the code by using `yaml.load`, and I found out this workaround to move forward. I'm re-running the code, and things seem to be fine now, so I will delete the extra function.

python/lsst/ip/isr/isrTask.py
```diff
@@ -959,6 +960,8 @@ def readIsrData(self, dataRef, rawExposure):
         # immediate=True required for functors and linearizers are functors; see ticket DM-6515
         linearizer = (dataRef.get("linearizer", immediate=True)
                       if self.doLinearize(ccd) else None)
+        if linearizer is not None and not isinstance(linearizer, numpy.ndarray):
```
timj, line 963: Is this check for ndarray a backwards compatibility thing?

czwa: Yes.  Until we move everything to consistent calibration types, we don't know what we got from the butler.

PR #135, opened 2020-04-23
DM-24537: Add Isr calibration base class

czwa:
Add IsrCalib and IsrProvenance classes.
IsrCalib should be the base class used by calibration data.
IsrProvenance supplies a concrete example subclass that will allow
calibrations to store their inputs in a way that can be used to
reconstruct the calibration.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -1,9 +1,10 @@
+# This file is part of ip_isr.
 #
-# LSST Data Management System
-# Copyright 2008-2017 AURA/LSST.
-#
-# This product includes software developed by the
-# LSST Project (http://www.lsst.org/).
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -15,19 +16,19 @@
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 #
-# You should have received a copy of the LSST License Statement and
-# the GNU General Public License along with this program.  If not,
-# see <https://www.lsstcorp.org/LegalNotices/>.
-#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
 import abc
 import copy
 import datetime
+import os.path
+import warnings
 import yaml
+from astropy.table import Table
 
-import lsst.afw.table as afwTable
 from lsst.log import Log
 from lsst.daf.base import PropertyList
-from astropy.table import Table
+
 
 __all__ = ["IsrCalib", "IsrProvenance"]
 
@@ -65,14 +66,13 @@ class IsrCalib(abc.ABC):
     def __init__(self, detectorName=None, detectorSerial=None, detector=None, log=None, **kwargs):
         self._detectorName = detectorName
         self._detectorSerial = detectorSerial
-        self._metadata = PropertyList()
+        self.setMetadata(PropertyList())
 
-        # Ensure that we have the obs type required by calibration ingest
-        self._metadata["OBSTYPE"] = self._OBSTYPE
-        self._metadata[self._OBSTYPE + "_SCHEMA"] = self._SCHEMA
-        self._metadata[self._OBSTYPE + "_VERSION"] = self._VERSION
+        # Define the required attributes for this calibration.
+        self.requiredAttributes = set(['_OBSTYPE', '_SCHEMA', '_VERSION'])
+        self.requiredAttributes.update(['_detectorName', '_detectorSerial', '_metadata'])
 
-        self.log = log if log else Log.getLogger("lsst.ip.isr.IsrCalib")
+        self.log = log if log else Log.getLogger(__name__.partition(".")[2])
 
         if detector:
             self.fromDetector(detector)
@@ -90,12 +90,19 @@ class IsrCalib(abc.ABC):
         if not isinstance(other, self.__class__):
             return False
 
-        for attr in ('_OBSTYPE', '_SCHEMA', '_VERSION',
-                     '_detectorName', '_detectorSerial', '_metadata'):
+        for attr in self._requiredAttributes:
             if getattr(self, attr) != getattr(other, attr):
                 return False
 
         return True
+
+    @property
+    def requiredAttributes(self):
+        return self._requiredAttributes
+
+    @requiredAttributes.setter
+    def requiredAttributes(self, value):
+        self._requiredAttributes = value
 
     def getMetadata(self):
 
@@ -119,7 +126,7 @@ class IsrCalib(abc.ABC):
             Metadata to associate with the calibration.  Will be copied and
             overwrite existing metadata.
         """
-        if metadata:
+        if metadata is not None:
             self._metadata = copy.copy(metadata)
 
         # Ensure that we have the obs type required by calibration ingest
@@ -127,7 +134,7 @@ class IsrCalib(abc.ABC):
         self._metadata[self._OBSTYPE + "_SCHEMA"] = self._SCHEMA
         self._metadata[self._OBSTYPE + "_VERSION"] = self._VERSION
 
-    def updateMetadata(self, *args, setDate=False, **kwargs):
+    def updateMetadata(self, setDate=False, **kwargs):
         """Update metadata keywords with new values.
 
         Parameters
@@ -149,9 +156,7 @@ class IsrCalib(abc.ABC):
             mdSupplemental['CALIB_CREATION_DATE'] = date.date().isoformat()
             mdSupplemental['CALIB_CREATION_TIME'] = date.time().isoformat()
 
-        for key, value in kwargs.items():
-            mdSupplemental[key] = value
-
+        mdSupplemental.update(kwargs)
         mdOriginal.update(mdSupplemental)
 
     @classmethod
@@ -167,83 +172,69 @@ class IsrCalib(abc.ABC):
         -------
         calib : `~lsst.ip.isr.IsrCalibType`
             Calibration class.
+
+        Raises
+        ------
+        RuntimeError :
+            Raised if the filename does not end in ".ecsv" or ".yaml".
         """
-        data = ''
-        try:
+        if filename.endswith((".ecsv", ".ECSV")):
             data = Table.read(filename, format='ascii.ecsv')
-        except Exception as e1:
-            try:
-                with open(filename, 'r') as f:
-                    data = yaml.load(f, Loader=yaml.CLoader)
-            except Exception as e2:
-                raise e2(f"Failed to read as ecsv as well: {e1}")
+            return cls.fromTable([data])
 
-        if isinstance(data, dict):
-            return cls().fromDict(data)
+        elif filename.endswith((".yaml", ".YAML")):
+            with open(filename, 'r') as f:
+                data = yaml.load(f, Loader=yaml.CLoader)
+            return cls.fromDict(data)
         else:
-            schema = afwTable.Schema()
-            keyDict = dict()
-            for colName in data.columns:
-                try:
-                    keyDict[colName] = schema.addField(colName,
-                                                       type=data[colName].dtype.type)
-                except KeyError:
-                    keyDict[colName] = schema.addField(colName, type="String", size=32)
+            raise RuntimeError(f"Unknown filename extension: {filename}")
 
-            table = afwTable.BaseCatalog(schema)
-            table.resize(len(data))
-
-            for colName in data.columns:
-                key = keyDict[colName]
-                if key.getTypeString() == 'String':
-                    for rowNum, row in enumerate(data[colName]):
-                        table[rowNum][key] = row
-                else:
-                    table[key] = data[colName]
-
-            metadata = PropertyList()
-            for k, v in data.meta.items():
-                metadata[k] = v
-            table.setMetadata(metadata)
-
-            return cls().fromTable([table])
-
-    def writeText(self, filename):
+    def writeText(self, filename, format='auto'):
         """Write the calibration data to a text file.
 
         Parameters
         ----------
         filename : `str`
             Name of the file to write.
-
+        format : `str`
+            Format to write the file as.  Supported values are:
+                ``"auto"`` : Determine filetype from filename.
+                ``"yaml"`` : Write as yaml.
+                ``"ecsv"`` : Write as ecsv.
         Returns
         -------
         used : `str`
-            The name of the file used to write the data.
+            The name of the file used to write the data.  This may
+            differ from the input if the format is explicitly chosen.
 
         Raises
         ------
         RuntimeError :
-            Raised if filename does not end in a known extension
+            Raised if filename does not end in a known extension, or
+            if all information cannot be written.
 
         Notes
         -----
         The file is written to YAML/ECSV format and will include any
         associated metadata.
+
         """
-        outDict = self.toDict()
-        if filename.lower().endswith((".yaml")):
+        if format == 'yaml' or (format == 'auto' and filename.lower().endswith((".yaml", ".YAML"))):
+            outDict = self.toDict()
+            path, ext = os.path.splitext(filename)
+            filename = path + ".yaml"
             with open(filename, 'w') as f:
                 yaml.dump(outDict, f)
-        elif filename.lower().endswith((".ecsv")):
+        elif format == 'ecsv' or (format == 'auto' and filename.lower().endswith((".ecsv", ".ECSV"))):
             tableList = self.toTable()
-            tableX = tableList[0]
+            if len(tableList) > 1:
+                # ECSV doesn't support multiple tables per file, so we
+                # can only write the first table.
+                raise RuntimeError(f"Unable to persist {len(tableList)}tables in ECSV format.")
 
-            table = tableX.asAstropy()
-            table.meta = tableX.getMetadata().toDict()
-
-            # ECSV doesn't support multiple tables per file, so we can only
-            # write the first table.
+            table = tableList[0]
+            path, ext = os.path.splitext(filename)
+            filename = path + ".ecsv"
             table.write(filename, format="ascii.ecsv")
         else:
             raise RuntimeError(f"Attempt to write to a file {filename} "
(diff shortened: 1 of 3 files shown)


python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        self._detectorName = detectorName
+        self._detectorSerial = detectorSerial
+        self._metadata = PropertyList()
+
+        # Ensure that we have the obs type required by calibration ingest
+        self._metadata["OBSTYPE"] = self._OBSTYPE
+        self._metadata[self._OBSTYPE + "_SCHEMA"] = self._SCHEMA
+        self._metadata[self._OBSTYPE + "_VERSION"] = self._VERSION
+
+        self.log = log if log else Log.getLogger("lsst.ip.isr.IsrCalib")
```
timj, line 75: Use `__name__` here. The convention for lsst.log loggers (as opposed to python `logging`) is I think to drop the leading `lsst.`. Such as:
```python
Log.getLogger(__name__.partition(".")[2])
```

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+    _OBSTYPE = 'generic'
+    _SCHEMA = 'NO SCHEMA'
+    _VERSION = 0
+
+    def __init__(self, detectorName=None, detectorSerial=None, detector=None, log=None, **kwargs):
+        self._detectorName = detectorName
+        self._detectorSerial = detectorSerial
+        self._metadata = PropertyList()
+
+        # Ensure that we have the obs type required by calibration ingest
```
timj, line 70: This duplicates the code in setMetadata below so I think it might be a good idea to unify this.
For example calling `self.setMetadata(PropertyList())` here.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        """Calibration equivalence.
+
+        Subclasses will need to check specific sub-properties.  The
+        default is only to check common entries.
+        """
+        if not isinstance(other, self.__class__):
+            return False
+
+        for attr in ('_OBSTYPE', '_SCHEMA', '_VERSION',
+                     '_detectorName', '_detectorSerial', '_metadata'):
```
timj, line 94: It might help to make this second line of attributes be a class property so that subclasses can set the one list and then reuse this `__eq__` implementation.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        """
+        if metadata:
+            self._metadata = copy.copy(metadata)
+
+        # Ensure that we have the obs type required by calibration ingest
+        self._metadata["OBSTYPE"] = self._OBSTYPE
+        self._metadata[self._OBSTYPE + "_SCHEMA"] = self._SCHEMA
+        self._metadata[self._OBSTYPE + "_VERSION"] = self._VERSION
+
+    def updateMetadata(self, *args, setDate=False, **kwargs):
```
timj, line 130: what are the `*args` parameters here?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        """
+        mdOriginal = self.getMetadata()
+        mdSupplemental = dict()
+
+        self._metadata["DETECTOR"] = self._detectorName
+        self._metadata["DETECTOR_SERIAL"] = self._detectorSerial
+
+        if setDate:
+            date = datetime.datetime.now()
+            mdSupplemental['CALIBDATE'] = date.isoformat()
```
timj, line 148: I'm sure at some point we will work out precisely what CALIBDATE means...

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            mdSupplemental['CALIB_CREATION_DATE'] = date.date().isoformat()
+            mdSupplemental['CALIB_CREATION_TIME'] = date.time().isoformat()
+
+        for key, value in kwargs.items():
+            mdSupplemental[key] = value
```
timj, line 153: doesn't `mdSupplemental.update(kwargs)` work here?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            Name of the file containing the calibration definition.
+
+        Returns
+        -------
+        calib : `~lsst.ip.isr.IsrCalibType`
+            Calibration class.
+        """
+        data = ''
+        try:
+            data = Table.read(filename, format='ascii.ecsv')
```
timj, line 173: Can't we trust the filename extension?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            for colName in data.columns:
+                key = keyDict[colName]
+                if key.getTypeString() == 'String':
+                    for rowNum, row in enumerate(data[colName]):
+                        table[rowNum][key] = row
+                else:
+                    table[key] = data[colName]
+
+            metadata = PropertyList()
+            for k, v in data.meta.items():
```
timj, line 205: `metadata.update(data.meta)`?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            data = Table.read(filename, format='ascii.ecsv')
+        except Exception as e1:
+            try:
+                with open(filename, 'r') as f:
+                    data = yaml.load(f, Loader=yaml.CLoader)
+            except Exception as e2:
+                raise e2(f"Failed to read as ecsv as well: {e1}")
+
+        if isinstance(data, dict):
+            return cls().fromDict(data)
```
timj, line 182: I don't think this needs to be `cls().` does it? Isn't it a class method?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        try:
+            data = Table.read(filename, format='ascii.ecsv')
+        except Exception as e1:
+            try:
+                with open(filename, 'r') as f:
+                    data = yaml.load(f, Loader=yaml.CLoader)
+            except Exception as e2:
+                raise e2(f"Failed to read as ecsv as well: {e1}")
+
+        if isinstance(data, dict):
```
timj, line 181: I don't think we need to be doing instance checks here.  I think that we can safely say:

```
if endswith yaml:
    read yaml
    pass yaml to cls.fromDict 
elif enswith ecsv:
    read table
    process table
else:
    get upset
```

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            return cls().fromDict(data)
+        else:
+            schema = afwTable.Schema()
```
timj, line 184: @TallJimbo has a strong preference for us to use astropy.table in all these curated calibrations and not use afw table at all. The reasoning is that we are never passing these tables to C++ and astropy is probably faster.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+                        table[rowNum][key] = row
+                else:
+                    table[key] = data[colName]
+
+            metadata = PropertyList()
+            for k, v in data.meta.items():
+                metadata[k] = v
+            table.setMetadata(metadata)
+
+            return cls().fromTable([table])
```
timj, line 209: Again don't need to create an instance of `cls` here.

tests/test_calibType.py
```diff
@@ -0,0 +1,74 @@
...
+                                   calibType="Test Calib")
+        self.calib.updateMetadata()
+        self.calib.fromDataIds([{'exposure': 1234, 'detector': 0, 'filter': 'G'},
+                                {'exposure': 1235, 'detector': 0, 'filter': 'G'},
+                                {'exposure': 1234, 'detector': 1, 'filter': 'G'},
+                                {'exposure': 1235, 'detector': 1, 'filter': 'G'}])
+
+    def runText(self, textType):
+        filename = tempfile.mktemp()
+        self.calib.writeText(filename + textType)
```
timj, line 45: filename is returned so you can take this file name and use it directly in the next line.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+
+        Parameters
+        ----------
+        filename : `str`
+            Name of the file to write.
+
+        Returns
+        -------
+        used : `str`
+            The name of the file used to write the data.
```
timj, line 222: I did this in Defects because Defects knew that only ecsv format was going to work. That meant I forced the suffix to be ecsv even if people tried to tell me they were writing `.txt` files.  It might be helpful for subclasses to be able to declare whether YAML or ECSV are supported and maybe have an optional param here for `format=auto|yaml|ecsv` -- forcing the suffix in yaml|ecsv mode (complaining if yaml is chosen but is not supported), allowing auto to use the supplied  suffix, and auto could even force ecsv if yaml is not supported.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        elif filename.lower().endswith((".ecsv")):
+            tableList = self.toTable()
+            tableX = tableList[0]
+
+            table = tableX.asAstropy()
+            table.meta = tableX.getMetadata().toDict()
+
+            # ECSV doesn't support multiple tables per file, so we can only
+            # write the first table.
+            table.write(filename, format="ascii.ecsv")
```
timj, line 247: I was thinking about Linearizer. It might be that we could sneak the look up table into the yaml metadata and use the table for the main component used by all.

Shouldn't we warn if we have multiple tables and can't write one out?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        ------
+        RuntimeError :
+            Raised if filename does not end in a known extension
+
+        Notes
+        -----
+        The file is written to YAML/ECSV format and will include any
+        associated metadata.
+        """
+        outDict = self.toDict()
```
timj, line 234: I think this is only used in the yaml branch. It's also not implemented by default so will raise an exception even if ecsv is requested.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            Filename to write data to.
+        """
+        tableList = self.toTable()
+
+        tableList[0].writeFits(filename)
+        for table in tableList[1:]:
+            table.writeFits(filename, "a")
+
+    def fromDetector(self, detector):
+        """Set calibration parameters from the detector.
```
timj, line 296: I find this phrasing confusing.  Maybe "Modify the calibration parameters to match the supplied detector"?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+
+        extNum = 2  # Fits indices start at 1, we've read one already.
+        try:
+            newTable = afwTable.BaseCatalog.readFits(filename, extNum)
+            tableList.append(newTable)
+            extNum += 1
+        except Exception:
+            pass
+
+        return cls().fromTable(tableList)
```
timj, line 279: Do not instantiate cls.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            Returns true if the calibration is valid and appropriate.
+        """
+        return False
+
+    def apply(self, target):
+        """Method to apply the calibration to the target object.
+
+        Parameters
+        ----------
+        target : object
```
timj, line 413: backticks around object.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+
+        Returns
+        -------
+        valid : `bool`
+            Returns true if the calibration is valid and appropriate.
+        """
+        return False
+
+    def apply(self, target):
+        """Method to apply the calibration to the target object.
```
timj, line 409: So for Defects this would take an afwExposure and update the mask?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+        NotImplementedError :
+            Raised if not implemented.
+        """
+        raise NotImplementedError("Must be implemented by subclass.")
+
+
+class IsrProvenance(IsrCalib):
+    """Class for the provenance of data used to construct calibration.
+
+    Provenance is not really a calibration, but as we would like to
```
timj, line 432: The "as" doesn't fit here.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            Other keyword parameters to set in the metadata.
+        """
+        kwargs["DETECTOR"] = self._detectorName
+        kwargs["DETECTOR_SERIAL"] = self._detectorSerial
+
+        kwargs['REGISTRY_NAME'] = self.registryName
+        kwargs['calibType'] = self.calibType
+        super().updateMetadata(self, setDate=setDate, **kwargs)
+
+    def fromDataIds(self, dataIdList, strict=False):
```
timj, line 490: strict parameter is not used.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+    Provenance is not really a calibration, but as we would like to
+    record this when constructing the calibration, and it provides an
+    example of the base calibration class.
+
+    Parameters
+    ----------
+    registryName : `str`, optional
+        Name of the butler registry the dataIds were obtained from.
+    calibType : `str`, optional
+        Type of calibration this provenance was generated for.
```
timj, line 441: Is this the gen3 dataset type name?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+class IsrProvenance(IsrCalib):
+    """Class for the provenance of data used to construct calibration.
+
+    Provenance is not really a calibration, but as we would like to
+    record this when constructing the calibration, and it provides an
+    example of the base calibration class.
+
+    Parameters
+    ----------
+    registryName : `str`, optional
```
timj, line 438: What is a registry name?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+                    self.dimensions.add(key)
+            self.dataIdList.append(dataId)
+
+    @classmethod
+    def fromTable(cls, tableList):
+        """Construct provenance from table list.
+
+        Parameters
+        ----------
+        tableList : `list` [`lsst.afw.table.Table`]
```
timj, line 510: Need a description string below

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,630 @@
...
+            inDict['dimensions'].add(key.getField().getName().lower())
+
+        for row in table:
+            entry = dict()
+            for dim in inDict['dimensions']:
+                key = schema[dim.upper()].asKey()
+                entry[dim] = row[key]
+            inDict['dataIdList'].append(entry)
+
+        return cls().fromDict(inDict)
```
timj, line 539: `cls` not `cls()`

python/lsst/ip/isr/calibType.py
czwa: Unused due to typo in IsrProvenance.

python/lsst/ip/isr/calibType.py
czwa: This was less obvious than I had hoped, but I've excised afwTable from this ticket.

python/lsst/ip/isr/calibType.py
czwa: I've changed this to raise if there is more than a single table, because not writing something you had is certainly an error.
I think we have to support multi-table things for Linearizer (Brighter-Fatter I think also is a place where this may happen).  I would prefer not to hide it in the metadata, but I don't have a strong feeling.

python/lsst/ip/isr/calibType.py
czwa: That's the idea that I had.  I don't know that it will be populated for every type (such as the IsrProvenance), but if there's a simple obvious way to apply the calibration, it would be good to have them use the same method name.

python/lsst/ip/isr/calibType.py
czwa: My thought was OtherCalib._OBSTYPE, but I would expect that to be identical to the gen3 dataset_type_name.

python/lsst/ip/isr/calibType.py
czwa: I had intended this to be some way to identify the Butler repository that constructed the calibration, to ensure all the dataIds recorded could be transformed if needed.  This now seems unnecessary, as any Butler instance should get the same dataIds for the same data from astro_metadata_translator.
I'm changing this to instrument, as it still feels like we need some additional information in the provenance, and I think that's provided by that.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,600 @@
+#
+# LSST Data Management System
+# Copyright 2008-2017 AURA/LSST.
```
timj, line 3: Can you use the new prologue for new files please?

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,600 @@
...
+            Raised if filename does not end in a known extension, or
+            if all information cannot be written.
+
+        Notes
+        -----
+        The file is written to YAML/ECSV format and will include any
+        associated metadata.
+
+        """
+        if format == 'yaml' or (format == 'auto' and filename.lower().endswith((".yaml", ".YAML"))):
```
timj, line 220: I think this means that I can pass in a file name of "hello.ecsv" and say "format=yaml" and end up with a yaml file that has a .ecsv extension.  In Defects I force the extension to be correct and I think you have to do that here -- it's why the returned filename doesn't have to match the supplied filename.

python/lsst/ip/isr/calibType.py
```diff
@@ -0,0 +1,600 @@
...
+        Returns
+        -------
+        provenance : `lsst.ip.isr.IsrProvenance`
+            The provenance defined in the tables.
+        """
+        calib = cls()
+        calib.updateMetadata(setDate=False, **dictionary['metadata'])
+        calib._detectorName = dictionary['detectorName']
+        calib._detectorSerial = dictionary['detectorSerial']
+        calib.instrument = dictionary['metadata']['INSTRUME']
```
timj, line 551: In fromTable doesn't an "instrument" field get filled in to the dict so I'm not sure why we need to look in the metadata here.

czwa:
Running a timing test shows that the new astropy code is significantly faster than afwTable in 100 iterations with timeit:
afwTable (ip_isr@cf0b964): read/write: 21.2s; write only: 10.1s
astropy(ip_isr@tickets/DM-24537): r/w: 8.3s; w: 5.7s

PR #136, opened 2020-04-24
DM-23369: Remove lsst.ip.isr.addDistortionModel

jdswinbank:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1086,8 +1077,9 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None, crosstalkSou
             The raw exposure that is to be run through ISR.  The
             exposure is modified by this method.
         camera : `lsst.afw.cameraGeom.Camera`, optional
-            The camera geometry for this exposure.  Used to select the
-            distortion model appropriate for this data.
+            The camera geometry for this exposure. Required if ``isGen3`` is
+            `True` and one or more of ``ccdExposure``, ``bias``, ``dark``, or
+            ``flat`` does not have an associated detector.
```
parejkoj, line 1082: Oof, this is rather tortured logic for this parameter. It's not in scope for this ticket, but that sounds like it should be cleaned up somehow.

czwa: It should probably be required in all cases to be explicit.

PR #138, opened 2020-04-29
DM-21711: Remove cameraGeom interfaces deprecated in DM-18610

jdswinbank:


python/lsst/ip/isr/isrTask.py
```diff
@@ -2449,9 +2446,6 @@ def getBBox(self):
     def getRawBBox(self):
         return self._bbox
 
-    def getHasRawInfo(self):
-        return True                     # but see getRawHorizontalOverscanBBox()
```
kfindeisen, line 2453: It may be worth investigating whether this `FakeAmp` class is still needed. Even if the functionality is still used, there may be a real `Amplifier` object that could do it.

PR #140, opened 2020-05-18
DM-23983: Cannot apply crosstalk in Gen 3 DECam processing

czwa:
Implement CrosstalkCalib class inheriting from IsrCalib. 

Add inter-chip crosstalk handling.
Switch to Exposure (no F) as the output product in gen3 to allow IsrTask to process an input in multiple passes.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -261,9 +261,10 @@ class IsrCalib(abc.ABC):
         tableList.append(Table.read(filename, hdu=1))
         extNum = 2  # Fits indices start at 1, we've read one already.
         try:
-            newTable = Table.read(filename, hdu=extNum)
-            tableList.append(newTable)
-            extNum += 1
+            with warnings.catch_warnings("error"):
+                newTable = Table.read(filename, hdu=extNum)
+                tableList.append(newTable)
+                extNum += 1
         except Exception:
             pass
 
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -25,6 +25,7 @@ import os.path
 import warnings
 import yaml
 from astropy.table import Table
+from astropy.io import fits
 
 from lsst.log import Log
 from lsst.daf.base import PropertyList
@@ -260,8 +261,7 @@ class IsrCalib(abc.ABC):
         tableList.append(Table.read(filename, hdu=1))
         extNum = 2  # Fits indices start at 1, we've read one already.
         try:
-            with warnings.catch_warnings("error"):
-                newTable = Table.read(filename, hdu=extNum)
+            newTable = Table.read(filename, hdu=extNum)
             tableList.append(newTable)
             extNum += 1
         except Exception:
@@ -284,13 +284,13 @@ class IsrCalib(abc.ABC):
 
         """
         tableList = self.toTable()
-
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", category=Warning, module="astropy.io")
-            tableList[0].write(filename)
-            for table in tableList[1:]:
-                table.write(filename, append=True)
+            astropyList = [fits.table_to_hdu(table) for table in tableList]
+            astropyList.insert(0, fits.PrimaryHDU())
 
+            writer = fits.HDUList(astropyList)
+            writer.writeto(filename, overwrite=True)
         return filename
 
     def fromDetector(self, detector):
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -142,7 +142,7 @@ class IsrCalib(abc.ABC):
         ----------
         setDate : `bool`, optional
             Ensure the metadata CALIBDATE fields are set to the current datetime.
-        kwargs :
+        kwargs : `dict` or `collections.abc.Mapping`, optional
             Set of key=value pairs to assign to the metadata.
         """
         mdOriginal = self.getMetadata()
@@ -474,7 +474,7 @@ class IsrProvenance(IsrCalib):
         setDate : `bool, optional
             Update the CALIBDATE fields in the metadata to the current
             time. Defaults to False.
-        kwargs :
+        kwargs : `dict` or `collections.abc.Mapping`, optional
             Other keyword parameters to set in the metadata.
         """
         kwargs["DETECTOR"] = self._detectorName
--- a/python/lsst/ip/isr/crosstalk.py
+++ b/python/lsst/ip/isr/crosstalk.py
@@ -23,15 +23,439 @@
 Apply intra-detector crosstalk corrections
 """
 import numpy as np
+from astropy.table import Table
 
 import lsst.afw.math
-import lsst.afw.table
 import lsst.afw.detection
 from lsst.pex.config import Config, Field, ChoiceField, ListField
 from lsst.pipe.base import Task
 
-__all__ = ["CrosstalkConfig", "CrosstalkTask", "subtractCrosstalk", "writeCrosstalkCoeffs",
+from lsst.ip.isr import IsrCalib
+
+
+__all__ = ["CrosstalkCalib", "CrosstalkConfig", "CrosstalkTask",
            "NullCrosstalkTask"]
+
+
+class CrosstalkCalib(IsrCalib):
+    """Calibration of amp-to-amp crosstalk coefficients.
+
+    Parameters
+    ----------
+    detector : `lsst.afw.cameraGeom.Detector`, optional
+        Detector to use to pull coefficients from.
+    log : `lsst.log.Log`, optional
+        Log to write messages to.
+    **kwargs :
+        Parameters to pass to parent constructor.
+    """
+    _OBSTYPE = 'CROSSTALK'
+    _SCHEMA = 'Gen3 Crosstalk'
+    _VERSION = 1.0
+
+    def __init__(self, detector=None, **kwargs):
+        self.hasCrosstalk = False
+        self.nAmp = 0
+        self.crosstalkShape = (self.nAmp, self.nAmp)
+
+        self.coeffs = None
+        self.coeffErr = None
+        self.coeffNum = None
+        self.interChip = None
+
+        super().__init__(**kwargs)
+        self.requiredAttributes.update(['hasCrosstalk', 'nAmp', 'coeffs',
+                                        'coeffErr', 'coeffNum', 'interChip'])
+        if detector:
+            self.fromDetector(detector)
+
+    def updateMetadata(self, setDate=False, **kwargs):
+        """Update calibration metadata.
+
+        This calls the base class's method after ensuring the required
+        calibration keywords will be saved.
+
+        Parameters
+        ----------
+        setDate : `bool`, optional
+            Update the CALIBDATE fields in the metadata to the current
+            time. Defaults to False.
+        kwargs :
+            Other keyword parameters to set in the metadata.
+        """
+        kwargs['DETECTOR'] = self._detectorName
+        kwargs['DETECTOR_SERIAL'] = self._detectorSerial
+        kwargs['HAS_CROSSTALK'] = self.hasCrosstalk
+        kwargs['NAMP'] = self.nAmp
+        self.crosstalkShape = (self.nAmp, self.nAmp)
+        kwargs['CROSSTALK_SHAPE'] = self.crosstalkShape
+
+        super().updateMetadata(setDate=setDate, **kwargs)
+
+    def fromDetector(self, detector, coeffVector=None):
+        """Set calibration parameters from the detector.
+
+        Parameters
+        ----------
+        detector : `lsst.afw.cameraGeom.Detector`
+            Detector to use to set parameters from.
+        coeffVector : `numpy.array`, optional
+            Use the geometry in the detector, but supply the
+            coefficients from elsewhere.
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.CrosstalkCalib`
+            The calibration constructed from the detector.
+        """
+        if detector.hasCrosstalk() or coeffVector:
+            self._detectorName = detector.getName()
+            self._detectorSerial = detector.getSerial()
+
+            self.nAmp = len(detector)
+            self.crosstalkShape = (self.nAmp, self.nAmp)
+
+            if coeffVector is not None:
+                crosstalkCoeffs = coeffVector
+            else:
+                crosstalkCoeffs = detector.getCrosstalk()
+            if len(crosstalkCoeffs) == 1 and crosstalkCoeffs[0] == 0.0:
+                return self
+            self.coeffs = np.array(crosstalkCoeffs).reshape(self.crosstalkShape)
+
+            if self.coeffs.shape != self.crosstalkShape:
+                raise RuntimeError("Crosstalk coefficients do not match detector shape. "
+                                   f"{self.crosstalkShape} {self.nAmp}")
+
+            self.interChip = {}
+            self.hasCrosstalk = True
+            self.updateMetadata()
+        return self
+
+    @classmethod
+    def fromDict(cls, dictionary):
+        """Construct a calibration from a dictionary of properties.
+
+        Must be implemented by the specific calibration subclasses.
+
+        Parameters
+        ----------
+        dictionary : `dict`
+            Dictionary of properties.
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.CalibType`
+            Constructed calibration.
+
+        Raises
+        ------
+        RuntimeError :
+            Raised if the supplied dictionary is for a different
+            calibration.
+        """
+        calib = cls()
+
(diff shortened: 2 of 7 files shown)


python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -165,20 +167,36 @@ def parseAndRun(cls, *args, **kwargs):
...
+        calib.coeffErr = coeffErr
+        calib.coeffNum = coeffNum
+
         outputFileName = results.parsedCmd.outputFileName
         if outputFileName is not None:
             butler = results.parsedCmd.butler
             dataId = results.parsedCmd.id.idList[0]
-            dataId["detector"] = butler.queryMetadata("raw", ["detector"], dataId)[0]
 
+            # Rework to use new crosstalk calib.
```
mrawls, line 182: Please edit comment so it's a bit less wrong in N years when this isn't "new" anymore.

python/lsst/ip/isr/measureCrosstalk.py
```diff
@@ -165,20 +167,36 @@ def parseAndRun(cls, *args, **kwargs):
...
+            provenance._detectorName = det.getName()
+            provenance.fromDataIds(results.parsedCmd.id.idList)
+            provenance.writeText(outputFileName + '_prov.yaml')
 
         return Struct(
             coeff=coeff,
             coeffErr=coeffErr,
-            coeffNum=coeffNum
+            coeffNum=coeffNum,
+            calib=calib,
```
mrawls, line 199: Please add this returned value to the docstring, and while you're at it, maybe make the summary something moderately more informative than "implement scatter/gather" because I don't know what that means.

python/lsst/ip/isr/isrTask.py
```diff
@@ -46,18 +46,35 @@
...
 from .overscan import OverscanCorrectionTask
 from .straylight import StrayLightTask
 from .vignette import VignetteTask
+from lsst.daf.butler import DataCoordinate, DimensionGraph, DimensionUniverse
 
 
 __all__ = ["IsrTask", "IsrTaskConfig", "RunIsrTask", "RunIsrConfig"]
 
 
+def ctSourceLUF(datasetType, registry, quantumDataId, collections):
```
mrawls: Docstring please, and consider renaming the function with fewer acronyms (I can guess ct is crosstalk but am not sure what LUF is) ... unless ct is connectionType which has been imported as cT ... why

python/lsst/ip/isr/calibType.py
```diff
@@ -262,8 +263,8 @@ def readFits(cls, filename):
         try:
             with warnings.catch_warnings("error"):
                 newTable = Table.read(filename, hdu=extNum)
-            tableList.append(newTable)
-            extNum += 1
+                tableList.append(newTable)
+                extNum += 1
```
mrawls, line 267: Does this actually change any behavior?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+        This calls the base class's method after ensuring the required
+        calibration keywords will be saved.
+
+        Parameters
+        ----------
+        setDate : `bool`, optional
+            Update the CALIBDATE fields in the metadata to the current
+            time. Defaults to False.
+        kwargs :
+            Other keyword parameters to set in the metadata.
```
mrawls, line 84: Should this be type `dict`? I've never seen kwargs documented quite like this.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+    def fromDetector(self, detector, coeffVector=None):
+        """Set calibration parameters from the detector.
+
+        Parameters
+        ----------
+        detector : `lsst.afw.cameraGeom.Detector`
+            Detector to use to set parameters from.
+        coeffVector : `numpy.array`, optional
+            Use the geometry in the detector, but supply the
+            coefficients from elsewhere.
```
mrawls, line 104: You mean if this is not provided, it uses detector geometry, but if it is, it uses the coefficients in `coeffVector`? Not clear to me as written.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+        Returns
+        -------
+        tableList : `list` [`lsst.afw.table.Table`]
+            List of tables containing the crosstalk calibration
+            information.
+
+        """
+        tableList = []
+        self.updateMetadata()
+        catalog = Table([{'CT_COEFFS': self.coeffs.reshape(self.nAmp*self.nAmp)}])
```
mrawls, line 271: I agree with you in principle for using astropy Table instead of afwTable, but it's worth asking if we sacrifice anything in speed by using astropy?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+        if self.interChip:
+            interChipTable = Table([{'IC_SOURCE_DET': sourceDet,
+                                     'IC_COEFFS': self.interChip[sourceDet].reshape(self.nAmp*self.nAmp)}
+                                    for sourceDet in self.interChip.keys()])
+            tableList.append(interChipTable)
+        return tableList
+
+    # Implementation methods.
+    def extractAmp(self, image, amp, ampTarget, isTrimmed=False):
+
```
mrawls, line 284: Extra blank row.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+        image : `lsst.afw.image.Image` or `lsst.afw.image.MaskedImage`
+            Image containing the amplifier of interest.
+        amp : `lsst.afw.cameraGeom.Amplifier`
+            Amplifier on image to extract.
+        ampTarget : `lsst.afw.cameraGeom.Amplifier`
+            Target amplifier that the extracted image will be flipped
+            to match.
+        isTrimmed : `bool`
+            The image is already trimmed.
+            This should no longer be needed once DM-15409 is resolved.
```
mrawls, line 298: Please add an explicit TODO here (per https://developer.lsst.io/python/style.html?highlight=todo#to-do-comments-should-include-a-jira-issue-key)

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+
+        output = image[amp.getBBox() if isTrimmed else amp.getRawDataBBox()]
+        thisAmpCorner = amp.getReadoutCorner()
+        targetAmpCorner = ampTarget.getReadoutCorner()
+
+        # Flipping is necessary only if the desired configuration doesn't match what we currently have
+        xFlip = X_FLIP[targetAmpCorner] ^ X_FLIP[thisAmpCorner]
+        yFlip = Y_FLIP[targetAmpCorner] ^ Y_FLIP[thisAmpCorner]
+        self.log.debug("Extract amp: %s %s %s %s",
+                       amp.getName(), ampTarget.getName(), thisAmpCorner, targetAmpCorner)
```
mrawls, line 322: f-strings preferred

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+        return lsst.afw.math.flipImage(output, xFlip, yFlip)
+
+    def calculateBackground(self, mi, badPixels=["BAD"]):
+        """Calculate median background in image
```
mrawls, line 326: Maybe call this a rough estimate instead of a calculation?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+    def subtractCrosstalk(self, thisExposure, sourceExposure=None, crosstalkCoeffs=None,
+                          badPixels=["BAD"], minPixelToMask=45000,
+                          crosstalkStr="CROSSTALK", isTrimmed=False,
+                          backgroundMethod="None"):
+        """Subtract the crosstalk from thisExposure, optionally using a different source.
+
+        We set the mask plane indicated by ``crosstalkStr`` in a target amplifier
+        for pixels in a source amplifier that exceed `minPixelToMask`. Note that
+        the correction is applied to all pixels in the amplifier, but only those
+        that have a substantial crosstalk are masked with ``crosstalkStr``.
```
mrawls, line 356: Inconsistencies in single vs. double backticks

python/lsst/ip/isr/crosstalk.py
```diff
@@ -23,17 +23,441 @@
...
+
+        The uncorrected image is used as a template for correction. This is good
+        enough if the crosstalk is small (e.g., coefficients < ~ 1e-3), but if it's
+        larger you may want to iterate.
+
+        This method needs unittests (DM-18876), but such testing requires
```
mrawls, line 362: Again, add a TODO please

python/lsst/ip/isr/crosstalk.py
```diff
@@ -167,255 +600,53 @@ def run(self, exposure, crosstalkSources=None, isTrimmed=False):
...
-            else:
-                raise TypeError("crosstalkSources not of the correct type: `np.array` or `dict`")
+        if not crosstalk:
+            crosstalk = CrosstalkCalib(log=self.log)
+            crosstalk = crosstalk.fromDetector(exposure.getDetector(),
+                                               coeffVector=self.config.crosstalkValues)
+            # detector = exposure.getDetector()
+            # if not self.config.hasCrosstalk(detector=detector):
+            # raise RuntimeError("Attempted to correct crosstalk without crosstalk coefficients")
+            # coeffs = self.config.getCrosstalk(detector=detector)
```
mrawls, line 611: What's happening here?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -167,255 +600,53 @@ def run(self, exposure, crosstalkSources=None, isTrimmed=False):
...
+        if not crosstalk.log:
+            crosstalk.log = self.log
+        if not crosstalk.hasCrosstalk:
+            raise RuntimeError("Attempted to correct crosstalk without crosstalk coefficients.")
+            self.log.warn("Attempted to correct crosstalk without crosstalk coefficients.")
```
mrawls, line 616: Is this redundant?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -167,255 +600,53 @@ def run(self, exposure, crosstalkSources=None, isTrimmed=False):
...
+                                        backgroundMethod=self.config.crosstalkBackgroundMethod)
+
+            if crosstalk.interChip:
+                if crosstalkSources:
+                    for detName in crosstalk.interChip:
+                        if isinstance(crosstalkSources[0], 'lsst.afw.image.Exposure'):
+                            # Received a proper exposure
+                            sourceNames = [exp.getDetector().getName() for exp in crosstalkSources]
+                        else:
+                            # Received deferred dataset
```
mrawls, line 631: Not immediately clear what "deferred dataset" means here

python/lsst/ip/isr/isrTask.py
```diff
@@ -836,6 +890,18 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
 
         detector = inputs['ccdExposure'].getDetector()
 
+        if self.config.doCrosstalk is True:
+            # Crosstalk sources need to be defined by the pipeline
+            # yaml if they exist.
```
mrawls, line 895: Perhaps add some kind of `else` catch for if doCrosstalk is requested but no sources are defined?

tests/test_crosstalk.py
```diff
@@ -188,12 +188,18 @@ def testDirectAPI(self):
         ratios = measure.extractCrosstalkRatios(self.exposure, threshold=self.value - 1)
         coeff, coeffErr, coeffNum = measure.measureCrosstalkCoefficients(ratios)
         self.checkCoefficients(coeff, coeffErr, coeffNum)
-        subtractCrosstalk(self.exposure, minPixelToMask=self.value - 1,
-                          crosstalkStr=self.crosstalkStr)
+        calib = CrosstalkCalib()
+        calib.coeffs = coeff.transpose()
+        calib.coeffErr = coeffErr.transpose()
+        calib.coeffNum = coeffNum.transpose()
```
mrawls, line 194: There is a moderately terrifying amount of transposing happening here. If it is happening in the test I assume it is happening elsewhere too, does it affect runtime significantly?

python/lsst/ip/isr/measureCrosstalk.py
czwa: Done.  I've also noted that this method will disappear in DM-24760, as the current logic is confusing at best.

python/lsst/ip/isr/calibType.py
czwa: The `with warnings.catch_warnings("error")` is needed because astropy will read the "first available" table if the requested table does not exist, throwing a warning instead of an exception.  This promotes the warning to an exception, and ensures that the list represents what's in the FITS table.

python/lsst/ip/isr/crosstalk.py
czwa: Updated with `dict` and `collections.abc.Mappings`, based on the upstream version: https://github.com/lsst/daf_base/blob/master/python/lsst/daf/base/propertyContainer/propertyContainerContinued.py#L469.

python/lsst/ip/isr/crosstalk.py
czwa: The astropy implementation is faster than afwTable, for reasons I don't completely remember.  I believe it has something to do with how afwTable stores the data to allow for row inserts.  My short timing test for DM-24537 confirmed this:  https://github.com/lsst/ip_isr/pull/135#issuecomment-620847931

timj: Yes, we need to switch all our curated calibrations over to astropy table. It's faster and @TallJimbo indicated that we should only be using afwTable if we needed to pass the tables to C++ (which we don't).

python/lsst/ip/isr/crosstalk.py
czwa: I thought they were not for lsst.Log messages, as the log only applies the formatting if the log level triggers?

python/lsst/ip/isr/crosstalk.py
czwa: Removed as those tickets have been resolved, and updated unittests were added previously.

python/lsst/ip/isr/crosstalk.py
czwa: Removed.

python/lsst/ip/isr/crosstalk.py
czwa: Switched this and previous comment to use abbreviated python class to clarify.  This now points to dafButler.DeferredDatasetHandle.

python/lsst/ip/isr/isrTask.py
czwa: I've added a log warning for the case where no crosstalkSources exist but there are interChip terms in the calibration.  The original code didn't raise anything for interChip coefficients that don't have matching sources, so I'm wary against breaking something by raising here, but you're right that it should say something so it's not a silent failure.

tests/test_crosstalk.py
czwa: Moving MeasureCrosstalkTask to cp_pipe, including this test, is on my list as DM-24760.  The transpose here is a result of me normalizing the crosstalk coefficient order everywhere except there.  The only place a transpose should now be happening is in directly using the output of MeasureCrosstalkTask, and I believe this test is the only place that happens.

python/lsst/ip/isr/crosstalk.py
r-owen: My two bits, as an kibbitzer: I agree that deferred formatting can be helpful in log messages, especially if formatting is expensive and the messages are unlikely to be output (e.g. debug level, as here).

That said, I agree with @mrawls that f strings are better for log messages that are simple or usually output (e.g. info and warning level), because the code is a always exercised and you don't have to wonder if the formatting is correct.

czwa: I'm going to leave this one alone (partially to avoid having to deal with a long line), and use f-strings for info and warning going forward.
I may try to bother someone about getting a definite answer put into the style guide, so I can either get it right or be consistently wrong until I get it right.

python/lsst/ip/isr/calibType.py
mrawls: Wow, that is subtle. Sounds good!

python/lsst/ip/isr/crosstalk.py
mrawls: Didn't mean to open a pandora's box about string formatting here 😄 not a big deal in this particular instance, but agree it should be in the style guide one way or another.

tests/test_crosstalk.py
mrawls: Oh that's interesting, it hadn't occurred to me to have **measure** crosstalk task somewhere else, but I agree, that's where it "should" ultimately land. Sounds good.

PR #141, opened 2020-05-29
DM-25149: Fix w18 ISR regression

czwa:
Mask saturated overscan pixels before fitting.

### DIFF
--- a/python/lsst/ip/isr/overscan.py
+++ b/python/lsst/ip/isr/overscan.py
@@ -59,6 +59,11 @@ class OverscanCorrectionTaskConfig(pexConfig.Config):
         doc="Rejection threshold (sigma) for collapsing overscan before fit",
         default=3.0,
     )
+    maskPlanes = pexConfig.ListField(
+        dtype=str,
+        doc="Mask planes to reject when measuring overscan",
+        default=['SAT'],
+    )
     overscanIsInt = pexConfig.Field(
         dtype=bool,
         doc="Treat overscan as an integer image for purposes of fitType=MEDIAN" +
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/overscan.py
```diff
@@ -249,6 +249,9 @@ def getImageArray(image):
         """
         if hasattr(image, "getImage"):
             calcImage = image.getImage().getArray()
+            mask = image.getMask()
+            calcImage = np.ma.masked_where(mask.getArray() & mask.getPlaneBitMask("SAT"),
```
yalsayyad, line 253: I don't understand why "SUSPECT" AND "SAT" need to be handled in different parts in the task, or why this is a static method, or "SAT" needs to be hard-coded here.  Since you do, will you add an inline comment explaining?

 I would have done the following: which fixes the problem and passes unit tests: 

`self.statControl.setAndMask(afwImage.Mask.getPlaneBitMask(["SAT", "SUSPECT"]))` in line 92 and  (with ["SAT", "SUSPECT"] as a config parameter,  and the following as the getImageArray()
```

    def getImageArray(self, image):
        """Extract the numpy array from the input image.

        Parameters
        ----------
        image : `lsst.afw.image.Image` or `lsst.afw.image.MaskedImage`
            Image data to pull array from.

        calcImage : `numpy.ndarray`
            Image data array for numpy operating.
        """
        if hasattr(image, "getImage"):
            calcImage = image.getImage().getArray()
            mask = image.getMask()
            import pdb; pdb.set_trace()
            calcImage = np.ma.masked_where(mask.getArray() & self.statControl.getAndMask(),
                                           calcImage)
        else:
            calcImage = image.getArray()
        return calcImage
```

czwa: Switched to this form, adding config.maskPlanes that defaults to ['SAT'] (to match the previous behavior).  This adds flexibility for the future.

PR #146, opened 2020-07-10
Fix Exception raised in docstring

sr525:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -484,7 +484,7 @@ def overscanCorrection(ampMaskedImage, overscanImage, fitType='MEDIAN', order=1,
             (`lsst.afw.image.Image`)
     Raises
     ------
-    pexExcept.Exception
+    RuntimeError
```
mfisherlevine, line 487: I'd suggest `ValueError` for that, but either way is clearly an improvement :)

timj: Well this is documentation so the question is what happens when a config item is assigned a value that is not allowed. Is it a RuntimeError or a ValueError? It's not open to debate since the code is the code. It's clearly not a pex exception anyhow.

mfisherlevine: They don't have a freeform emoji response here, but if they did, I'd have used 🔥 , fwiw.

PR #149, opened 2020-08-05
DM-26217: Edge masking in IsrTask needs a "level" option to mask detector edges or amplifier edges.

czwa:
Add level at which to do Edge masking.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -366,6 +366,15 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         doc="Number of edge pixels to be flagged as untrustworthy.",
         default=0,
     )
+    edgeMaskLevel = pexConfig.ChoiceField(
+        dtype=str,
+        doc="Mask edge pixels in which coordinate frame: DETECTOR or AMP?",
+        default="DETECTOR",
+        allowed={
+            'DETECTOR': 'Mask only the edges of the full detector.',
+            'AMP': 'Mask edges of each amplifier.',
+        },
+    )
 
     # Initial masking options.
     doSetBadRegions = pexConfig.Field(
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -2244,20 +2253,28 @@ def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT"):
...
+            for box in boxes:
+                # This makes a bbox numEdgeSuspect pixels smaller than the image on each side
+                subImage = maskedImage[box]
+                box.grow(-numEdgePixels)
+                # Mask pixels outside box
+                SourceDetectionTask.setEdgeBits(
+                    subImage,
+                    box,
+                    maskBitMask
+                )
```
mfisherlevine, line 2277: Could you move this closing paren up a line so it comes after `maskBitMask`?

python/lsst/ip/isr/isrTask.py
```diff
@@ -2276,7 +2293,7 @@ def maskAndInterpolateDefects(self, exposure, defectBaseList):
         """
         self.maskDefect(exposure, defectBaseList)
         self.maskEdges(exposure, numEdgePixels=self.config.numEdgeSuspect,
-                       maskPlane="SUSPECT")
+                       maskPlane="SUSPECT", level=self.config.edgeMaskLevel)
         isrFunctions.interpolateFromMask(
             maskedImage=exposure.getMaskedImage(),
             fwhm=self.config.fwhm,
```
mfisherlevine, line 2299: I have problems with this whole function, but not your changes to it. But while looking at it I can't help but wonder what part of `maskAndInterpolateDefects()` implies that edges will also be masked. Neither the function name, nor the docstring, mention that this will happen.

python/lsst/ip/isr/isrTask.py
```diff
@@ -366,6 +366,15 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         doc="Number of edge pixels to be flagged as untrustworthy.",
         default=0,
     )
+    edgeMaskLevel = pexConfig.ChoiceField(
+        dtype=str,
+        doc="Which level to mask edge values at?",
```
mfisherlevine, line 371: I feel like this could be a little more descriptive. Maybe something like "Mask edge pixels around the perimeter of the detector, or around each amplifier?" or something like that, otherwise it's a little tautological.

python/lsst/ip/isr/isrTask.py
czwa: I don't believe this function is called directly anymore.  `ip_isr` probably needs a cleanup ticket to get rid of outdated code.

PR #151, opened 2020-08-14
DM-26316: Allow bias generation to retain overscan signal

czwa:
Add option to reverse Bias and Overscan correction ordering.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -847,6 +847,8 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         super().validate()
         if self.doFlat and self.doApplyGains:
             raise ValueError("You may not specify both doFlat and doApplyGains")
+        if self.doBiasBeforeOverscan and self.doTrimToMatchCalib:
+            raise ValueError("You may not specify both doBiasBeforeOverscan and doTrimToMatchCalib")
         if self.doSaturationInterpolation and "SAT" not in self.maskListToInterpolate:
             self.config.maskListToInterpolate.append("SAT")
         if self.doNanInterpolation and "UNMASKEDNAN" not in self.maskListToInterpolate:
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -1325,6 +1330,12 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None,
             self.log.info("Converting exposure to floating point values.")
             ccdExposure = self.convertIntToFloat(ccdExposure)
 
+        if self.config.doBias and self.config.doBiasBeforeOverscan:
+            self.log.info("Applying bias correction.")
+            isrFunctions.biasCorrection(ccdExposure.getMaskedImage(), bias.getMaskedImage(),
```
RobertLuptonTheGood, line 1335: We decided not to add protocol numbers for now, so I thought we were relying on the sizes of the biases to detect whether the master biases were built with `isr.doOverscan=False isr.doAssembleCcd=False isr.doBiasBeforeOverscan=True`.   If that's right, wouldn't `config.doTrimToMatchCalib` defeat the check?  Maybe validate should check that it's not set with `isr.doBiasBeforeOverscan`

czwa: We are not relying on the sizes to detect the master bias construction.  The configuration files for `constructBias.py` and `runIsr.py/processCcd.py` have to be consistent for now.  I've added the check in `validate`, because although we don't do automatic detection, those two config options should be mutually exclusive.

RobertLuptonTheGood: Will it fail if the sizes don't match?  If so, maybe we could have a more helpful error message

czwa: The `isrFunctions.biasCorrection` raises with a message if the bounding boxes do not match:
```
raise RuntimeError("maskedImage bbox %s != biasMaskedImage bbox %s" %
                           (maskedImage.getBBox(afwImage.LOCAL), biasMaskedImage.getBBox(afwImage.LOCAL)))

```

PR #152, opened 2020-08-25
DM-26452: Fix fringe filter inconsistency

czwa:
Use the same filter name determination in all places that the fringe is checked.
Add fringe support for matching against filter name aliases.

### DIFF
--- a/python/lsst/ip/isr/fringe.py
+++ b/python/lsst/ip/isr/fringe.py
@@ -54,6 +54,7 @@ class FringeStatisticsConfig(Config):
 class FringeConfig(Config):
     """Fringe subtraction options"""
     filters = ListField(dtype=str, default=[], doc="Only fringe-subtract these filters")
+    useFilterAliases = Field(dtype=bool, default=False, doc="Search filter aliases during check.")
     num = Field(dtype=int, default=30000, doc="Number of fringe measurements")
     small = Field(dtype=int, default=3, doc="Half-size of small (fringe) measurements (pixels)")
     large = Field(dtype=int, default=30, doc="Half-size of large (background) measurements (pixels)")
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/fringe.py
```diff
@@ -248,7 +249,12 @@ def checkFilter(self, exposure):
             If True, then the exposure has a filter listed in the
             configuration, and should have the fringe applied.
         """
-        return exposure.getFilter().getName() in self.config.filters
+        filterObj = afwImage.Filter(exposure.getFilter().getId())
+        if self.config.useFilterAliases:
```
SimonKrughoff, line 253: Why not always include the aliases in the list being checked?  It seems unlikely to cause a problem.  Are you worried there will be a collision between a physical filter name and an alias that would cause confusion?

czwa: Avoiding possible collisions was one issue.  I also wanted to keep the default strict so that the code does the least surprising thing.
I'm also starting to wonder if this is an issue elsewhere, and if so whether the solution isn't a method on the `Filter` itself.  `myFilter.isThisMe('HSC-y')` or something.

timj: You might want to talk to @parejkoj about requirements for new Filter implementation.

parejkoj: `getId()` will probably be going away. Also, the plan is to get rid of basically all aliases. Everything used in ISR will be the physical filter name (e.g. `HSC-R` or `HSC-R2`), full stop.

What is the failure mode being fixed here?

That said, I think we'll have a variety of things to clean up, so this will just be another one. But this does seem somewhat excessively complicated.

czwa: >  Everything used in ISR will be the physical filter name (e.g. HSC-R or HSC-R2), full stop.
That works for me.  Enforced consistency using the physical filter name makes the ISR checks much simpler. The issue here is that `exposure.getFilter().getName()` and `afwImage.Filter(exposure.getFilter().getId()).getName()` yield different results.  This caused the top level `IsrTask` check to pass (I think there the filter claimed to be 'y', matching the config file), but the internal `FringeTask` check to fail (the filter there claimed to be 'HSC-y', causing the mismatch).

parejkoj: Can you change the config file to use `HSC-Y` (note: all caps) instead of `y`? That would future proof it, and also make this extra code unnecessary.

czwa: The fix was in the first commit, which probably would be needed in any case (the inverse case is that the `IsrTask` check fails, and no fringe is applied when it should be).  Adding the potential all-alias match was a suggestion, and implemented on a second commit.  I have a passing Jenkins build, so I'm going to call this done, and will revisit when the `Filter` changes hit.

PR #154, opened 2020-08-26
DM-24703: Make linearity a subclass of lsst.ip.isr.IsrCalib

czwa:
This makes the linearity dataset a subclass of IsrCalib, which hopefully gives better code reuse and consistency between datasets.

### DIFF
--- a/python/lsst/ip/isr/linearize.py
+++ b/python/lsst/ip/isr/linearize.py
@@ -20,24 +20,21 @@
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
 import abc
-import copy
-import datetime
-
 import numpy as np
-import yaml
 
-import lsst.afw.table as afwTable
-from lsst.daf.base import PropertyList
+from astropy.table import Table
+
 from lsst.pipe.base import Struct
 from lsst.geom import Box2I, Point2I, Extent2I
 from .applyLookupTable import applyLookupTable
+from .calibType import IsrCalib
 
 __all__ = ["Linearizer",
            "LinearizeBase", "LinearizeLookupTable", "LinearizeSquared",
            "LinearizeProportional", "LinearizePolynomial", "LinearizeNone"]
 
 
-class Linearizer(abc.ABC):
+class Linearizer(IsrCalib):
     """Parameter set for linearization.
 
     These parameters are included in cameraGeom.Amplifier, but
@@ -51,12 +48,12 @@ class Linearizer(abc.ABC):
             - one column for each image value
         To avoid copying the table the last index should vary fastest
         (numpy default "C" order)
-    detector : `lsst.afw.cameraGeom.Detector`
-        Detector object
-    override : `bool`, optional
-        Override the parameters defined in the detector/amplifier.
+    detector : `lsst.afw.cameraGeom.Detector`, optional
+        Detector object.  Passed to self.fromDetector() on init.
     log : `lsst.log.Log`, optional
         Logger to handle messages.
+    kwargs : `dict`, optional
+        Other keyword arguments to pass to the parent init.
 
     Raises
     ------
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/linearize.py
```diff
@@ -20,24 +20,21 @@
...
 import abc
-import copy
-import datetime
-
 import numpy as np
-import yaml
 
-import lsst.afw.table as afwTable
-from lsst.daf.base import PropertyList
+from astropy.table import Table
```
plazas, line 25: Are we moving away from afwTable in favor of astropy.Table?

python/lsst/ip/isr/linearize.py
```diff
@@ -51,43 +48,65 @@ class Linearizer(abc.ABC):
...
+        Whether a linearity correction is defined for this detector.
+    override : `bool`
+        Whether the detector parameters should be overridden.
+    ampNames : `list` [`str`]
+        List of amplifier names to correct.
+    linearityCoeffs : `dict` [`str`, `numpy.array`]
+        Coefficients to use in correction.  Indexed by amplifier
+        names.  The format of the array depends on the type of
+        correction to apply.
+    linearityType : `dict` [`str`, `str`]
```
plazas, line 78: Having this as a dictionary would imply that, in principle, each amp would have a different linearity type. But i don't think that's currently the case, correct?

python/lsst/ip/isr/linearize.py
```diff
@@ -51,43 +48,65 @@ class Linearizer(abc.ABC):
...
+        Type of correction to use, indexed by amplifier names.
+    linearityBBox : `dict` [`str`, `lsst.geom.Box2I`]
```
plazas, line 80: What is the exact meaning of this bounding box? what are the regions where the correction is valid and where it is not? I'm a bit confused by the explanation n the docstring (sorry!)

python/lsst/ip/isr/linearize.py
timj: Yes for tables that are all in python and not needed to be in C++. Astropy table is faster than afw table for this use case.

python/lsst/ip/isr/linearize.py
czwa: Each amp can have a different linearity type, and the correction will apply whatever is defined.  We currently don't have a way to measure different linearity types simultaneously in `cp_pipe`, but such a calibration could be constructed if necessary.  This is largely a carryover from the `cameraGeom` defined linearity parameters, which were amplifier based.

python/lsst/ip/isr/linearize.py
czwa: Again, carryover from the `cameraGeom` definitions.  The individual amplifiers have been assembled and trimmed so this is used to define the region to apply that amplifier's correction.  This means that the correction can be run without passing a detector (which also defines the BBox).  The alternative is to make `linearizer.applyLinearity()` require the `detector` argument.

python/lsst/ip/isr/linearize.py
jdswinbank: @timj, @ktlim — if that's official project policy, can you please make sure it's captured in the Developer Guide? In particular, I think it directly contradicts the statement at https://developer.lsst.io/python/astropy.html that

> If the code is already using afw it is strongly preferred that afw equivalents be used until such time as specific afw interfaces are deprecated

timj: It was more a discovery by @czwa that he learned when he was writing these calibration tasks (which means Defects is slow at the moment and can be improved when it becomes a subclass of the new calibration base class). We haven't formally written that discovery down anywhere.

jdswinbank: That's fine, but I think we do need to make sure that this is captured as part of appropriate policy. “Performance concerns” alone are not an adequate reason to ignore the Dev Guide.

TallJimbo: I very much endorse the idea of making an exception to this rule for `afw.table`.  But I think this is an easy case to make that call on, and writing a policy to cover all cases (including the role of pandas vs. afw.table for Source/Object catalogs) is a lot harder.  I hope we can capture that action without blocking this ticket on it.

timj: I hope we aren't considering blocking this ticket. The performance hit was pretty bad if I recall correctly.

jdswinbank: This ticket doesn't seem to have anything to do with mitigating a performance hit; I don't really understand why that's coming up here.

However, I don't see any need to block this work, as long as the Architecture Team can reassure us that a) it doesn't actually contravene the Developer Guide, and b) they'll be updating the Developer Guide to provide appropriate guidance in future.

timj: This code is bringing this class into compliance with the pre-existing IsrCalib which is the thing that required astropy tables for performance.

jdswinbank: Ah; it's a shame we didn't take that opportunity to update the Dev Guide! :-)

PR #157, opened 2020-10-01
DM-26725: Update the PTC dataset to be a proper IsrCalib

plazas:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -0,0 +1,472 @@
...
+            Raised if the supplied dictionary is for a different
+            calibration.
+        """
+        calib = cls()
+        if calib._OBSTYPE != dictionary['metadata']['OBSTYPE']:
+            raise RuntimeError(f"Incorrect Photon Transfer Curve dataset  supplied. "
+                               f"Expected {calib._OBSTYPE}, found {dictionary['metadata']['OBSTYPE']}")
+
+        calib.setMetadata(dictionary['metadata'])
+        calib.ptcFitType = list(dictionary['ampNames'].values())[0]['ptcFitType']
```
czwa, line 222: Aren't ptcFitType and badAmps stored directly in the dictionary?

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -0,0 +1,472 @@
...
+
+        metadata = ptcTable.meta
+        inDict = dict()
+        inDict['metadata'] = metadata
+        inDict['ampNames'] = dict()
+
+        for record in ptcTable:
+            ampName = record['AMPLIFIER_NAME']
+
+            inDict['ampNames'][ampName] = {
```
czwa, line 359: This seems to be different than the format in `fromDict`.  Am I just misreading one of these methods?

czwa (changes requested):
Can you add a unit test that makes one of these, and then round trips it through the text and fits formats?  This is something I didn't do with linearity, and it's created problems.  Other than me potentially misunderstanding the `fromTable` method, I think the rest is just typo level things.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -0,0 +1,472 @@
...
+        Other keyword arguments to pass to the parent init.
+    Notes
+    -----
+    The stored attributes are:
+    badAmps : `list`
+        List with bad amplifiers names.
+    inputExpIdPairs : `dict`, [`str`, `list`]
+        Dictionary keyed by amp names containing the input exposures IDs.
+    expIdMask : `dict`, [`str`, `list`]
+        Dictionary keyed by amp names containing the mask produced after outlier rejection. The mask produced
```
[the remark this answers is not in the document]
plazas: Right at 110, but I have split them further.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -0,0 +1,472 @@
...
+            calib.rawVars.setdefault(ampName, amp['rawVars'])
+            calib.gain.setdefault(ampName, amp['gain'])
+            calib.gainErr.setdefault(ampName, amp['gainErr'])
+            calib.noise.setdefault(ampName, amp['noise'])
+            calib.noiseErr.setdefault(ampName, amp['noiseErr'])
+            calib.ptcFitPars.setdefault(ampName, amp['ptcFitPars'])
+            calib.ptcFitParsError.setdefault(ampName, amp['ptcFitParsError'])
+            calib.ptcFitChiSq.setdefault(ampName, amp['ptcFitChiSq'])
+            if calib.ptcFitType in ['FULLCOVARIANCE', ]:
+                calib.covariances.setdefault(ampName, amp['covariances'].reshape((nSignalPoints,
```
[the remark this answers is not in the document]
plazas: They are indeed stored as covArrays (they come from `astierCovPtcFit`). if it's a bad amp, there're also  NumPy arrays with nans.

python/lsst/ip/isr/ptcDataset.py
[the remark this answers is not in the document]
czwa: https://developer.lsst.io/python/style.html#docstring-and-comment-line-length-must-be-less-than-or-equal-to-79-columns
The link contains the important information.

python/lsst/ip/isr/ptcDataset.py
plazas: I changed it so that it is similar to the format in `fromDict`

czwa (approved):
I think this clarifies the parts that I was confused about before.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -0,0 +1,534 @@
...
+            calib.rawVars.setdefault(ampName, np.array(dictionary['rawVars'][ampName]).tolist())
+            calib.gain.setdefault(ampName, np.array(dictionary['gain'][ampName]).tolist())
+            calib.gainErr.setdefault(ampName, np.array(dictionary['gainErr'][ampName]).tolist())
+            calib.noise.setdefault(ampName, np.array(dictionary['noise'][ampName]).tolist())
+            calib.noiseErr.setdefault(ampName, np.array(dictionary['noiseErr'][ampName]).tolist())
+            calib.ptcFitPars.setdefault(ampName, np.array(dictionary['ptcFitPars'][ampName]).tolist())
+            calib.ptcFitParsError.setdefault(
+                ampName, np.array(dictionary['ptcFitParsError'][ampName]).tolist())
+            calib.ptcFitChiSq.setdefault(ampName, np.array(dictionary['ptcFitChiSq'][ampName]).tolist())
+            if calib.ptcFitType in ['FULLCOVARIANCE', ]:
```
czwa, line 282: This looks incorrect.  If the input dictionary contains a field, why would we drop it?

plazas: If fixed this as we discussed, and the {if} clause is no longer there (wasn't needed).

PR #158, opened 2020-10-07
DM-26946: readFits in calibType.ptc seems to fail to read more than one table

czwa:
Handle multi-extension tables correctly.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -305,13 +305,17 @@ class IsrCalib(abc.ABC):
         tableList = []
         tableList.append(Table.read(filename, hdu=1))
         extNum = 2  # Fits indices start at 1, we've read one already.
-        try:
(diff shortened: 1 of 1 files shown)


plazas (approved):
It looks good to me. The only comment I have is, it seems that it stops when it fails because there are no more tables to read. Could it be possible to have a few good tables, then a bad table, and then more good tables that won't get read because it quit at the first bad table that it encountered? Maybe this is something that actually doesn't happen.

czwa:
> It looks good to me. The only comment I have is, it seems that it stops when it fails because there are no more tables to read. Could it be possible to have a few good tables, then a bad table, and then more good tables that won't get read because it quit at the first bad table that it encountered? Maybe this is something that actually doesn't happen.

It's possible, but I think unlikely (and may only be possible if there's some file corruption).  The main issue is that astropy Table doesn't have access to the full HDUList, so there's no simple way to know how many tables there are.

PR #159, opened 2020-10-08
DM-26545: Add spline linearizer.

czwa:
This changeset adds the spline linearizer, and some additional metadata handling tools to try and enforce more uniform calibration headers.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -272,9 +272,6 @@ class IsrCalib(abc.ABC):
         self._filter = search(dictionary, ['FILTER', 'filterName'])
         self._calibId = search(dictionary, ['CALIB_ID'])
 
-        if not self._detectorId and self._detectorSerial:
-            self._detectorId = self._detectorSerial
-
     @classmethod
     def readText(cls, filename):
         """Read calibration representation from a yaml/ecsv file.
@@ -634,7 +631,6 @@ class IsrProvenance(IsrCalib):
         table = tableList[0]
         metadata = table.meta
         inDict = dict()
-
         inDict['metadata'] = metadata
         inDict['calibType'] = metadata['calibType']
         inDict['dimensions'] = set()
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -60,17 +60,18 @@ class IsrCalib(abc.ABC):
     _SCHEMA = 'NO SCHEMA'
     _VERSION = 0
 
-    def __init__(self, camera=None, detector=None, detectorName=None, detectorId=None, log=None, **kwargs):
+    def __init__(self, camera=None, detector=None, log=None, **kwargs):
         self._instrument = None
         self._raftName = None
         self._slotName = None
-        self._detectorName = detectorName
+        self._detectorName = None
         self._detectorSerial = None
-        self._detectorId = detectorId
+        self._detectorId = None
         self._filter = None
         self._calibId = None
         self._metadata = PropertyList()
         self.setMetadata(PropertyList())
+        self.calibInfoFromDict(kwargs)
 
         # Define the required attributes for this calibration.
         self.requiredAttributes = set(['_OBSTYPE', '_SCHEMA', '_VERSION'])
@@ -82,7 +83,7 @@ class IsrCalib(abc.ABC):
 
         if detector:
             self.fromDetector(detector)
-        self.updateMetadata(camera=camera, detector=detector, setDate=False)
+        self.updateMetadata(camera=camera, detector=detector)
 
     def __str__(self):
         return f"{self.__class__.__name__}(obstype={self._OBSTYPE}, detector={self._detectorName}, )"
@@ -145,7 +146,7 @@ class IsrCalib(abc.ABC):
             self.calibInfoFromDict(metadata.toDict())
 
     def updateMetadata(self, camera=None, detector=None, filterName=None,
-                       setCalibId=False, setDate=False,
+                       setCalibId=False, setCalibInfo=False, setDate=False,
                        **kwargs):
         """Update metadata keywords with new values.
 
@@ -166,6 +167,9 @@ class IsrCalib(abc.ABC):
         """
         mdOriginal = self.getMetadata()
         mdSupplemental = dict()
+
+        if setCalibInfo:
+            self.calibInfoFromDict(kwargs)
 
         if camera:
             self._instrument = camera.getName()
@@ -261,7 +265,7 @@ class IsrCalib(abc.ABC):
         self._detectorId = search(dictionary, ['DETECTOR', 'detectorId'])
         self._detectorName = search(dictionary, ['DET_NAME', 'DETECTOR_NAME', 'detectorName'])
         self._detectorSerial = search(dictionary, ['DET_SER', 'DETECTOR_SERIAL', 'detectorSerial'])
-        self._filter = search(dictionary, ['FILTER'])
+        self._filter = search(dictionary, ['FILTER', 'filterName'])
         self._calibId = search(dictionary, ['CALIB_ID'])
 
         if not self._detectorId and self._detectorSerial:
@@ -626,10 +630,8 @@ class IsrProvenance(IsrCalib):
         table = tableList[0]
         metadata = table.meta
         inDict = dict()
+
         inDict['metadata'] = metadata
-        inDict['detectorName'] = metadata.get('DET_NAME', None)
-        inDict['detectorSerial'] = metadata.get('DET_SER', None)
-        inDict['instrument'] = metadata.get('INSTRUME', None)
         inDict['calibType'] = metadata['calibType']
         inDict['dimensions'] = set()
         inDict['dataIdList'] = list()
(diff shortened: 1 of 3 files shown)


python/lsst/ip/isr/calibType.py
```diff
@@ -205,6 +218,63 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
...
+                else:
+                    return None
+            elif len(test) == 1:
+                value = list(test)[0]
+                if value == '':
+                    return None
+                else:
+                    return value
+            else:
+                raise ValueError(f"Too many values found: {len(test)} {test} {needles}")
```
mfisherlevine, line 257: I would likely have had comments here, but we agreed this OOB before hand, so that is fine for now.

python/lsst/ip/isr/calibType.py
```diff
@@ -205,6 +218,63 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
...
+                raise RuntimeError(f"Incorrect calibration supplied.  Expected {self._OBSTYPE}, "
+                                   f"found {metadata['OBSTYPE']}")
+
+        self._instrument = search(dictionary, ['INSTRUME', 'instrument'])
+        self._raftName = search(dictionary, ['RAFTNAME'])
+        self._slotName = search(dictionary, ['SLOTNAME'])
+        self._detectorId = search(dictionary, ['DETECTOR', 'detectorId'])
+        self._detectorName = search(dictionary, ['DET_NAME', 'DETECTOR_NAME', 'detectorName'])
+        self._detectorSerial = search(dictionary, ['DET_SER', 'DETECTOR_SERIAL', 'detectorSerial'])
+        self._filter = search(dictionary, ['FILTER', 'filterName'])
```
mfisherlevine, line 272: Is this the `raw_md` here, _i.e._ the untranslated header value? Or will this have been through `astro_metadata_translator` and obs_package translation at this point?

mfisherlevine: If the latter then that's fine, if the former then this will cause problems.

python/lsst/ip/isr/calibType.py
```diff
@@ -205,6 +218,63 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
...
+        self._calibId = search(dictionary, ['CALIB_ID'])
+
+        if not self._detectorId and self._detectorSerial:
+            self._detectorId = self._detectorSerial
```
mfisherlevine, line 276: If you have no detId but you do have a serial, just make the detId equal to the serial? 🤨 I don't really understand how these can be interchangeable - is there scary downstream logic somewhere/everywhere for dealing with this, or am I misunderstanding?

python/lsst/ip/isr/calibType.py
```diff
@@ -555,10 +630,8 @@ def fromTable(cls, tableList):
         table = tableList[0]
         metadata = table.meta
         inDict = dict()
+
```
mfisherlevine, line 633: Some people care deeply about random whitespace additions, but I am not one of those people 🙂

python/lsst/ip/isr/linearize.py
```diff
@@ -283,15 +284,15 @@ def fromTable(cls, tableList):
...
 
         for record in coeffTable:
             ampName = record['AMPLIFIER_NAME']
 
-            fitParams = record['FIT_PARAMS'] if 'FIT_PARAMS' in record else None
-            fitParamsErr = record['FIT_PARAMS_ERR'] if 'FIT_PARAMS_ERR' in record else None
-            fitChiSq = record['RED_CHI_SQ'] if 'RED_CHI_SQ' in record else None
+            fitParams = record['FIT_PARAMS'] if 'FIT_PARAMS' in record.columns else np.array([0.0])
+            fitParamsErr = record['FIT_PARAMS_ERR'] if 'FIT_PARAMS_ERR' in record.columns else np.array([0.0])
+            fitChiSq = record['RED_CHI_SQ'] if 'RED_CHI_SQ' in record.columns else np.nan
```
mfisherlevine, line 295: I'm sure there's a good reason for changing all these sentinel values here and above, so that's probably fine, but in a vacuum I certainly prefer `None` to zero and nan (but like I say, I'm sure there's a good reason, I've just not dug into it here).

python/lsst/ip/isr/linearize.py
```diff
@@ -335,14 +336,13 @@ def toTable(self):
...
-
+        catalog.meta = self.getMetadata().toDict()
         tableList.append(catalog)
 
         if self.tableData:
             catalog = Table([{'LOOKUP_VALUES': value} for value in self.tableData])
             tableList.append(catalog)
-
-        return(catalog)
+        return(tableList)
```
mfisherlevine, line 345: Was the previous returning of `catalog` just a bug?

python/lsst/ip/isr/linearize.py
```diff
@@ -679,6 +680,52 @@ def __call__(self, image, **kwargs):
...
+    """
+    LinearityType = "Spline"
+
+    def __call__(self, image, **kwargs):
+        """Correct for non-linearity.
+
+        Parameters
+        ----------
+        image : `lsst.afw.image.Image`
+            Image to be corrected
```
mfisherlevine, line 704: This is obviously _waaaay_ beyond this scope of this review, but does this really only ever run on the `Image`? Doesn't the variance plane need to be subjected to (some function of) this correction too? Or is that so n-th order as to be deep in the weeds?

python/lsst/ip/isr/linearize.py
```diff
@@ -679,6 +680,52 @@ def __call__(self, image, **kwargs):
...
+        kwargs : `dict`
+            Dictionary of parameter keywords:
+            ``"coeffs"``
+                Coefficient vector (`list` or `numpy.array`).
+            ``"log"``
+                Logger to handle messages (`lsst.log.Log`).
```
mfisherlevine, line 710: I'm actually not familiar with documenting kwargs - does this mean it's considered a mandatory kwarg? Either way, it's not made use of here.

python/lsst/ip/isr/linearize.py
```diff
@@ -679,6 +680,52 @@ def __call__(self, image, **kwargs):
...
+
+        Returns
+        -------
+        output : `bool`
+            If true, a correction was applied successfully.
```
mfisherlevine, line 715: You say you return a single bool, but actually return `True, 0`.

python/lsst/ip/isr/linearize.py
```diff
@@ -679,6 +680,52 @@ def __call__(self, image, **kwargs):
...
+        splineCoeff = kwargs['coeffs']
+        centers, values = np.split(splineCoeff, 2)
+        interp = afwMath.makeInterpolate(centers.tolist(), values.tolist(),
+                                         afwMath.stringToInterpStyle("AKIMA_SPLINE"))
+
+        ampArr = image.getArray()
+        delta = interp.interpolate(ampArr.flatten())
+        ampArr -= np.array(delta).reshape(ampArr.shape)
+
+        return True, 0
```
mfisherlevine, line 726: The fact that you return something saying whether it was applied successfully implies that you might return something else, so I expected to see, from the docs/form of API, to maybe see a copy, a try, and a finally resetting to the copy on fail, and a return `False`. That seems expensive and probably unnecessary, but just saying what I'd expect to happen given that it says you get a return value if successful here.

python/lsst/ip/isr/calibType.py
czwa: I am open to other more elegant solutions, but I don't know that that can exist until some future point where all of the metadata keywords are standardized completely.

python/lsst/ip/isr/calibType.py
czwa: This is an untranslated header value.  However, it's the untranslated header value of the calibration product, which is derived either from a user specified value (`updateMetadata(filterName='myFavoriteFilter')`), or is set by pulling the value from the input exposure, which should have the translated version.  My goal of including this was to allow validation (something we don't do yet) to confirm that the calibration matches the input data.

python/lsst/ip/isr/calibType.py
czwa: I'm removing this, as I think all of the cases that were inconsistent with the detectorId have been purged from the code.

python/lsst/ip/isr/linearize.py
czwa: `None` does not serialize well with astropy tables (where it becomes `astropy.io.fits.card.Undefined`).  I also wanted to keep datatypes consistent for the downstream code.

python/lsst/ip/isr/linearize.py
czwa: Just a bug.

python/lsst/ip/isr/linearize.py
czwa: That's the way it has been written for a long time.  However, it runs in `ip_isr` after the variance plane has been constructed, which makes me wonder if it should be being run prior to that step.  I think in general this is a small enough term to ignore that issue.  Since the nonlinearity is at the bright end, I think the astronomical effects of this would be making significant sources more significant, and possibly biasing surface brightness fits to galaxy cores.  I'll think about it more, as it's certainly out of scope here.

python/lsst/ip/isr/linearize.py
czwa: The lookup table linearizer has log output, and with the goal of "the linearizers should quack identically", I've taken that to mean that they should accept a log kwarg, even if it's unused.

python/lsst/ip/isr/linearize.py
czwa: I wonder how many reviews have missed that.  Fixed the docstrings to match reality.

python/lsst/ip/isr/linearize.py
czwa: The numeric linearizers (spline/squared/polynomial) have no way to fail.  The lookup table can fail, if all pixel values fall outside the lookup table range.  This might be an issue with splines, although the low-end of the fit is padded to ensure a linear extrapolation in that direction.

PR #160, opened 2020-10-14
DM-27157: constructDark.py fails on LSSTCam with 'Image contains no Pixels'

czwa:
This issue is a result of an overscan bug when `MEDIAN_BY_ROW` is active and an overscan row is completely masked.  In addition, the DM-26281 overscan debug rework is added here, as it was necessary to diagnose the issue.

### DIFF
--- a/python/lsst/ip/isr/overscan.py
+++ b/python/lsst/ip/isr/overscan.py
@@ -346,7 +346,10 @@ class OverscanCorrectionTask(pipeBase.Task):
         fitType = afwMath.stringToStatisticsProperty('MEDIAN')
         for row in integerMI:
             newRow = row.compressed()
-            rowMedian = afwMath.makeStatistics(newRow, fitType, self.statControl).getValue()
+            if len(newRow) > 0:
+                rowMedian = afwMath.makeStatistics(newRow, fitType, self.statControl).getValue()
+            else:
+                rowMedian = np.nan
             collapsed.append(rowMedian)
 
         return np.array(collapsed)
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/overscan.py
```diff
@@ -342,7 +346,10 @@ def collapseArrayMedian(self, maskedArray):
         fitType = afwMath.stringToStatisticsProperty('MEDIAN')
         for row in integerMI:
             newRow = row.compressed()
```
plazas, line 348: What is this `compressed` returning? How could it be empty? (which is what you check for in the next conditional)

python/lsst/ip/isr/overscan.py
```diff
@@ -523,6 +534,13 @@ def debugView(self, image, model):
         axes.plot(indices, plotModel, 'r-')
         plot.xlabel("centered/scaled position along overscan region")
         plot.ylabel("pixel value/fit value")
+        if amp:
```
plazas, line 537: So, it's the {{amp}} only used to get information for the string in the title? (just confirming)

plazas (approved):
I think this looks good; I just asked for some minor clarifications, especially in the part that seems to be doing the fix for this ticket ({{len(newRow)}}).

plazas (approved):

python/lsst/ip/isr/overscan.py
czwa: `compressed` returns all values of the maskedArray that are not masked.  The case where `len(newRow) == 0` is the case where all elements of this row's overscan are masked, and so the `newRow` is empty.

python/lsst/ip/isr/overscan.py
czwa: Yes, this is code I pulled from DM-26281 before marking that invalid.  Debugging overscan quickly becomes impossible without information about which amp you're looking at, which wasn't available before.  Passing the optional `amp` gives information to make the title informative.

PR #163, opened 2020-10-26
DM-27319: Examine behavior of IsrTask wrt doSaturation and doSaturationInterpolation

czwa:
Fix handling of saturation interpolation.

doSaturationInterpolation now fully disables saturation interpolation,
even if doSaturation is set to True.  This also now uses the
saturatedMaskName field, as that could cause inconsistency as well.

Fix a related typo.


mfisherlevine (approved):
Nice.

PR #164, opened 2020-11-02
DM-25424: Convert Defect to IsrCalib

czwa:
Migrate Defects to ip_isr and make an IsrCalib.


timj:
@czwa I've pushed a couple of commits to switch to GitHub Actions because Travis had hung up again. To merge I need to turn off the travis requirement -- I can do that any time but better to do it when you are about to merge.

python/lsst/ip/isr/calibType.py
```diff
@@ -209,7 +215,7 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
         self._metadata["INSTRUME"] = self._instrument if self._instrument else None
         self._metadata["RAFTNAME"] = self._raftName if self._raftName else None
         self._metadata["SLOTNAME"] = self._slotName if self._slotName else None
-        self._metadata["DETECTOR"] = self._detectorId if self._detectorId else None
+        self._metadata["DETECTOR"] = self._detectorId if self._detectorId is not None else None
```
timj, line 218: I don't understand what this is doing. As far I can tell it is saying to set this entry to the value in the object if it it not None but to set it to None if it is None. Isn't that the same as letting it set the value directly? Since the only case where you get None is if the `self._detectorId` was itself None and therefore you get the extra logic without any gain.

The other lines are converting False to None but this one is converting None to None.

python/lsst/ip/isr/calibType.py
```diff
@@ -273,7 +279,7 @@ def search(haystack, needles):
         self._calibId = search(dictionary, ['CALIB_ID'])
 
     @classmethod
-    def readText(cls, filename):
+    def readText(cls, filename, **kwargs):
```
timj, line 282: You need to document the purpose of these kwargs arguments.

python/lsst/ip/isr/defects.py
```diff
@@ -0,0 +1,680 @@
+#
+# LSST Data Management System
+#
+# Copyright 2008-2017  AURA/LSST.
```
timj, line 4: This is wrong.

tests/data/fits_region.fits
```diff
@@ -0,0 +1 @@
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
```
timj, line 1: You should add a .gitattributes file to declare this as a binary. Copy the one from meas_algorithms.

timj (approved):
Looks good. Minor comments since this is mostly copying stuff around.

Since this *is* mostly copying my code from another package, can you please reorganize the commits so that you have one commit that is copying the files from meas_algorithms (defects.py and the test code) and then when you commit it make use of the co-author feature so that I turn up somewhere if people start doing a git blame?   See eg https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

python/lsst/ip/isr/calibType.py
czwa: This was a fix for the case where `self._detectorId = 0`.  `_detectorId` is now expected to be an integer, not a string, and in this case `if self._detectorId` returns false, converting the integer 0 into a None.  In turn, this was breaking round-trip tests.

timj: I still don't understand though. I must be missing something.

For:

```python
x = self._detectorId if self._detectorId is not None else None
```

This can be rewritten as:

```python
if self._detectorId is not None:
    x = self_detectorId
else:
    x = None
```

which is the same as:

```python
if self._detectorId is None:
    x = None
else:
    x = self._detectorId
```

which is the same as:

```python
x = self._detectorId
```
isn't it?

czwa: I agree with all that, leading me to believe the test case failure confused me.  Replaced with the simple version.

timj: I'm sure you're are right that `self._detectorId if self._detectorId else None` was wrong because of the `0` so this did need fixing.

czwa:
> Looks good. Minor comments since this is mostly copying stuff around.
> 
> Since this _is_ mostly copying my code from another package, can you please reorganize the commits so that you have one commit that is copying the files from meas_algorithms (defects.py and the test code) and then when you commit it make use of the co-author feature so that I turn up somewhere if people start doing a git blame? See eg https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

I did not know that was an option, but my rebased commit: https://github.com/lsst/ip_isr/pull/164/commits/b57577753e48a26b6ec6c914d8487e18965f6c95 seems to have the attribution correct now.

tests/data/fits_region.fits
timj: can you reorganize the commit history so that the .gitattributes file turns up before the fits file is committed?

PR #166, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


python/lsst/ip/isr/calibType.py
```diff
@@ -188,6 +188,8 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
                 (self._raftName, self._slotName) = self._detectorName.split("_")
 
         if filterName:
+            # TOD0 DM-28093: I think this whole comment can go away, if we
+            # always use physicalLabel everywhere in ip_isr.
             # If set via:
             # exposure.getInfo().getFilter().getName()
             # then this will hold the abstract filter.
```
kfindeisen, line 195: I'm not sure you can use `physicalLabel` everywhere in `ip_isr` (at least, not at the UI level), but this comment seems out of date regardless.

python/lsst/ip/isr/fringe.py
```diff
@@ -53,8 +53,12 @@ class FringeStatisticsConfig(Config):
 
 class FringeConfig(Config):
     """Fringe subtraction options"""
+    # TODO DM-28093: change the doc to specify that these are physical labels
     filters = ListField(dtype=str, default=[], doc="Only fringe-subtract these filters")
```
kfindeisen, line 57: I believe [this comment on `#dm-middleware-dev`](https://lsstc.slack.com/archives/C2JPT1KB7/p1599762703142500) was specifically talking about this field (and the analogous one for straylight).

kfindeisen (approved):
I'm a bit unsure about what DM-28093 should be doing, but no objections to the changes so far.

python/lsst/ip/isr/calibType.py
parejkoj: I'd rather not muck with it yet: it might be useful guidance for the implementer of DM-28093.

python/lsst/ip/isr/fringe.py
parejkoj: That's exactly what DM-28093 is for. We can make ISR require that those values are specified as physical filter names (and I think they have to be, because ISR deals with the raw data, so is very physical filter-specific), but I don't think I have enough ISR expertise to get it right on this ticket.

Thanks for mentioning the config in straylight.py: I've added a TODO note there too.

kfindeisen: I thought @czwa specifically asked for the UI to support bands? Or did I misunderstand that comment?

parejkoj: There's two questions, one for middleware and one for config. I assert that the config has to be physical, because it is referring directly to the hardware. The middleware doesn't matter here, I don't think, because we can do the necessarily lookups. If it doesn't actually work out that way on DM-28093, I'm ok with these TODO comments being incorrect, the important thing is that ISR becomes internally self-consistent. Right now it depends on what is specified in a given config file and that's obs-package dependent.

czwa: It would be useful if ISR could still use bands (issuing warnings after DM-28093).  If this won't be possible, then DM-28093 will also need to ensure that any filter comparison checks have useful log messages.

PR #167, opened 2021-01-07
DM-23159: Make PTC task compatible with RFC-352 

plazas:


czwa (approved):
I think there are minor formatting issues.  Most of the multiline `'KEY': value if len(value) else [np.nan]` entries have the `if` line shifted left by one character, but some others don't.  I don't know which is correct (both, if flake8 didn't complain?), but they should be consistent.

PR #172, opened 2021-01-28
DM-27833: Enable gen3 inter-chip crosstalk for DECam

czwa:
Improve inter-chip crosstalk application so it works.

### DIFF
--- a/python/lsst/ip/isr/crosstalk.py
+++ b/python/lsst/ip/isr/crosstalk.py
@@ -27,6 +27,7 @@ from astropy.table import Table
 
 import lsst.afw.math
 import lsst.afw.detection
+import lsst.daf.butler
 from lsst.pex.config import Config, Field, ChoiceField, ListField
 from lsst.pipe.base import Task
 
@@ -700,7 +701,8 @@ class CrosstalkTask(Task):
         ------
         RuntimeError
             Raised if called for a detector that does not have a
-            crosstalk correction.
+            crosstalk correction.  Also raised if the crosstalkSource
+            is not an expected type.
         """
         if not crosstalk:
             crosstalk = CrosstalkCalib(log=self.log)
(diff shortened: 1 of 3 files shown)


python/lsst/ip/isr/crosstalk.py
```diff
@@ -717,20 +720,31 @@ def run(self, exposure, crosstalk=None,
 
             if crosstalk.interChip:
                 if crosstalkSources:
+                    # Parse crosstalkSources: Identify which detectors we have available
+                    if isinstance(crosstalkSources[0], lsst.afw.image.Exposure):
+                        # Received afwImage.Exposure
+                        sourceNames = [exp.getDetector().getName() for exp in crosstalkSources]
+                    else:
```
mrawls, line 727: To be a little better at catching weird situations, I suggest making this an `elif` and having an `else` which raises.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -717,20 +720,31 @@ def run(self, exposure, crosstalk=None,
...
                         if detName not in sourceNames:
                             self.log.warn("Crosstalk lists %s, not found in sources: %s",
                                           detName, sourceNames)
                             continue
+                        # Get the coefficients.
                         interChipCoeffs = crosstalk.interChip[detName]
+
                         sourceExposure = crosstalkSources[sourceNames.index(detName)]
+                        if not isinstance(sourceExposure, lsst.afw.image.Exposure):
+                            # Dereference the dafButler.DeferredDatasetHandle.
```
mrawls, line 742: Suggest explicitly checking to see if it's an instance of `dafButler.DeferredDatasetHandle`, not just checking to see if it's *not* an exposure

python/lsst/ip/isr/crosstalk.py
```diff
@@ -717,20 +720,31 @@ def run(self, exposure, crosstalk=None,
...
+                            sourceExposure = sourceExposure.get()
+
+                        self.log.warn("Correcting detector %s with ctSource %s",
```
mrawls, line 745: Is `warn` the best level here? This seems like a good thing to me 😁

mrawls (commented):
Dare I ask if there are any crosstalk tests which need augmenting (or writing in the first place)?

python/lsst/ip/isr/crosstalk.py
czwa: I've done this, and changed the `isinstance(sourceExposure, lsst.afw.image.Exposure)` block to raise as well.

python/lsst/ip/isr/crosstalk.py
czwa: Updated to `info`, because that should be less confusing.

czwa:
> Dare I ask if there are any crosstalk tests which need augmenting (or writing in the first place)?

I've added a test case where we only have inter-chip crosstalk from an exposure passed in.

mrawls (approved):

PR #173, opened 2021-02-22
DM-24389: Implement amp-to-amp offset correction task

leeskelvin:
This commit adds a placeholder amp offset task which is not implemented
here, and instead should be retargeted by a camera specific version.

### DIFF
--- a/python/lsst/ip/isr/ampOffset.py
+++ b/python/lsst/ip/isr/ampOffset.py
@@ -1,0 +1,103 @@
+# This file is part of ip_isr.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# import os
+
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+from lsst.meas.algorithms import (SubtractBackgroundTask, SourceDetectionTask)
+
+
+class AmpOffsetConfig(pexConfig.Config):
+    """Configuration parameters for AmpOffsetTask.
+    """
+    ampEdgeInset = pexConfig.Field(
+        doc="Number of pixels the amp edge strip is inset from the amp edge. A thin strip of pixels running "
+        "parallel to the edge of the amp is used to characterize the average flux level at the amp edge.",
+        dtype=int,
+        default=5,
+    )
+    ampEdgeWidth = pexConfig.Field(
+        doc="Pixel width of the amp edge strip, starting at ampEdgeInset and extending inwards.",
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/ampOffset.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    ampEdgeMinFrac = pexConfig.Field(
+        doc="Minimum allowed fraction of viable pixel rows along an amp edge. No amp-to-amp step estimate "
+        "will be generated for amp edges that do not have at least this fraction of unmasked pixel rows.",
+        dtype=float,
+        default=0.5,
+    )
+    ampEdgeMaxStep = pexConfig.Field(
+        doc="Maximum allowed amp-to-amp step value. If a measured amp-to-amp step value is larger than this, "
+        "the result will be discarded and therefore not used to determine amp offset piston corrections.",
```
erykoff, line 50: We have (starting here in my review) the use of `amp-to-amp step` and `amp offset piston`.  I think we need to (a) standardize on a name, and (b) the more I think about it I don't like `piston` because it confused Josh (there's a physical up-down piston which relates to focus).  I'm not sure what name to use, maybe we should open this up on #dm-naming-things.

erykoff: The doc string should specify the units (ADU).

python/lsst/ip/isr/ampOffset.py
```diff
@@ -0,0 +1,86 @@
...
+
+class AmpOffsetTask(pipeBase.Task):
+    """Calculate and apply amp offset corrections to an exposure.
+    """
+    ConfigClass = AmpOffsetConfig
+    _DefaultName = "isrAmpOffset"
+
+    def run(self, exposure):
+        """Calculate amp-to-amp step values, determine corrective piston
+        pedestals for each amp, and update the input exposure in-place. This
```
erykoff, line 77: And now we have `piston pedestals`.  We want to make sure this is consistent everywhere.

python/lsst/ip/isr/isrTask.py
```diff
@@ -721,6 +721,17 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         default=True,
     )
 
+    # Amp offset correction.
+    doAmpOffset = pexConfig.Field(
```
erykoff, line 725: This and below might be renamed, but I'm actually satisfied with this.  But we should ask more broadly.  I just want to make sure that a search of the codebase will come up with something consistent and relatively unique.  SEO style!

erykoff (commented):
I think we need to discuss naming/terminology to make this as clear and consistent as possible.  Otherwise, looks good.

python/lsst/ip/isr/ampOffset.py
leeskelvin: Thanks for these comments Eli, and apologies it's taken me some time to reply. So, following your comments here and the previous discussion on #dm-naming-things, I have now standardized on two names to use for this ticket: amp offset describes the jump (in ADU) from one amp to another, and amp pedestal describes the specific value (in ADU) added/subtracted from an individual amplifier. All mentions of 'step' and 'piston' have been removed from the code and associated doc strings. I hope that reads well, but keen to hear your thoughts if not.

leeskelvin: PS - mention of units (ADU) added to the docstring here also, thank you.

python/lsst/ip/isr/ampOffset.py
leeskelvin: Following above, this opening sentence now reads: `Calculate amp offset values, determine corrective pedestals for each amp, and update the input exposure in-place.`

python/lsst/ip/isr/isrTask.py
leeskelvin: This name has been preserved.

erykoff (approved):
Looks good.  See my comments on obs_subaru that the subtask initialization should be brought to the parent task here.

PR #174, opened 2021-03-04
DM-28093: Change isrTask (and relevant configs) to use physicalLabel

czwa:
This ticket also provides a `checkFilter` function, to reduce code duplication between fringe, straylight, etc filter checks against config parameters.

### DIFF
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -888,12 +888,16 @@ def checkFilter(exposure, filterList, log=None):
         True if the exposure's filter is contained in the list.
     """
     thisFilter = exposure.getFilterLabel()
-    if thisFilter.physical_filter in filterList:
+    if thisFilter is None:
+        log.warn("No FilterLabel attached to this exposure!")
+        return False
+
+    if thisFilter.physicalLabel in filterList:
         return True
-    elif thisFilter.band in filterList:
+    elif thisFilter.bandLabel in filterList:
         if log:
             log.warn("Physical filter (%s) should be used instead of band %s (%s)",
-                     thisFilter.physical_filter, thisFilter.band, filterList)
+                     thisFilter.physicalLabel, thisFilter.bandLabel, filterList)
         return True
     else:
         return False
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -896,7 +896,7 @@ def checkFilter(exposure, filterList, log):
         return True
     elif thisFilter.bandLabel in filterList:
         if log:
-            log.warn("Physical filter (%s) should be used instead of band %s (%s)",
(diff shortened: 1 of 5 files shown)


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -865,3 +865,39 @@ def setBadRegions(exposure, badStatistic="MEDIAN"):
...
+    for all filter dependent stages.
+
+    Parameters
+    ----------
+    exposure : `lsst.afw.image.Exposure`
+        Exposure to examine.
+    filterList : `list` [`str`]
+        List of physical_filter names to check.
+    log : `lsst.log.Log`, optional
+        Logger to handle messages.
```
parejkoj, line 883: This is not optional below in the first `if` branch. Given that we probably do want those warnings, I think it should be non-optional.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -865,3 +865,39 @@ def setBadRegions(exposure, badStatistic="MEDIAN"):
...
+    thisFilter = exposure.getFilterLabel()
+    if thisFilter is None:
+        log.warn("No FilterLabel attached to this exposure!")
+        return False
+
+    if thisFilter.physicalLabel in filterList:
+        return True
+    elif thisFilter.bandLabel in filterList:
+        if log:
+            log.warn("Physical filter (%s) should be used instead of band %s (%s)",
```
parejkoj, line 899: I feel like some kind of information is missing from this warning. Maybe something about the exposure that would allow the user to track down the problem? Or is this a warning about the `filterList` itself?

python/lsst/ip/isr/isrTask.py
```diff
@@ -1059,8 +1059,7 @@ def readIsrData(self, dataRef, rawExposure):
             dateObs = None
 
         ccd = rawExposure.getDetector()
-        # TODO DM-28093: change this to: rawExposure.getFilterLabel().physicalLabel
-        filterName = afwImage.Filter(rawExposure.getFilter().getId()).getName()  # Canonical name for filter
+        filterName = rawExposure.getFilterLabel()
```
parejkoj, line 1062: Please do a bulk rename of `filterName` to `filterLabel`.

parejkoj (commented):
See comments.

Were these all the uses of `filter` in ISR?

python/lsst/ip/isr/isrFunctions.py
czwa: Made it non-optional, as both current uses (fringe and straylight) have their own loggers to pass in.

python/lsst/ip/isr/isrFunctions.py
czwa: It's a warning about the `filterList`.  This ticket is migrating everything to use `physicalLabel`, but if someone changes the config file (or adds a new camera), I want there to be a warning that they should not use `band`.

PR #175, opened 2021-03-16
DM-24704: Make brighter-fatter correction a subclass of lsst.ip.isr.IsrCalib

czwa:
Add IsrCalib subclass for Brighter-Fatter Kernels, allowing the existing numpy kernels to continue to be used where they exist.


python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -0,0 +1,406 @@
...
+        -------
+        tableList : `list` [`lsst.afw.table.Table`]
+            List of tables containing the crosstalk calibration
+            information.
+
+        """
+        tableList = []
+        self.updateMetadata()
+
+        # Lengths
```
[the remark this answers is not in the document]
czwa: Replaced with a new method so that all three places these numbers are used, they're calculated consistenly.

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -0,0 +1,413 @@
...
+
+    ampKernels are the kernels for each amplifier in a detector, as
+    generated by having level == 'AMP'
+
+    detectorKernel is the kernel generated for a detector as a
+    whole, as generated by having level == 'DETECTOR'
+
+    makeDetectorKernelFromAmpwiseKernels is a method to generate the
+    kernel for a detector, constructed by averaging together the
+    ampwise kernels in the detector.
```
plazas, line 45: `makeDetectorKernelFromAmpwiseKernels`: in terms of usage, do we get this when selecting `AMP`, or when selecting `DETECTOR`, or in both cases? Perhaps it would be good to clarify in the docstring.

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -0,0 +1,413 @@
...
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.BrighterFatterKernel`
+            The initialized calibration.
+        """
+        self._instrument = camera.getName()
+
+        if self.level == 'AMP':
+            detector = camera[detectorId]
```
plazas, line 127: `camera[detectorID]` won't fail if `detectorID` is `None`? (as it is initialized) Is it guaranteed that it will not be `None` when it reaches this point?

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -0,0 +1,413 @@
...
+                self.ampKernels[ampName] = []
+                self.valid[ampName] = []
+        elif self.level == 'DETECTOR':
+            for det in camera:
+                detName = det.getName()
+                self.detKernels[detName] = []
+
+        return self
+
+    def getLengths(self):
```
plazas, line 150: I'm not sure if it's just me, but it's a bit hard to figure out what the purpose of the function is. I guess it's a helper function that gives you the total length of the kernel (from the product of its two dimensions), and also the shorter length for the case when you want to tile the matrix when doing convolutions? Also, why do you pop an observation at the end from the `nObservations` set?

plazas: I see, `pop` is just selecting one of the `nObs` (which we assume have all the same length, right?)

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -0,0 +1,413 @@
...
+                       dictionary['metadata'].get('KERNEL_DY', 0))
+        calib.badAmps = dictionary['badAmps']
+
+        calib.means = {amp: np.array(dictionary['means'][amp]) for amp in dictionary['means']}
+        calib.variances = {amp: np.array(dictionary['variances'][amp]) for amp in dictionary['variances']}
+
+        # Lengths for reshape:
+        _, smallShape, nObs = calib.getLengths()
+
+        calib.rawXcorrs = {amp: np.array(dictionary['rawXcorrs'][amp]).reshape((nObs,
```
plazas, line 211: I'm still a bit confused by the reshaping that is happening here. If, let's say, the original shape is `(8,8)` (default) and `nObs` is, say, `20`, then `rawXcorrs` total length is `20*8*8 = 1280`. However, `smallShape` here (which I think has a more appropriate name inside the function, as it is `smallLength`) would be `int((8-1)*(8-1)/4) = 12`, so the reshape here would be `1280-->2800`? I'm probably missing something, sorry!

plazas: And, why the reshape pf the covariance matrices from PTC? in preparation for the tiling when doing the convolution?

python/lsst/ip/isr/brighterFatterKernel.py
plazas: Upon reading further (`isrTask.py`), I see that there's two options to control this: `brighterFatterKernel.level` and `self.config.brighterFatterLevel`., and that `makeDetectorKernelFromAmpwiseKernels` happens when `self.config.brighterFatterLevel == 'DETECTOR'` and `brighterFatterKernel.level == AMP`. So maybe you can add this info here too?

czwa: I've updated the docstring.  At some point, there will probably be questions about why the application code expects {{level == 'DETECTOR'}}, but the generation code defaults to {{level == 'AMP'}}.

python/lsst/ip/isr/brighterFatterKernel.py
czwa: The `detectorId` is required for the `level='AMP'` case.  I've added a `raise` if this isn't the case.

python/lsst/ip/isr/brighterFatterKernel.py
czwa: I added this in response to the comments above that I was duplicating this code in multiple places.  The issue is that we want the kernel and all components to be 2-d arrays, but those are not easily persistable in FITS.  This calculates the kernel size (17x17 for the standard 8 pixel covariance range), the covariance size (8x8 for the 8 pixel covariances), and the number of observations.  As you mention, the pop is just there to get the single number of observations; I'm using the `set` to ensure that all amplifiers have the same number of means, and raising if there's a difference.

python/lsst/ip/isr/brighterFatterKernel.py
czwa: The `calib.shape` is the shape of the kernel, which has been tiled to fill all quadrants.  As in the previous comment, this is 17x17 for an 8x8 covariance.  So for 20 observations, the returned lengths will be `kernelLength = 17*17 = 289, smallLength = ((17-1)*(17-1))/4 = 8*8 = 64, nObs = 20`.

The reshape is here just because the code stores the covariances for future validation, and they need to be persistable to FITS.

However, in looking at this block again, I made a mistake on the code deduplication ticket.  As you point out, `smallLength` is the better name, as it's just a single value.  However, this is the "to FITS" reshape size, so in constructing the calibration ("from FITS"), it needs to use the `sqrt` of this value as the shapes.

I think this also implies I need to update the unit tests as well, as a read/write test would have found that error.

PR #176, opened 2021-03-18
DM-29280: Look into source of NaN/Infs values in calexp images

laurenam:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1542,11 +1542,11 @@ def run(self, ccdExposure, camera=None, bias=None, linearizer=None,
             # convolution to the science image. At the edges this
             # convolution may not have sufficient valid pixels to
             # produce a valid correction. Mark pixels within the size
-            # of the brighter-fatter kernel as EDGE to warn of this
+            # of the brighter-fatter kernel as SUSPECT to warn of this
             # fact.
             self.log.info("Ensuring image edges are masked as SUSPECT to the brighter-fatter kernel size.")
             self.maskEdges(ccdExposure, numEdgePixels=numpy.max(bfKernel.shape) // 2,
-                           maskPlane="EDGE")
+                           maskPlane="SUSPECT")
```
czwa, line 1549: Is this change to make the action match the log message?  It looks like the suggestion for EDGE was to indicate this is a kernel width rather than a electronically suspect issue, but you were [linked to that discussion](https://jira.lsstcorp.org/browse/DM-5081?focusedCommentId=200225&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-200225) for comments.

czwa (approved):

python/lsst/ip/isr/isrTask.py
laurenam: Yes...I debated whether to change what the doc said to follow the code, but went the other way.  Thanks for reminding me of that ticket...looks like I should leave EDGE and update the doc!

laurenam: It also just occurred to me that `NO_DATA` should also be added to to list being looped over here: https://github.com/lsst/ip_isr/blob/79ae8c2cdcc7bf0247a0ed64343b148f69aa7b36/python/lsst/ip/isr/isrTask.py#L1553
What do you think about creating a `maskToInterpolateInBrighterFatter` config?

mfisherlevine: FWIW (and I know you already converged on this) but I think EDGE is a better semantic match too. Although it's not technically an electro_static_ problem, it's not far off (it's an electrodynamic one, after all). I don't know in great detail _exactly_ which planes get used for what downstream, but I much suspect that EDGE usually maps roughly to "there's enough of an assumed E-field distortion (static or otherwise) here that you shouldn't use these shapes for weak lensing" and given that's probably the only thing these pixels really need excluding for, piggy-backing on EDGE for that purpose seems like a good solution to me.

laurenam: Awesome, thanks for the input Merlin...back to `EDGE` it goes!

czwa: I don't really see a better solution than to add a new config option.  I would prefer not to (there are already too many config options), but the only other solution is to make brighter-fatter a subtask (which it probably should be, but that's way out-of-scope here).
I'm happy to take another peek when you're done.

PR #177, opened 2021-03-30
DM-29271: PTC task: Refactoring/simplification of code ported from Pierre Astier's repository.

plazas:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -215,6 +215,56 @@ def __eq__(self, other):
                     return False
         return True
 
+    def setAmpValues(self, ampName, inputExpIdPair=[(np.nan, np.nan)], expIdMask=[np.nan],
+                     rawExpTime=[np.nan], rawMean=[np.nan], rawVar=[np.nan], photoCharge=[np.nan],
+                     gain=np.nan, gainErr=np.nan, noise=np.nan, noiseErr=np.nan, ptcFitPars=[np.nan],
+                     ptcFitParsError=[np.nan], ptcFitChiSq=np.nan, covArray=[], covArrayModel=[],
+                     covSqrtWeights=[], aMatrix=[], bMatrix=[], covArrayModelNoB=[], aMatrixNoB=[],
+                     finalVar=[np.nan], finalModelVar=[np.nan], finalMean=[np.nan]):
+        """Function to initialize an amp of a PhotonTransferCurveDataset."""
```
czwa, line 224: Closing """ should be on a separate line.  It might be good to add a note pointing that these parameters are all documented in the class __init__.

laurenam: I think this requirement has been relaxed (to accommodate the on-same-line enforcement in `black`): https://developer.lsst.io/python/numpydoc.html#docstrings-should-begin-with-and-terminate-with-on-its-own-line

plazas: Thank you both! I'll add the comment that Chris suggests, but keep in mind what Lauren points out.

czwa: I did not know this, so thanks from me as well!

PR #178, opened 2021-04-02
DM-27103: Not all IsrCalib tests round trip calibrations

czwa:
This fixes the missing tests, and enhances the base class __eq__ to reduce code duplication.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -23,6 +23,8 @@ import datetime
 import os.path
 import warnings
 import yaml
+import numpy as np
+
 from astropy.table import Table
 from astropy.io import fits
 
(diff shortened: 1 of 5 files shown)


python/lsst/ip/isr/crosstalk.py
```diff
@@ -334,11 +339,11 @@ def fromTable(cls, tableList):
...
 
         inDict['coeffs'] = coeffTable['CT_COEFFS']
-        if 'CT_ERRORS' in coeffTable:
+        if 'CT_ERRORS' in coeffTable.columns:
             inDict['coeffErr'] = coeffTable['CT_ERRORS']
-        if 'CT_COUNTS' in coeffTable:
+        if 'CT_COUNTS' in coeffTable.columns:
             inDict['coeffNum'] = coeffTable['CT_COUNTS']
-        if 'CT_VALID' in coeffTable:
+        if 'CT_VALID' in coeffTable.columns:
```
plazas, line 346: Why now do you have to check `coeffTable.columns` instead of just `coeffTable`? Is this the only place where this needs to be changed?

plazas (approved):
Looks good, just two minor questions asking for clarificaton.

python/lsst/ip/isr/calibType.py
```diff
@@ -91,15 +93,42 @@ def __str__(self):
...
+        Running ``calib.log.setLevel(0)`` enables debug statements to
+        identify problematic fields.
         """
         if not isinstance(other, self.__class__):
+            self.log.debug("Incorrect class type: %s %s", self.__class__, other.__class__)
             return False
 
         for attr in self._requiredAttributes:
-            if getattr(self, attr) != getattr(other, attr):
+            attrSelf = getattr(self, attr)
```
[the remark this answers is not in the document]
czwa: Yes.  Since they were doing essentially the same kind of checks, it made sense to collect them in the base class `__eq__` so any future calibs will have a functional equivalence with no extra work.

python/lsst/ip/isr/crosstalk.py
czwa: Without the change, astropy throws a
`astropy/table/row.py:76: FutureWarning: elementwise == comparison failed and returning scalar instead; this will raise an error or perform elementwise comparison in the future.
  return self.as_void() == other` warning that I found annoying.  I don't know why checking inclusion requires an `__eq__` run on the row, and didn't care to dig deeper.  Checking the `.columns` attribute should have the same behavior, and as it isn't calling whatever the other style is calling, it probably runs infinitesimally faster.

PR #179, opened 2021-04-06
Defects instantiation creates undefined/infinite bbox

czwa:
Do not normalize an empty Defects() object.

### DIFF
--- a/python/lsst/ip/isr/defects.py
+++ b/python/lsst/ip/isr/defects.py
@@ -185,6 +185,10 @@ class Defects(IsrCalib):
         if self._bulk_update:
             return
 
+        # If we have no defects, there is nothing to normalize.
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/defects.py
```diff
@@ -185,6 +185,10 @@ def _normalize(self):
         if self._bulk_update:
             return
 
+        # If we have no defects, there is nothing to normalize.
+        if len(self) <= 0:
```
timj, line 189: can `len` return a negative number?

czwa: python is smarter than I thought, and even if you maliciously craft a class, you cannot return a negative length.  Updated to "==".

erykoff (approved):
Looks good!

PR #184, opened 2021-06-01
DM-30351: Make sure brighter-fatter correction is working on DC2 data for DP0.2

czwa:
Improve level = DETECTOR brighter-fatter kernels:

 * Fix incorrect super().__init__() call
 * Fix handling of initFromCamera
 * Switch to lowercase calib name
 * Fix handling for DETECTOR models not having amp-level products
 * Only persist amp-level products if level == AMP


plazas (approved):
Looks good.  My main question was if the changes here were only related to the logic for AMP and DETECTOR levels and how that is related to fixing the BFKs for DC2. Chris confirmed that, in fact, these are only AMP/DETECTOR differences to match what gen3 is expecting (a single kernel per detector).

PR #185, opened 2021-06-06
DM-30630: Filter look-up for in isrTask is broken in a few places

laurenam:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -897,12 +897,46 @@ def checkFilter(exposure, filterList, log):
...
+                     thisPhysicalFilter, thisFilter.bandLabel, filterList)
         return True
     else:
         return False
+
+
+def getPhysicalFilter(filterLabel, log):
+    """Get the physical filter label associated with the given filterLabel.
+
+    If ``filterLabel`` is `None` or there is no physicaLabel attribute
```
czwa, line 915: Typo on physicalLabel.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1380,14 +1382,14 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
             raise RuntimeError("Must supply a flat exposure if config.doFlat=True.")
         if self.config.doDefect and defects is None:
             raise RuntimeError("Must supply defects if config.doDefect=True.")
-        if (self.config.doFringe and filterLabel in self.fringe.config.filters
+        if (self.config.doFringe and physicalFilter in self.fringe.config.filters
```
czwa, line 1385: These three similar changes probably should be calling `isrFunctions.checkFilter`.  I'm not sure how these were missed when that function was added, but this seems like a good time to fix it so all filters are checked identically against config parameters.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2541,18 +2543,21 @@ def roughZeroPoint(self, exposure):
             Exposure to process.
         """
         filterLabel = exposure.getFilterLabel()
-        if filterLabel in self.config.fluxMag0T1:
-            fluxMag0 = self.config.fluxMag0T1[filterLabel]
+        physicalFilter = isrFunctions.getPhysicalFilter(filterLabel, self.log)
+
+        if physicalFilter in self.config.fluxMag0T1:
```
czwa, line 2548: Using `isrFunctions.checkFilter` could be used here as well.

czwa (approved):
I've suggested throughout to use `isrFunctions.checkFilter` to ensure that we handle configs that have bands instead of physical filters.  That change should have been added previously (and hopefully all such config files have been weeded out), but since this ticket is revisiting ISR filter checks, it's probably best to be consistent.

python/lsst/ip/isr/isrTask.py
czwa: I think I retract this comment, because the physicalFilter is needed multiple times (as this isn't just a simple check).

python/lsst/ip/isr/isrTask.py
laurenam: Ok, so I did consider this option having made note of the `checkFilter` function.  However, I also thought I discovered why you didn't use it (and is related to what I alluded to in the commit message re tests).  So, if I try adding 
`hasFilter = isrFunctions.checkFilter(ccdExposure, filterList=["HSC-G", "HSC-R", "HSC-I"], log=self.log)`
just after
`filterLabel = rawExposure.getFilterLabel()`
at line 1107 above, then in `pipe_tasks` tests, we get:
```
lauren@lsst-devl01:~/LSST/pipe_tasks[master] $ python tests/test_processCcd.py
  File "/home/lauren/LSST/ip_isr/python/lsst/ip/isr/isrTask.py", line 1366, in run
    hasFilter = isrFunctions.checkFilter(ccdExposure, filterList=["HSC-G", "HSC-R", "HSC-I"],
  File "/home/lauren/LSST/ip_isr/python/lsst/ip/isr/isrFunctions.py", line 902, in checkFilter
    if thisFilter.physicalFilter in filterList:
AttributeError: 'lsst.afw.image._imageLib.FilterLabel' object has no attribute 'physicalFilter'
```

So...we can't blindly do a `filterLabel.physicalLabel` if we don't want it to raise just because the `physicalLabel` is not present.  I _think_ this should only occur in tests with mocked up `filterLabel` objects...but I'm not sure of that...hence the new function with the `try/except` and returning `Unknown` if that fails with the appropriate error type.

laurenam: Also, I do like knowing what the `physicalFilter` is being evaluated to (if just to help in the log message), so I considered just updating `checkFilter` to output that as well (as a struct), but that would've complicated your cased for when you just want `True/False` and the various overrides of that function would also need updating.  Sooooo....I opted for the perhaps-slight-overkill extra function and also updated `checkFilter` to use it so that it won't bomb on the `pipe_task` tests if they ever hit it.

czwa: I agree with the new function that does the proper checks and returns `Unknown`.  I think this works fine, and if you want to skip updating the fringe/etc checks on this ticket and merge, that works too.

laurenam: Ok, so I think I've confirmed that I can change the other (than the one you retracted below) instances to `checkFilter` calls without test breakage (I would need Jenkins to confirm, though!).  However, given that you've conceded to to the new `getPhyscialLabel` function, I'm inclined to actually leave things as is (if only because it seems a bit more obvious [to me, anyway] and a tad less verbose to obtain the `physicalFilter` and check against `physicalFilter in selfconfig.appropriateFilterList` as opposed to `checkFilter(ccdExposure, self.config.appropriate.FilterList, self.log)`, but I will happily bow to your preference on this one!

PR #187, opened 2021-06-15
DM-30738: Backport gen3 imSim BF-kernels for gen2

czwa:
Fix gen2 handling of detector level BF kernels.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -1147,15 +1147,15 @@ class IsrTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
             except NoResults:
                 try:  # Fall back to the old-style numpy-ndarray style kernel if necessary.
                     brighterFatterKernel = dataRef.get("bfKernel")
-                    self.log.info("Old style brighter-fatter kernel (np.array) loaded")
+                    self.log.info("Old style brighter-fatter kernel (bfKernel) loaded")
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -1147,15 +1147,15 @@ def readIsrData(self, dataRef, rawExposure):
             except NoResults:
                 try:  # Fall back to the old-style numpy-ndarray style kernel if necessary.
                     brighterFatterKernel = dataRef.get("bfKernel")
-                    self.log.info("Old style brighter-fatter kernel (np.array) loaded")
+                    self.log.info("Old style brighter-fatter kernel (bfKernel) loaded")
                 except NoResults:
                     brighterFatterKernel = None
             if brighterFatterKernel is not None and not isinstance(brighterFatterKernel, numpy.ndarray):
```
czwa, line 1153: All of the changes (except for the log message above) is protected by this if check.  This means if we received a numpy array (which is the case for HSC), then this section is skipped.  All of the other changes to support new brighter-fatter kernels has been done in gen3, which uses the `runQuantum` method to check and restructure calibrations.  When we switch fully to gen3, this method will be removed, as it's only called in gen2.

laurenam: Ah, that's what I was missing...thanks for the explanation!

laurenam (approved):

PR #188, opened 2021-07-07
DM-17471: Save intermediate data product with pre-interpolation pixel values

czwa:
Fix typo in returned struct.


arunkannawadi (approved):
Looks good to me. Do I understand it right that before this PR, the code has typos (or multiple variable names) which prevented accessing pre-interpolated exposure?

czwa:
Yes, the connection class had an output connection "preInterExposure", but the code was saving an item "preInterpolatedExposure" into the Struct that was returned.  The discard statements (to drop the connection when the config wouldn't populate them) had the correct name.  I think this was just a case of the gen2 and gen3 code being written on different dates, and the inconsistency slipping in because of that.

PR #191, opened 2021-08-13
DM-31394: Overscan subtraction can create negative variance

czwa:
Add options to mask pixels with negative variance by default.  This should never happen on science images, so this should be a safe change.

Add a test to check this, but let it be disabled in most of the IsrTask tests, as those operate on mock images that are not necessarily constructed the same way a science image is.

### DIFF
--- a/tests/test_isrTask.py
+++ b/tests/test_isrTask.py
@@ -429,6 +429,31 @@ class IsrTaskUnTrimmedTestCases(lsst.utils.tests.TestCase):
         self.assertEqual(countMaskedPixels(results.exposure, "SUSPECT"), 0)
         self.assertEqual(countMaskedPixels(results.exposure, "BAD"), 40800)
 
+    def test_maskingCase_noMasking(self):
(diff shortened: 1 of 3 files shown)


tests/test_isrTask.py
```diff
@@ -404,7 +404,7 @@ def test_failCases(self):
         with self.assertRaises(RuntimeError):
             self.validateIsrResults()
 
-    def test_maskingCase_noMasking(self):
+    def test_maskingCase_negativeVariance(self):
```
cmsaunders, line 407: Why don't you just want to add a new test instead of altering this one?

cmsaunders (approved):
This looks good, except I just had a question about one of the tests.

tests/test_isrTask.py
czwa: That was my original plan, but it looks like I never pasted a copy of the original test before editing.  I've restored the missing test.

cmsaunders:
Ok, with the masking test restored I think this is okay to merge.

PR #194, opened 2021-09-08
DM-31696: Switch to python logging

timj:
Python logging now preferred for all python code so change IsrCalib to create a python logger.


czwa (approved):

czwa:
Now realizing this is part of a larger ticket;  I do not see any extra cases of `lsst.log` remaining, so this looks complete to me.

PR #196, opened 2021-09-15
DM-29531: Update IsrCalib so it can serve as StorageClass for subclasses

czwa:
Allow generic readFits/readText usage for IsrCalib.

IsrCalib has previously relied on subclasses implementing fromDict and
fromTable methods, calling the inherited readText and readFits
methods.  However, adding the calibration class to the metadata by
default (added during calls to updateMetadata) allows the base class
to be used for generic reads, eliminating the need to add
storageClasses/formatters.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -465,7 +465,7 @@ class IsrCalib(abc.ABC):
                 if isinstance(v, fits.card.Undefined):
                     table.meta[k] = None
 
-        calibClass = cls.determineCalibClass(table[0].meta, "readFits")
+        calibClass = cls.determineCalibClass(tableList[0].meta, "readFits")
         return calibClass.fromTable(tableList, **kwargs)
 
     def writeFits(self, filename):
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -60,8 +60,8 @@ class IsrCalib(abc.ABC):
     log : `logging.Logger`, optional
         Log for messages.
     """
-    _OBSTYPE = 'generic'
-    _SCHEMA = 'NO SCHEMA'
+    _OBSTYPE = "generic"
+    _SCHEMA = "NO SCHEMA"
     _VERSION = 0
 
     def __init__(self, camera=None, detector=None, log=None, **kwargs):
@@ -78,10 +78,10 @@ class IsrCalib(abc.ABC):
         self.calibInfoFromDict(kwargs)
 
         # Define the required attributes for this calibration.
-        self.requiredAttributes = set(['_OBSTYPE', '_SCHEMA', '_VERSION'])
-        self.requiredAttributes.update(['_instrument', '_raftName', '_slotName',
-                                        '_detectorName', '_detectorSerial', '_detectorId',
-                                        '_filter', '_calibId', '_metadata'])
+        self.requiredAttributes = set(["_OBSTYPE", "_SCHEMA", "_VERSION"])
+        self.requiredAttributes.update(["_instrument", "_raftName", "_slotName",
+                                        "_detectorName", "_detectorSerial", "_detectorId",
+                                        "_filter", "_calibId", "_metadata"])
 
         self.log = log if log else logging.getLogger(__name__.partition(".")[2])
 
@@ -228,9 +228,9 @@ class IsrCalib(abc.ABC):
 
         if setDate:
             date = datetime.datetime.now()
-            mdSupplemental['CALIBDATE'] = date.isoformat()
-            mdSupplemental['CALIB_CREATION_DATE'] = date.date().isoformat()
-            mdSupplemental['CALIB_CREATION_TIME'] = date.time().isoformat()
+            mdSupplemental["CALIBDATE"] = date.isoformat()
+            mdSupplemental["CALIB_CREATION_DATE"] = date.date().isoformat()
+            mdSupplemental["CALIB_CREATION_TIME"] = date.time().isoformat()
 
         if setCalibId:
             values = []
@@ -240,7 +240,7 @@ class IsrCalib(abc.ABC):
             values.append(f"detector={self._detectorId}") if self._detectorId else None
             values.append(f"filter={self._filter}") if self._filter else None
 
-            calibDate = mdOriginal.get('CALIBDATE', mdSupplemental.get('CALIBDATE', None))
+            calibDate = mdOriginal.get("CALIBDATE", mdSupplemental.get("CALIBDATE", None))
             values.append(f"calibDate={calibDate}") if calibDate else None
 
             self._calibId = " ".join(values)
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/calibType.py
```diff
@@ -333,12 +336,20 @@ def readText(cls, filename, **kwargs):
         """
         if filename.endswith((".ecsv", ".ECSV")):
             data = Table.read(filename, format='ascii.ecsv')
-            return cls.fromTable([data], **kwargs)
+            if 'CALIBCLS' in data.meta:
+                storedClass = doImport(data.meta['CALIBCLS'])
+                return storedClass.fromTable([data], **kwargs)
+            else:
+                return cls.fromTable([data], **kwargs)
```
timj, line 343: Quick note that I don't like duplicating the business logic here (and elsewhere). I'd prefer. the if-clause be determining the correct class. So something like:
```python
calib_cls = cls
if "CALIBCLS" in data.meta:
    calib_class = doImport(data.meta["CALIBCLS"])
return calib_class.fromTable([data], **kwargs)
```

PS Single vs double quote consistency.

eacharles:
I had to make a couple of modification to get this to work with the eotask stuff I'm doing:

1. Look for "SCHEMA" keyword in the metadata to know which version of the tables to expect
2. Look for "SCHEMA" and "CALIBCLS" in the primaryHDU header instead of the first table.  The tables have their own metadata, and for me "SCHEMA" in the table metadata means "version of that table" while in the primary HDU it means "version of the entire object.   . 
3. I also had to add IsrCalib to storageClasses.yaml and formaters.yaml over in dafworld.

Here are my diffs:

```
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -422,6 +422,10 @@ class IsrCalib(abc.ABC):
         calib : `lsst.ip.isr.IsrCalib`
             Calibration contained within the file.
         """
+        hdu = fits.open(filename)
+        storedClassName = hdu[0].header.get('CALIBCLS', None)
+        schemaName =  hdu[0].header.get('SCHEMA', None)
+
         tableList = []
         tableList.append(Table.read(filename, hdu=1))
         extNum = 2  # Fits indices start at 1, we've read one already.
@@ -442,8 +446,11 @@ class IsrCalib(abc.ABC):
                 if isinstance(v, fits.card.Undefined):
                     table.meta[k] = None
 
-        if 'CALIBCLS' in table[0].meta:
-            storedClass = doImport(table[0].meta['CALIBCLS'])
+        storedClassName = table[0].meta.get('CALIBCLS', storedClassName)
+        if storedClassName:
+            storedClass = doImport(storedClassName)
+            if schemaName is not None:
+                kwargs['schema'] = storedClass.schemaDict()[schemaName]()
```

python/lsst/ip/isr/calibType.py
```diff
@@ -332,17 +335,22 @@ def readText(cls, filename, **kwargs):
...
         """
         if filename.endswith((".ecsv", ".ECSV")):
-            data = Table.read(filename, format='ascii.ecsv')
-            return cls.fromTable([data], **kwargs)
-
+            data = Table.read(filename, format="ascii.ecsv")
+            calibClass = cls
+            if "CALIBCLS" in data.meta:
+                calibClass = doImport(data.meta["CALIBCLS"])
+            return calibClass.fromTable([data], **kwargs)
```
timj, line 342: It might be worth adding a:
```python
if calibClass is IsrCalib:
    raise ValueError(f"Calibration file {filename} does not know its calibration class and so can not be read using IsrCalib")
```
because otherwise you can going to get a random error from `fromTable` that will likely not explain the actual problem.

That might suggest we need a helper method

```python
@classmethod
def determine_class(cls, metadata: Dict[str, Any], msg: str = "") -> Type:
    calibClass_name = metadata.get("CALIBCLS")
    calibClass = doImport(calibClass_name) if calibClass_name is not None else cls
    if calibClass is IsrCalib:
        raise ValueError(f"Can not use base class to read old data: {msg}")
    return calibClass
```
(I'm exploring possibilities with those first two lines but what you have for those is also fine). The point though is that once we have that check for base class and old file format in place there becomes enough code duplication that it should be factored out.

timj (approved):
This looks good. If an old file is read it will presumably crash because fromTable isn't implemented. Nothing we can do about that but an explicit error message as I describe might be better.

timj:
@eacharles hdu=0 has extra information in it that astropy Table didn't read even though we wrote it with astropy Table?

PS there is a dangling file descriptor in your code example. Use `with fits.open`.

eacharles:
In the derived class that I pointed Chris at we are writing the schema and the class to the PrimaryHDU header, which I think is a better solution, since those to things aren't tied to a single table but rather to the file as a whole. 

https://github.com/lsst-camera-dh/eotask-gen3/blob/f2751a041970717f03de48aafcdbcad14424f57f/python/lsst/eotask_gen3/eoCalib.py#L316

czwa:
The majority of the calibration classes only have one table, which is why we've avoided putting things into the PHU.  I think we currently only use multiple tables with the `CrosstalkCalib` and `BrighterFatterKernel`, both of which have optional secondary tables.

timj:
Does `IsrCalib.readFits` work with `CrossTalkCalib` then? (and how do we handle multiple tables for ECSV files?)

czwa:
> Does `IsrCalib.readFits` work with `CrossTalkCalib` then? (and how do we handle multiple tables for ECSV files?)

Yes, after reading the first table, it attempts to read any further tables until astropy throws an Exception.  We don't handle these for ECSV, but that's a case that we've accepted (yaml provides an easier to read version of that data, as the extra tables store square matrices for both `CrosstalkCalib` and `BrighterFatterKernel`).

eacharles:
Anyway, the code I posted above checks the primary hdu, then lets the first HDU override the class name, I think that something like that would work generically.

timj:
I'm fine with trying to read hdu=0 on the off chance it has something useful in it.

eacharles:
My other comment was about checking if the "SCHEMA" keyword is present, and if so using that to get a particular schema to read the data correctly.    I think this, or something like this, will be useful going forward.

czwa:
I think after fighting this this afternoon, that I'm not adding reading hdu=0.  Adding that read and appending to the existing metadata means that a calibration just constructed is no longer equivalent to one written to disk as FITS.  Furthermore, that FITS version is different from a YAML or ECSV round-tripped version.  Filtering the keywords to ensure equivalence essentially requires all keys to appear either in hdu=0 and/or in the first table.  If this is true, then keeping the read in the first table ensures equivalence in all file formats, and for all write-read-write-read-etc paths.

timj:
Are the files created by @eacharles not created withing this framework but come from some external software?

I was hoping that a simple `astro_metadata_translator.merge_headers()` call would deal with the problem of some headers existing in hdu=0 with the metadata read from hdu=1 taking priority over that in hdu=0 (which is how butler handles header merging for ingest).

timj:
I realize that this is all caused by outside code creating calibrations but the question is whether a small tweak can make that possible rather than requiring that all the external calibrations are rewritten.

eacharles:
We haven’t written many calibrations yet with the new code, we are still just testing things, so there is some flexibility there.   

I guess my perspective is that I don’t see much point in storing tables in escv or yaml files, put if you guys need to do that for some reason and if that creates problems with using HDU 0, I can figure out a way to stuff things into HDU 1. 

-e

timj:
The background is that this system was developed to allow us to store plain text versions of curated calibrations in the obs_lsst_data package which would make them more open to inspections than you would get with a FITS file. These are generally for calibrations that are not per-exposure but have a much longer lifetime (defect masks were the first). This made the process of writeText -> _data repo -> readText -> write-curated-calibrations -> writeFits work.

PR #197, opened 2021-10-01
DM-30648: Add photodiode calibration class

czwa:
Add photodiode calibration class.

### DIFF
--- a/python/lsst/ip/isr/photodiode.py
+++ b/python/lsst/ip/isr/photodiode.py
@@ -49,7 +49,7 @@ class PhotodiodeCalib(IsrCalib):
     _VERSION = 1.0
 
     def __init__(self, timeSamples=None, currentSamples=None, **kwargs):
-        if timeSamples and currentSamples:
+        if timeSamples is not None and currentSamples is not None:
             if len(timeSamples) != len(currentSamples):
                 raise RuntimeError(f"Inconsitent vector lengths: {len(timeSamples)} {len(currentSamples)}")
             else:
@@ -59,9 +59,18 @@ class PhotodiodeCalib(IsrCalib):
             self.timeSamples = np.array([])
             self.currentSamples = np.array([])
 
-        self.integrationMethod = "DIRECT_SUM"
-
         super().__init__(**kwargs)
+
+        if 'integrationMethod' in kwargs:
+            self.integrationMethod = kwargs.pop('integrationMethod')
+        else:
+            self.integrationMethod = 'DIRECT_SUM'
+
+        if 'day_obs' in kwargs:
+            self.updateMetadata(day_obs=kwargs['day_obs'])
+        if 'seq_num' in kwargs:
+            self.updateMetadata(seq_num=kwargs['seq_num'])
+
         self.requiredAttributes.update(['timeSamples', 'currentSamples', 'integrationMethod'])
 
     @classmethod
@@ -177,6 +186,20 @@ class PhotodiodeCalib(IsrCalib):
 
         return([catalog])
 
+    @classmethod
+    def readSummitPhotodiode(cls, filename):
+        """
+        """
+        rawData = np.loadtxt(filename, dtype=[('time', 'float'), ('current', 'float')])
+
+        import os.path
+        basename = os.path.basename(filename)
+        cleaned = os.path.splitext(basename)[0]
+        _, _, day_obs, seq_num = cleaned.split("_")
+
+        return cls(timeSamples=rawData['time'], currentSamples=rawData['current'],
+                   day_obs=day_obs, seq_num=seq_num)
+
     def integrate(self):
         """Integrate the current.
 
@@ -188,6 +211,8 @@ class PhotodiodeCalib(IsrCalib):
 
         if self.integrationMethod == 'DIRECT_SUM':
             return self.integrateDirectSum()
+        elif self.integrationMethod == 'TRIMMED_SUM':
+            return self.integrateTrimmedSum()
         else:
             raise RuntimeError(f"Unknown integration method {self.integrationMethod}")
 
@@ -202,3 +227,23 @@ class PhotodiodeCalib(IsrCalib):
             Total charge measured.
         """
         return np.trapz(self.currentSamples, x=self.timeSamples)
+
+    def integrateTrimmedSum(self):
+        """Integrate points consistent with the median value.
+
+        This uses numpy's trapezoidal integrator.
+
+        Returns
+        -------
+        sum : `float`
+            Total charge measured.
+        """
+        good = np.where(self.currentSamples > 0.0)
+        (q25, q50, q75) = np.nanpercentile(np.log10(self.currentSamples[good]), [25, 50, 75])
+
+        good = np.where(np.abs(np.log10(self.currentSamples[good]) - q50) < 3.0 * 0.74 * (q75 - q25))
+
+        current = self.currentSamples[good]
+        time = self.timeSamples[good]
+
+        return np.trapz(current, time)
--- a/python/lsst/ip/isr/photodiode.py
+++ b/python/lsst/ip/isr/photodiode.py
@@ -208,9 +208,10 @@ class PhotodiodeCalib(IsrCalib):
         calib : `lsst.ip.isr.PhotodiodeCalib`
             The calibration defined in the file.
         """
+        import os.path
+
         rawData = np.loadtxt(filename, dtype=[('time', 'float'), ('current', 'float')])
 
-        import os.path
         basename = os.path.basename(filename)
         cleaned = os.path.splitext(basename)[0]
         _, _, day_obs, seq_num = cleaned.split("_")
--- a/python/lsst/ip/isr/photodiode.py
+++ b/python/lsst/ip/isr/photodiode.py
@@ -1,0 +1,204 @@
+# This file is part of ip_isr.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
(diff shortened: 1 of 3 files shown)


python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+    currentSamples : `list` or `numpy.ndarray`
+        List of current measurements at each time sample.
+    log : `lsst.log.Log`, optional
+        Log to write messages to.
+    **kwargs :
+        Paramters to pass to parent constructor.
+    """
+
+    _OBSTYPE = 'PHOTODIODE'
+    _SCHEMA = 'Gen3 Photodiode'
```
mfisherlevine, line 50: I thought we were trying not to say Gen3, but you know a lot more about this than I do, so ignore as needed.

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+    _VERSION = 1.0
+
+    def __init__(self, timeSamples=None, currentSamples=None, **kwargs):
+        if timeSamples is not None and currentSamples is not None:
+            if len(timeSamples) != len(currentSamples):
+                raise RuntimeError(f"Inconsitent vector lengths: {len(timeSamples)} {len(currentSamples)}")
```
mfisherlevine, line 56: Maybe as a "vs" between the ints? Not really important though.

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+            else:
+                self.timeSamples = np.array(timeSamples)
+                self.currentSamples = np.array(currentSamples)
+        else:
+            self.timeSamples = np.array([])
+            self.currentSamples = np.array([])
+
+        super().__init__(**kwargs)
+
+        if 'integrationMethod' in kwargs:
```
mfisherlevine, line 66: Should this, and its acceptable values, get a mention in the class docs? Normally there would be a pexConfig for such things, so you wouldn't need to put it there as they're ~self-documenting, but given that there's not, how would people know they could choose?

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+            Dictionary of properties.
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.PhotodiodeCalib`
+            Constructed photodiode data.
+
+        Raises
+        ------
+            Raised if the supplied dictionary is for a different
```
mfisherlevine, line 94: Should there be an exception type here?

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if calib._OBSTYPE != dictionary['metadata']['OBSTYPE']:
+            raise RuntimeError(f"Incorrect photodiode supplied.  Expected {calib._OBSTYPE}, "
+                               f"found {dictionary['metadata']['OBSTYPE']}")
+
+        calib.setMetadata(dictionary['metadata'])
+
+        calib.timeSamples = np.array(dictionary.get('timeSamples'))
+        calib.currentSamples = np.array(dictionary.get('currentSamples'))
+        calib.integrationMethod = dictionary.get('integrationMethod')
```
mfisherlevine, line 107: Do you really want to use `.get()` here rather than just a direct `['timeSamples']`? I would have thought that they're mandatory, and so the KeyError there would be a feature? Or do you want the `None`s back for some reason?

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+        Returns
+        -------
+        dictionary : `dict`
+            Dictionary of properties.
+        """
+        self.updateMetadata()
+
+        outDict = {}
+        metadata = self.getMetadata()
+        outDict['metadata'] = metadata
```
mfisherlevine, line 127: Any reason not to do this on one line with `outDict['metadata'] = self.getMetadata()`? (I don't know why I care, probably just to make them look the same as the others, and copy the three lines in the `fromDict()` method or something)

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+
+        outDict['timeSamples'] = self.timeSamples.tolist()
+        outDict['currentSamples'] = self.currentSamples.tolist()
+
+        outDict['integrationMethod'] = self.integrationMethod
```
mfisherlevine, line 132: Now this seems like an instance for a `.get()`, especially as, if I'm reading it right, the `fromDict()` method isn't dealing with the `integrationMethod`, or is it?

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+        -------
+        calib : `lsst.ip.isr.PhotodiodeCalib`
+            The calibration defined in the tables.
+        """
+        dataTable = tableList[0]
+
+        metadata = dataTable.meta
+        inDict = {}
+        inDict['metadata'] = metadata
+        inDict['integrationMethod'] = metadata.pop('INTEGRATION_METHOD', None)
```
mfisherlevine, line 160: Right, just like that 🙂

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+
+        inDict['timeSamples'] = dataTable['TIME']
+        inDict['currentSamples'] = dataTable['CURRENT']
+
+        return cls().fromDict(inDict)
```
mfisherlevine, line 165: Ah, it looks like that's exactly what it is. Arguably I should read everything through, take notes, and then start writing review comments, but that's like 10x the work, so instead I'm just sorry I've not done that 😔

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+        self.updateMetadata()
+        catalog = Table([{'TIME': self.timeSamples,
+                          'CURRENT': self.currentSamples}])
+        inMeta = self.getMetadata().toDict()
+        outMeta = {k: v for k, v in inMeta.items() if v is not None}
+        outMeta.update({k: "" for k, v in inMeta.items() if v is None})
+        outMeta['INTEGRATION_METHOD'] = self.integrationMethod
+        catalog.meta = outMeta
+
+        return([catalog])
```
mfisherlevine, line 189: I imagine there's a much bigger picture reason why this returns something of this shape... Is this so that it's in keeping with other `calib.toTable()` methods?

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @classmethod
+    def readSummitPhotodiode(cls, filename):
```
mfisherlevine, line 192: Not a big fan of the name unless this really is only for files made on the summit. Isn't this actually mostly/only for BOT lab data though? If so, I think either a generic name, or a specific but correct name, like `readBotPhotodiodeData()`. I bet there's more than one format in the wild though, so maybe factor that into the naming too if you can.

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+            File to read samples from.
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.PhotodiodeCalib`
+            The calibration defined in the file.
+        """
+        rawData = np.loadtxt(filename, dtype=[('time', 'float'), ('current', 'float')])
+
+        import os.path
```
mfisherlevine, line 207: Is there a reason this is imported down here? Does this pass linting?!

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+        basename = os.path.basename(filename)
+        cleaned = os.path.splitext(basename)[0]
+        _, _, day_obs, seq_num = cleaned.split("_")
```
mfisherlevine, line 210: Ahhhh, not having some inscrutable regex here is such a breath of fresh air 😊🙏

python/lsst/ip/isr/photodiode.py
```diff
@@ -0,0 +1,260 @@
...
+        ----------
+        filename : `str`
+            File to read samples from.
+
+        Returns
+        -------
+        calib : `lsst.ip.isr.PhotodiodeCalib`
+            The calibration defined in the file.
+        """
+        rawData = np.loadtxt(filename, dtype=[('time', 'float'), ('current', 'float')])
```
mfisherlevine, line 205: I didn't know about that way of using `loadtxt` - cool!

python/lsst/ip/isr/photodiode.py
mfisherlevine: Very good rename, thanks.

python/lsst/ip/isr/photodiode.py
mfisherlevine: Hmm, OK, I see. I didn't think imports not at the top were legal, but if it passes, it passes!

timj: I'm also confused. I could have sworn that imports have to be at the start of a code block but this one comes after that variable is set.

tests/test_photodiode.py
```diff
@@ -0,0 +1,109 @@
...
+        detBuilder.setPixelSize(pixelSize)
+
+        self.camera = camBuilder
+        self.detector = detBuilder
+
+    def testDataOnly(self):
+        calib = PhotodiodeCalib(timeSamples=self.timeSeries,
+                                currentSamples=self.currentSeries)
+
+        self.assertAlmostEqual(calib.integrate(), 2.88414e-10, 5)
```
czwa, line 63: This should use `self.assertFloatsAlmostEqual`.

PR #203, opened 2021-11-11
DM-29326: Fix logic of the doVignette option in isrTask

czwa:
Logic changed, and masking function updated so it can be used generically.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -32,10 +32,8 @@ import lsst.pipe.base.connectionTypes as cT
 from contextlib import contextmanager
 from lsstDebug import getDebugFrame
 
-from lsst.afw.cameraGeom import (PIXELS, FOCAL_PLANE, NullLinearityType,
-                                 ReadoutCorner)
+from lsst.afw.cameraGeom import NullLinearityType, ReadoutCorner
 from lsst.afw.display import getDisplay
-from lsst.afw.geom import Polygon
 from lsst.daf.persistence import ButlerDataRef
 from lsst.daf.persistence.butler import NoResults
 from lsst.meas.algorithms.detection import SourceDetectionTask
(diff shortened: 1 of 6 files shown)


python/lsst/ip/isr/vignette.py
```diff
@@ -52,39 +53,136 @@ class VignetteConfig(Config):
...
     ConfigClass = VignetteConfig
     _DefaultName = "isrVignette"
 
-    def run(self, exposure):
+    def run(self, exposure=None, doUpdateMask=True, maskPlane="NO_DATA", vignetteValue=None, log=None):
         """Generate circular vignette pattern.
 
         Parameters
         ----------
         exposure : `lsst.afw.image.Exposure`
```
plazas, line 70: This `exposure` parameter seems now optional, so it should be indicated as such in the doc string.  In addition, the new parameters should also be documented in the docstring.

python/lsst/ip/isr/vignette.py
```diff
@@ -52,39 +53,136 @@ class VignetteConfig(Config):
...
+        if doUpdateMask:
+            polygon = exposure.getInfo().getValidPolygon()
+            maskVignettedRegion(exposure, polygon, maskPlane="NO_DATA", vignetteValue=vignetteValue, log=log)
+        return fpPolygon
+
+
+def setValidPolygonCcdIntersect(ccdExposure, fpPolygon, log=None):
+    """Set valid polygon on ccdExposure associated with focal plane polygon.
+
+    Where the ccd exposure's valid polygon is considered the intersection of
```
plazas, line 99: Maybe rephrase a bit with something like: "The CCD exposure's valid polygon is defined as the intersection of ... "

python/lsst/ip/isr/isrTask.py
```diff
@@ -859,9 +857,21 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
     # Vignette correction configuration.
     doVignette = pexConfig.Field(
         dtype=bool,
-        doc="Apply vignetting parameters?",
+        doc="Compute and attach validPolygon to exposure according to vignetting parameters?",
```
plazas, line 860: It's not that clear here if a "validPolygon" is some sort of formal type, or if it is defined as something else. I know we should try to be concise here, but if possible, perhaps you could add a few more phrases to explain what "validPolygon" is for someone who looks at this doc.

python/lsst/ip/isr/isrTask.py
```diff
@@ -859,9 +857,21 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
...
     )
+    doMaskVignettePolygon = pexConfig.Field(
+        dtype=bool,
+        doc=("Add a mask bit for pixels within the vingetted region?  Ignored if doVignette "
+             "is False"),
+        default=True,
+    )
+    vignetteValue = pexConfig.Field(
+        dtype=float,
+        doc="Value to replance image array pixels with in the vingetted region?  Ignored if None.",
```
plazas, line 871: Typo: "vingetted"

plazas (approved):

python/lsst/ip/isr/isrTask.py
laurenam: Totally inherited from me...sorry!  And I'm also questioning my decision to add a question mark here!

laurenam:
Hey Chris, just in case you haven't noted this already, you may want to remove the WIP: from the summary of any commit messages that are indeed getting merged (i.e. no longer WIP!)

PR #204, opened 2021-11-23
DM-32682: Use standard python syntax for setting metadata

timj:
`metadata.set(a, b)` is the same as `metadata[a] = b` and the latter can work for more general types of metadata objects.


ktlim (approved):
We don't seem to have consistency about naming of metadata items...

timj:
Regarding naming -- no we don't. It's all over the place.

PR #205, opened 2021-12-07
DM-31592: Fix overlong docstrings

timj:


czwa (approved):
I believe the typo in the raise in `calibType.py` hasn't been discovered before because we haven't added a test for that case, and it hasn't come up organically because we are still generally using the specific class types.

timj:
The calibType error turned up for me with flake8 F821 error code but I guess our GitHub Action doesn't activate those.

PR #208, opened 2022-04-12
DM-34407: cp_pipe flat construction fails due to missing vignette polygon

czwa:
Return if no polygon is supplied.


python/lsst/ip/isr/vignette.py
```diff
@@ -177,6 +177,7 @@ def maskVignettedRegion(exposure, polygon, maskPlane="NO_DATA", vignetteValue=No
     log = log if log else logging.getLogger(__name__)
     if not polygon:
         log.info("No polygon provided.  Masking nothing.")
+        return
```
erykoff, line 180: Can't comment on the line above, but can you also fix the docstring on line 161 which has an extra `.`?

erykoff (approved):
Looks good, minor comment on the docstring which may as well be fixed now.

PR #209, opened 2022-04-19
DM-34481: Fix cp_pipe fringe PipelineDataCycleError

czwa:
Discard the correct input for fringes.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -271,54 +271,54 @@ class IsrTaskConnections(pipeBase.PipelineTaskConnections,
         super().__init__(config=config)
 
         if config.doBias is not True:
-            self.prerequisiteInputs.discard("bias")
+            self.prerequisiteInputs.remove("bias")
         if config.doLinearize is not True:
-            self.prerequisiteInputs.discard("linearizer")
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -287,7 +287,7 @@ def __init__(self, *, config=None):
         if config.doFlat is not True:
             self.prerequisiteInputs.discard("flat")
         if config.doFringe is not True:
-            self.prerequisiteInputs.discard("fringe")
+            self.prerequisiteInputs.discard("fringes")
```
PaulPrice, line 290: `self.prerequisiteInputs.remove` might be more suitable: it detects typos.

PaulPrice (approved):

python/lsst/ip/isr/isrTask.py
erykoff: That's what I said on slack!  All of these here should be `remove` instead of `discard`.

PaulPrice (approved):
💯

python/lsst/ip/isr/isrTask.py
czwa: I've switched them all, so this shouldn't come up again.

PR #215, opened 2022-06-03
DM-31938: Have PTC dataset unpad results

plazas:


czwa (approved):

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -64,8 +64,9 @@ def __init__(self, camera=None, level=None, **kwargs):
         self.level = level
 
         # Things inherited from the PTC
-        self.means = dict()
```
czwa, line 67: Can you file a ticket, and leave a TODO in the class docstring to document what is stored in the BFK calibration?  I was wondering why this change didn't extend to that, but then I realized that I never filled that out when I did this ticket.

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -221,8 +223,10 @@ def fromDict(cls, dictionary):
         calib.shape = (dictionary['metadata'].get('KERNEL_DX', 0),
                        dictionary['metadata'].get('KERNEL_DY', 0))
 
-        calib.means = {amp: np.array(dictionary['means'][amp]) for amp in dictionary['means']}
-        calib.variances = {amp: np.array(dictionary['variances'][amp]) for amp in dictionary['variances']}
+        calib.expIdMask = {amp: np.array(dictionary['expIdMask'][amp]) for amp in dictionary['rawXcorrs']}
```
czwa, line 226: Can these iterate over their own amplifier sets (`for amp in dictionary['expIdMask']}` etc)?  There's nothing wrong as it is, but if these have different lengths, I'd like it to give a clear message about what entry doesn't have a full set of amps.

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -64,8 +64,9 @@ def __init__(self, camera=None, level=None, **kwargs):
         self.level = level
```
czwa, line 64: I can't add this comment on line 61 (as you didn't change that), but please increment the `_VERSION` variable to `1.1` at least.  It would be better to provide backwards compatibility for the new and old versions, but at least catching the version change and having it bomb out with a clear message would be good.  Otherwise, I think old on-disk BFK calibrations will raise.

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -221,8 +223,10 @@ def fromDict(cls, dictionary):
         calib.shape = (dictionary['metadata'].get('KERNEL_DX', 0),
```
czwa, line 223: The failure for the `_VERSION` increment can probably put in this method, as part of the `_OBSTYPE` check.  I think this must either have a case where `_VERSION == 1.0` is packed into the v1.1 format (with a warning), or a `RuntimeError` that the old version is unsupported.

tests/test_ptcDataset.py
```diff
@@ -98,11 +98,11 @@ def test_ptcDatset(self):
                     localDataset.ptcTurnoff[ampName] = localDataset.rawMeans[ampName][-1]
 
                     localDataset.covariances[ampName] = np.full(
-                        (nSignalPoints, nSideCovMatrix, nSideCovMatrix), np.nan).tolist()
```
czwa, line 101: I see these match the values in the FULLCOVARIANCE tests.  Are these just the new correct values for the test?  I think I just don't see why these are switching from NaN to a value (probably just me not spending enough time thinking through the test).

czwa (commented):
We're starting to get into things that need versioning.  I think this is a sufficient change that we need to either support the old version (more work for ~10s of files we can recreate) or deprecate that old version with a message (probably the more reasonable effort here).

python/lsst/ip/isr/brighterFatterKernel.py
plazas: DM-35260

python/lsst/ip/isr/brighterFatterKernel.py
plazas: I think I'll produce a `RuntimeError`, as I don't see how I could turn, for example, `finalMeans` into `rawMeans`, and also create an `expIdMask` entry that might not correspond with reality (i.e., there's missing info. If we were going the other way, then it could be done... `rawMeans[expIdMask] = finalMeans`)

tests/test_ptcDataset.py
plazas: Before, the code only calculated/measured the covariances when fitting the `FULLCOVARIANCE` model. Then, when we decided to split the PTC pipeline into `extract` and `solve`, we decided that we would measure the covariances during `extract`, regardless of the model being fit in `solve`. So now these two models ( `EXPAPROXIMATION` and `POLYNOMIAL`  ) are expected to have a measured covariances attribute.

python/lsst/ip/isr/brighterFatterKernel.py
plazas: I decided to make it backward compatible.

czwa (approved):
I think this all works.  ci_cpp_gen3 should find any length errors, so I'm happy as long as that passes.

PR #217, opened 2022-06-21
DM-35278: Increase version of ptcDataset and make it backwards compatible

plazas:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -440,6 +440,9 @@ def fromTable(cls, tableList):
         inDict['badAmps'] = []
         inDict['photoCharge'] = dict()
 
+        calibVersion = metadata['PTC_VERSION']
```
czwa, line 443: Do you need a check on `calibVersion` for this warning message?  It seems like that message should only trigger if `calibVersion == 1.0`.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -471,6 +473,11 @@ def fromTable(cls, tableList):
             inDict['finalMeans'][ampName] = record['FINAL_MEANS']
             inDict['badAmps'] = record['BAD_AMPS']
             inDict['photoCharge'][ampName] = record['PHOTO_CHARGE']
+            if calibVersion == 1.0:
```
czwa, line 476: Can you add a docstring to the class that mentions that v1.1 adds the `ptcTurnoff` attribute?  I think it'll be easier to understand changes if it doesn't require looking for code tests on `calibVersion.`

czwa (commented):
A few small issues that I'd like to be a bit clearer.

python/lsst/ip/isr/ptcDataset.py
plazas: Yes, good catch, thanks. I had the message inside the loop, with the check, but when I decided to just have one message outside the loop, I did not put a check.

python/lsst/ip/isr/ptcDataset.py
plazas: Agree. On a more general note, are the docstrings the only places where version changes are tracked? I just want to make sure that we don't have to update another document (or perhaps this is the time to decide if we want o create + maintain such an external document).

czwa (approved):
This solves all my concerns, now.

PR #218, opened 2022-06-28
DM-35349: Fix brighter fatter kernel backwards compatibility

czwa:
Fix brighter-fatter kernel backwards compatibility.

This fixes the translation between the FITS table and the intermediate dictionary representation.

### DIFF
--- a/python/lsst/ip/isr/brighterFatterKernel.py
+++ b/python/lsst/ip/isr/brighterFatterKernel.py
@@ -233,7 +233,7 @@ class BrighterFatterKernel(IsrCalib):
 
         calibVersion = dictionary['metadata']['bfk_VERSION']
         if calibVersion == 1.0:
-            cls().log.warning("Old Version of brightter-fatter kernel found. Current version: "
+            calib.log.warning("Old Version of brighter-fatter kernel found. Current version: "
                               f"{calib._VERSION}. The new attribute 'expIdMask' will be "
                               "populated with 'True' values, and the new attributes 'rawMeans'"
                               "and 'rawVariances' will be populated with the masked 'means'."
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -343,9 +343,32 @@ def fromTable(cls, tableList):
 
         amps = ampTable['AMPLIFIER']
 
-        expIdMaskList = ampTable['EXP_ID_MASK']
-        rawMeanList = ampTable['RAW_MEANS']
-        rawVarianceList = ampTable['RAW_VARIANCES']
+        # Determine version for expected values.  The ``fromDict``
+        # method can unpack either, but the appropriate fields need to
+        # be supplied.
+        calibVersion = metadata.get('bfk_VERSION', 1.0)
```
timj, line 349: line 234 does not do this logic. It seems to assume that the version is going to always be there.

timj (commented):
I'm not sure I understand the duplication of logic between fromDict and fromTable given that fromTable calls fromDict.

python/lsst/ip/isr/brighterFatterKernel.py
czwa: Switched to be a direct get.  If this doesn't exist, than there's a larger issue.

czwa: The logic duplication is something I didn't see an obvious way to avoid.  I've tried to keep the `fromDict` logic as the place that does transformations (`mean` -> `rawMean`, etc), as that is used by both the FITS and yaml readers.  This ensures that `fromTable` only needs to do simple repacking of fields for the two versions.

timj (approved):
This looks okay. I can now see that the `fromDict` is using the outputs of `fromTable`.

As an aside line 236 has a typo in the `brightter` text and also should be using `calib` and not `cls()` I think.

czwa:
> As an aside line 236 has a typo in the `brightter` text and also should be using `calib` and not `cls()` I think.

Agree on both points.

PR #222, opened 2022-07-30
DM-35772: Remove gen2 code from ip_isr.

erykoff:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1138,193 +1128,14 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         outputs = self.run(**inputs)
         butlerQC.put(outputs, outputRefs)
 
-    def readIsrData(self, dataRef, rawExposure):
```
czwa, line 1141: I think this is the method I'm most happy is going away.

czwa (approved):
This looks fine to me.

PR #224, opened 2022-08-03
DM-35821: Fix CTI run errors

czwa:
Fix errors in creating CTI statistics.


plazas (approved):
This looks good to me as long as it fixes the errors reported in Jira. I think I understand better the fix to serialize the data (it needs to be a list, it seems), but I'm not sure why this check is not needed anymore. Also, shouldn't these errors have been captured by some sort of unit testing?

czwa:
> This looks good to me as long as it fixes the errors reported in Jira. I think I understand better the fix to serialize the data (it needs to be a list, it seems), but I'm not sure why this check is not needed anymore. Also, shouldn't these errors have been captured by some sort of unit testing?

The additional check was added during review, and I wasn't sure at the time that it was needed.  Unfortunately, the testing on the previous ticket was focused on the Solve stage, and so these bugs slipped through.

PR #227, opened 2022-09-06
DM-36114: Allow the Sphinx documentation to be built

timj:


czwa (approved):
A definite improvement.

PR #228, opened 2022-09-09
DM-36163: Remove unnecessary ISR log messages

czwa:
Remove noisy log message.


plazas (approved):
Looks good to me. If the performance is already comparable in time to other methods, I agree it's not needed anymore.

PR #229, opened 2022-09-09
DM-35569: ip_isr doEmpiricalReadNoise fails if an amplifier is fully masked

czwa:
Move countMaskedPixels from cp_pipe, and use it to check that the amplifier is fully masked.

### DIFF
--- a/tests/test_isrFunctions.py
+++ b/tests/test_isrFunctions.py
@@ -23,6 +23,7 @@
 import unittest
 import numpy as np
 
+import lsst.geom as geom
 import lsst.afw.image as afwImage
 import lsst.utils.tests
 import lsst.ip.isr as ipIsr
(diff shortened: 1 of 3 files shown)


tests/test_isrFunctions.py
```diff
@@ -322,6 +301,38 @@ def test_attachTransmissionCurve_None(self):
...
+
+        NODATABIT = mi.mask.getPlaneBitMask("NO_DATA")
+        noDataBox = geom.Box2I(geom.Point2I(31, 49), geom.Extent2I(3, 6))
+        mi.mask[noDataBox] |= NODATABIT
+
+        self.assertEqual(ipIsr.countMaskedPixels(mi, 'NO_DATA'), noDataBox.getArea())
+        self.assertEqual(ipIsr.countMaskedPixels(mi, 'BAD'), 0)
+
+        mi.mask[noDataBox] ^= NODATABIT  # XOR to reset what we did
+        self.assertEqual(ipIsr.countMaskedPixels(mi, 'NO_DATA'), 0)
```
plazas, line 334: What exactly is happening in this unit test? I see all comparisons to zero. Are these null tests always? Shouldn't there be tests for when the number of masked pixels is different from zero? Maybe I misunderstood the test.

plazas (approved):
LGTM. A small comment about the unit test.

tests/test_isrFunctions.py
czwa: The test above this line applies a mask bit (NO_DATA) to a region, and then checks that the count of those pixels in that mask is equal to the area of that region.  The check for BAD is checking that it's correctly respecting mask planes and only measuring what it's told.

PR #231, opened 2022-09-19
DM-36260: Deferred charge trap array contains NaNs, butler access fails

czwa:
Fix handling of NaN padding in serial trap coefficients.
This version properly truncates the padded values, and filters incoming NaN values from the trap coefficients on construction.

### DIFF
--- a/python/lsst/ip/isr/deferredCharge.py
+++ b/python/lsst/ip/isr/deferredCharge.py
@@ -81,7 +81,12 @@ class SerialTrap():
             raise ValueError('Unknown trap type: %s', self.trap_type)
 
         if self.trap_type == 'spline':
-            centers, values = np.split(np.array(self.coeffs), 2)
+            centers, values = np.split(np.array(self.coeffs, dtype=np.float), 2)
+            # Ensure all NaN values are stripped out
+            values = values[~np.isnan(centers)]
+            centers = centers[~np.isnan(centers)]
+            centers = centers[~np.isnan(values)]
+            values = values[~np.isnan(values)]
             self.interp = interp.interp1d(centers, values)
 
         self._trap_array = None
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/deferredCharge.py
```diff
@@ -378,7 +383,25 @@ def fromTable(cls, tableList):
...
-            ampTrap['coeffs'] = np.array(coeffs[index])[~np.isnan(coeffs[index])].tolist()
+
+            # Unpad any trailing NaN values: find the continuous array
+            # of NaNs at the end of the coefficients, and remove them.
+            inCoeffs = coeffs[index]
+            breakIndex = 1
+            nanValues = np.where(np.isnan(inCoeffs))[0]
+            if nanValues is not None:
+                coeffLength = len(inCoeffs)
+                while (breakIndex < coeffLength):
```
leeskelvin, line 394: Does this need parentheses (just `while (breakIndex < coeffLength):`)?

python/lsst/ip/isr/deferredCharge.py
```diff
@@ -378,7 +383,25 @@ def fromTable(cls, tableList):
...
+                    if coeffLength - breakIndex in nanValues:
+                        breakIndex += 1
+                    else:
+                        break
+            breakIndex -= 1  # Remove the fixed offset.
+            if breakIndex != 0:
+                outCoeffs = inCoeffs[0:coeffLength - breakIndex]
```
leeskelvin, line 401: You might need a space either side of the `:` here.

czwa: Flake8 didn't like the leading space, so I only added the one.

timj: black will add the space but this package doesn't use black (which requires a tweak to flake8 settings).

PR #232, opened 2022-09-23
DM-36337: Brighter-fatter kernels cannot be converted for disk due to length error

czwa:
Repack correlations if needed to ensure consistent lengths.

### DIFF
--- a/python/lsst/ip/isr/brighterFatterKernel.py
+++ b/python/lsst/ip/isr/brighterFatterKernel.py
@@ -342,7 +342,9 @@ class BrighterFatterKernel(IsrCalib):
                     # Repack data.
                     self.repackCorrelations(amp, correlationShape)
                 else:
-                    raise ValueError("Could not coerce rawXcorrs into appropriate shape.")
+                    raise ValueError("Could not coerce rawXcorrs into appropriate shape "
+                                     "(have %d correlations, but expect to see %d.",
+                                     correlationShape[0], np.sum(self.expIdMask[amp]))
 
         outDict['rawXcorrs'] = {amp: np.array(self.rawXcorrs[amp]).reshape(nObs*smallLength).tolist()
                                 for amp in self.rawXcorrs}
@@ -485,7 +487,9 @@ class BrighterFatterKernel(IsrCalib):
                         # Repack data.
                         self.repackCorrelations(amp, correlationShape)
                     else:
-                        raise ValueError("Could not coerce rawXcorrs into appropriate shape.")
+                        raise ValueError("Could not coerce rawXcorrs into appropriate shape "
+                                         "(have %d correlations, but expect to see %d.",
+                                         correlationShape[0], np.sum(self.expIdMask[amp]))
 
                 rawXcorrs.append(np.array(self.rawXcorrs[amp]).reshape(nObs*smallLength).tolist())
                 gainList.append(self.gain[amp])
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -333,6 +333,17 @@ def toDict(self):
...
+
+        for amp in self.rawXcorrs.keys():
+            # Check to see if we need to repack the data.
+            correlationShape = np.array(self.rawXcorrs[amp]).shape
+            if nObs != correlationShape[0]:
+                if correlationShape[0] == np.sum(self.expIdMask[amp]):
+                    # Repack data.
+                    self.repackCorrelations(amp, correlationShape)
+                else:
+                    raise ValueError("Could not coerce rawXcorrs into appropriate shape.")
```
plazas, line 345: Should the error be more informative?

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -504,6 +524,27 @@ def toTable(self):
 
         return tableList
 
+    def repackCorrelations(self, amp, correlationShape):
```
plazas, line 527: It's nice to have a helper function for repacking the correlations. My general worry is, wouldn't we need to repack also other quantities like the fluxes (means) at the same time? Are we running out of sync somehow between the different arrays? The PTC code has an associated "covariances weights" array, although it seems that it is not used in the BFK code.

plazas (changes requested):
It seems that the changes will address the problem at hand, thanks! I'd like to see a bit more information in the error message, maybe the shapes that could not be repacked or something that helps determine the problem a bit better.  And I'd also like for us to think a bit more about whether we need to do this type of repacking in other arrays as well at the same time, or not.

python/lsst/ip/isr/brighterFatterKernel.py
czwa: I've updated the message to note how many covariances we have (`correlationShape[0]`), and how many we expect from the mask (`np.sum(self.expIdMask[amp])`).  I think this will be a rare/never called error, as I think we'll either have all the entries (and need to mask them) or we'll have the pre-masked set (and will need to repad here).

python/lsst/ip/isr/brighterFatterKernel.py
czwa: For the means and variances, the BF code uses the raw versions.  Previously, the covariances had the same length as those, which is likely why I used the raw values.  I've filed a ticket, DM-36344, to put more thought into this choice; If the final values have the same length as the covariances now, those may be the more natural choice.
What are the weights used for in the PTC?  I'd have to think more about the BF algorithm to determine if we should be using those as well.

plazas: The "covariances weights" in the PTC code are calculated from Pierre's code (using the variance of the variance formula ``2*Var/Npix``), and then used in the fit to the ``FULLCOVARIANCE`` model.

plazas (approved):

python/lsst/ip/isr/brighterFatterKernel.py
```diff
@@ -467,6 +480,15 @@ def toTable(self):
...
                 rawMeanList.append(self.rawMeans[amp])
                 rawVarianceList.append(self.rawVariances[amp])
+
+                correlationShape = np.array(self.rawXcorrs[amp]).shape
+                if nObs != correlationShape[0]:
+                    if correlationShape[0] == np.sum(self.expIdMask[amp]):
+                        # Repack data.
+                        self.repackCorrelations(amp, correlationShape)
+                    else:
+                        raise ValueError("Could not coerce rawXcorrs into appropriate shape.")
```
plazas, line 490: Do you need to expand the error message here too?

plazas (approved):

python/lsst/ip/isr/brighterFatterKernel.py
czwa: Ah, good catch.  Thanks!

PR #234, opened 2022-09-27
DM-36068: Parallel overscan correction seems to cause failures in PTC

aferte:
If too many pixels (default is more than 10% of all pixels) are masked, we don't make the parallel overscan correction

### DIFF
--- a/python/lsst/ip/isr/overscan.py
+++ b/python/lsst/ip/isr/overscan.py
@@ -29,6 +29,7 @@ import lsst.pipe.base as pipeBase
 import lsst.pex.config as pexConfig
 
 from .isr import fitOverscanImage
+from .isrFunctions import makeThresholdMask, countMaskedPixels
 
 
 class OverscanCorrectionTaskConfig(pexConfig.Config):
@@ -78,6 +79,16 @@ class OverscanCorrectionTaskConfig(pexConfig.Config):
         dtype=bool,
         doc="Correct using parallel overscan after serial overscan correction?",
         default=False,
+    )
+    parallelOverscanMaskThreshold = pexConfig.RangeField(
+        dtype=float,
+        doc="Minimum fraction of pixels in parallel overscan region necessary "
+        "for parallel overcan correction.",
+        default=0.1,
+        min=0.0,
+        max=1.0,
+        inclusiveMin=True,
+        inclusiveMax=True,
     )
 
     leadingColumnsToSkip = pexConfig.Field(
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/overscan.py
```diff
@@ -79,6 +80,12 @@ class OverscanCorrectionTaskConfig(pexConfig.Config):
         doc="Correct using parallel overscan after serial overscan correction?",
         default=False,
     )
+    maskPixFracThresh = pexConfig.Field(
```
erykoff, line 83: This should be a `RangeField` which allows you to specify a min/max and whether they are inclusive or not (they are in this case), and then that handles the checks behind the scenes.

python/lsst/ip/isr/overscan.py
```diff
@@ -197,14 +204,26 @@ def run(self, exposure, amp, isTransposed=False):
...
+            maskIm = maskIm.Factory(maskIm, parallelOverscanBBox)
+            makeThresholdMask(maskIm, threshold=self.config.numSigmaClip, growFootprints=0)
+            maskPix = countMaskedPixels(maskIm, self.config.maskPlanes)
+            xSize, ySize = parallelOverscanBBox.getDimensions()
+            if maskPix > xSize*ySize*self.config.maskPixFracThresh:
+                self.log.warning('Fraction of masked pixels for parallel overscan calculation larger'
+                                 ' than %f of total pixels (i.e. %f masked pixels) on amp %s.',
+                                 self.config.maskPixFracThresh, maskPix, amp.getName())
+                self.log.warning('Not doing parallel overscan correction.')
+                parallelResults = pipeBase.Struct()
```
erykoff, line 217: This `parallelResults` isn't used at all, so is unnecessary.  But I'm wondering if `overscanMean` should be a tuple `(overscanMean, 0.0)` or it may be fine to not have the second part of the tuple because the parallel overscan wasn't run at all.

python/lsst/ip/isr/overscan.py
```diff
@@ -197,14 +204,26 @@ def run(self, exposure, amp, isTransposed=False):
...
-            overscanMean = (overscanMean, parallelResults.overscanMean)
-            overscanSigma = (overscanSigma, parallelResults.overscanSigma)
-            residualMean = (residualMean, parallelResults.overscanMeanResidual)
-            residualSigma = (residualSigma, parallelResults.overscanSigmaResidual)
+            maskIm = exposure.getMaskedImage()
+            maskIm = maskIm.Factory(maskIm, parallelOverscanBBox)
+            makeThresholdMask(maskIm, threshold=self.config.numSigmaClip, growFootprints=0)
+            maskPix = countMaskedPixels(maskIm, self.config.maskPlanes)
+            xSize, ySize = parallelOverscanBBox.getDimensions()
+            if maskPix > xSize*ySize*self.config.maskPixFracThresh:
```
erykoff, line 212: Just to be clear, this is a check on the _total_ parallel overscan region, and is separate from any particular check on a given column, which presumably happens in the `correctOverscan` code?  If so, the documentation string on the config field should maybe be "total overscan region" or something.

erykoff (changes requested):
Some minor comments, but I want to discuss how to rebase to clean up the commit history.

python/lsst/ip/isr/overscan.py
erykoff: I don't like this name because it doesn't tell us that this is about parallel overscan.

erykoff (approved):

PR #238, opened 2022-10-21
DM-36653: Ensure overscan task returns all overscan models and images

czwa:
Add parallel overscan results to the task output.


aferte (approved):
It all looks good to me

PR #239, opened 2022-11-08
DM-36597: Add LATISS filter transmission data

czwa:
This ticket adds a new `IntermediateTransmissionCurve` calibration class (and subclasses) that can read a common transmission curve format (with metadata), and write an `lsst.afw.image.TransmissionCurve` to butler for use as a calibration.


python/lsst/ip/isr/transmissionCurve.py
```diff
@@ -0,0 +1,214 @@
...
+        else:
+            wavelengths = self.data['wavelength']
+            throughput = self.data[throughputKey]
+
+        # Convert units:
+        # import pdb; pdb.set_trace()
+        with cds.enable():
+            # These need to be in Angstroms, for consistency.
+            wavelengths = wavelengths.to(u.Angstrom).to_value()
+            # This is ugly.  Fix.
```
aferte, line 152: Does 'fix' mean 'to be fixed' or is the fix below?

czwa: It remains ugly (to me at least), but the subsequent `if` line is the fix.  I have deleted this comment as it was intended as an intermediate note to myself.

PR #240, opened 2022-11-16
DM-37022: overscan.py parallel overscan sigma clip is really a threshold clip

czwa:
Fix parallelOverscan bleed mask threshold.


erykoff (approved):
I must ask ... is there a test for this anywhere?

czwa:
> I must ask ... is there a test for this anywhere?

Yes, but perhaps it's not as well designed as it could be.  `tests/test_overscanCorrection.py` has a `test_badParallelOverscanCorrection`, which sets the parallel overscan to a value of 400 (essentially saturation given that the other pixel values are ~10ADU or less).  That does trigger the skip case for the test exposure, which I think was all that was considered when the test was added.

PR #241, opened 2022-11-29
DM-37129: Store list of calibrations used in output header

czwa:
Add calibration provenance header info.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -377,6 +377,11 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         dtype=isrQa.IsrQaConfig,
         doc="QA related configuration options.",
     )
+    doHeaderProvenance = pexConfig.Field(
+        dtype=bool,
+        default=True,
+        doc="Write calibration identifiers into output exposure header?",
+    )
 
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -1611,6 +1665,31 @@ def ensureExposure(self, inputExp, camera=None, detectorNum=None):
...
+            Calibration to pull date information from.
+
+        Returns
+        -------
+        dateString : `str`
+            Calibration creation date string to add to header.
+        """
+        if hasattr(calib, "getMetadata"):
+            if 'CALIB_CREATION_DATE' in calib.getMetadata():
+                return " ".join((calib.getMetadata().get("CALIB_CREATION_DATE", "Unknown"),
```
mfisherlevine, line 1685: Presumably the `CALIB_CREATION_DATE` vs `CALIB_CREATE_DATE` is to do with versioning (I see you changing it on the other ticket), so maybe just a comment here about why this tried this first?

mfisherlevine (approved):

timj:
Quick comments:

* "LSST CALIB" is repeated a lot here so maybe factor it out.
* You can do `exposureMetadata["KEY"] = value`.
* We don't test the `run()` method anywhere?

PR #243, opened 2022-12-02
DM-34332:  Document ip_isr

czwa:
Update ancient docstrings to modern format.  Fix some remaining typos.  Populate `doc/` tree.


python/lsst/ip/isr/ampOffset.py
```diff
@@ -27,9 +27,13 @@
 from lsst.meas.algorithms import (SubtractBackgroundTask, SourceDetectionTask)
 
 
+__all__ = ['AmpOffsetConfig', 'AmpOffsetTask']
```
timj, line 30: Doesn't main already have this change (but with different quotes)?

timj: (Ah, I see that the main one is a few lines above this -- the `__all__` should always be before the first import)

python/lsst/ip/isr/isrQa.py
```diff
@@ -27,6 +27,9 @@
 import lsst.pex.config as pexConfig
 
 
+__all__ = ['IsrQaFlatnessConfig', 'IsrQaConfig', 'makeThumbnail']
```
timj, line 30: This one is also duplicating one that is higher up the file. I'll stop commenting about them now since you should be able to find ones that have multiple `__all__` in a single file.

python/lsst/ip/isr/ampOffset.py
czwa: It looks like the additions happened in parallel on different tickets, and I didn't notice in the diff because they're separated.  I'll go through and clean up the extras throughout.

python/lsst/ip/isr/isrQa.py
timj: It was all done in #227.

PR #244, opened 2022-12-06
DM-37011-v24: Remove second Notes section so docs build

mwittgen:


timj:
I fixed many sphinx build problems on main with #227 (this specific fix was in ad79872).

mwittgen:
Two many changes in https://github.com/lsst/ip_isr/commit/ad798728adc194f1f1eb9059f26c1819a0602ce1to cherry pick without
creating more conflicts. I just fixed the conflict preventing merging.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -146,6 +146,11 @@ class PhotonTransferCurveDataset(IsrCalib):
          Dictionary keyed by amp names containing the integrated photocharge
          for linearity calibration.
 
+    Returns
```
timj, line 149: I'm very confused. This addition is wrong. It should not be there because we do not include returns sections in class documentation (it was removed from main). The "Version 1.1" text should be part of the above Notes section not a returns section.

timj (changes requested):

mwittgen:
My conflict resolution was done before realizing that this branch was based on `main` and not `24.0.x`. Reverted to Yusra's original state.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -151,8 +151,6 @@ class PhotonTransferCurveDataset(IsrCalib):
     `lsst.cp.pipe.ptc.PhotonTransferCurveDataset`
```
timj, line 151: This returns section also needs to be removed (because otherwise the Version 1.1 text is in the wrong place and not part of the Notes).

timj (approved):
Looks good. Thanks.

mwittgen:
Renamed the target branch from `24.0.x` `v24.0.x`.
Thanks @yalsayyad for catching this.
I'll sleep over pushing the merge button.

PR #250, opened 2023-02-09
DM-36390: calibrations should track sequencer crc

czwa:
Add new IsrCalib method to handle sequencer checks.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -72,6 +72,9 @@ class IsrCalib(abc.ABC):
         self._detectorId = None
         self._filter = None
         self._calibId = None
+        self._seqfile = None
+        self._seqname = None
+        self._seqcksum = None
         self._metadata = PropertyList()
         self.setMetadata(PropertyList())
         self.calibInfoFromDict(kwargs)
@@ -80,7 +83,8 @@ class IsrCalib(abc.ABC):
         self.requiredAttributes = set(["_OBSTYPE", "_SCHEMA", "_VERSION"])
         self.requiredAttributes.update(["_instrument", "_raftName", "_slotName",
                                         "_detectorName", "_detectorSerial", "_detectorId",
-                                        "_filter", "_calibId", "_metadata"])
+                                        "_filter", "_calibId", "_seqfile", "_seqname", "_seqcksum",
+                                        "_metadata"])
 
         self.log = log if log else logging.getLogger(__name__)
 
@@ -261,10 +265,39 @@ class IsrCalib(abc.ABC):
         self._metadata["DET_SER"] = self._detectorSerial if self._detectorSerial else None
         self._metadata["FILTER"] = self._filter if self._filter else None
         self._metadata["CALIB_ID"] = self._calibId if self._calibId else None
+        self._metadata["SEQFILE"] = self._seqfile if self._seqfile else None
+        self._metadata["SEQNAME"] = self._seqname if self._seqname else None
+        self._metadata["SEQCKSUM"] = self._seqcksum if self._seqcksum else None
         self._metadata["CALIBCLS"] = get_full_type_name(self)
 
         mdSupplemental.update(kwargs)
         mdOriginal.update(mdSupplemental)
+
+    def updateMetadataFromExposures(self, exposures):
+        """Extract and unify metadata information.
+
+        Parameters
+        ----------
+        exposures : `list`
+            Exposures or other calibrations to scan.
+        """
+        keywords = ["SEQNAME", "SEQFILE", "SEQCKSUM"]
+        metadata = {}
+
+        for exp in exposures:
+            try:
+                expMeta = exp.getMetadata()
+            except AttributeError:
+                continue
+            for key in keywords:
+                if key in expMeta:
+                    if key in metadata:
+                        if metadata[key] != expMeta[key]:
+                            self.log.warning("Metadata mismatch! Have: %s Found %s",
+                                             metadata[key], expMeta[key])
+                    else:
+                        metadata[key] = expMeta[key]
+        self.updateMetadata(**metadata)
 
     def calibInfoFromDict(self, dictionary):
         """Handle common keywords.
@@ -318,6 +351,9 @@ class IsrCalib(abc.ABC):
         self._detectorSerial = search(dictionary, ["DET_SER", "DETECTOR_SERIAL", "detectorSerial"])
         self._filter = search(dictionary, ["FILTER", "filterName"])
         self._calibId = search(dictionary, ["CALIB_ID"])
+        self._seqfile = search(dictionary, ["SEQFILE"])
+        self._seqname = search(dictionary, ["SEQNAME"])
+        self._seqcksum = search(dictionary, ["SEQCKSUM"])
 
     @classmethod
     def determineCalibClass(cls, metadata, message):
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -79,7 +79,12 @@ class IsrCalib(abc.ABC):
         self.setMetadata(PropertyList())
         self.calibInfoFromDict(kwargs)
 
-        # Define the required attributes for this calibration.
+        # Define the required attributes for this calibration.  These
+        # are entries that are automatically filled and propagated
+        # from the metadata.  The existence of the attribute is
+        # required (they're checked for equivalence), but they do not
+        # necessarily need to have a value (None == None in this
+        # case).
         self.requiredAttributes = set(["_OBSTYPE", "_SCHEMA", "_VERSION"])
         self.requiredAttributes.update(["_instrument", "_raftName", "_slotName",
                                         "_detectorName", "_detectorSerial", "_detectorId",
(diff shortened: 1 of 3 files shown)


erykoff (approved):
Looks good with the new comments, but I think a more verbose config var would be useful.

tests/test_calibType.py
```diff
@@ -39,6 +47,8 @@ def setUp(self):
                                 {'exposure': 1235, 'detector': 0, 'filter': 'G'},
                                 {'exposure': 1234, 'detector': 1, 'filter': 'G'},
                                 {'exposure': 1235, 'detector': 1, 'filter': 'G'}])
+        self.exposure1 = FakeExposure(SEQNAME='sequencer A', SEQFILE='my.seq', SEQCKSUM='abcdef')
+        self.exposure2 = FakeExposure(SEQNAME='sequencer B', SEQFILE='my.seq')
```
mfisherlevine, line 51: This seems good to add, but I can't quite work out where this is being exercised. Shouldn't there be something with a `self. assertRaises()` and a config which says to raise, and ideally one with the warnings too?

python/lsst/ip/isr/calibType.py
```diff
@@ -261,11 +270,44 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
...
+        Parameters
+        ----------
+        exposures : `list`
+            Exposures or other calibrations to scan.
+        """
+        # This list of keywords is the set of header entries that
+        # should be checked and propagated.  Not having an entry is
+        # not a failure, as they may not be defined for the exposures
+        # being used.
+        keywords = ["SEQNAME", "SEQFILE", "SEQCKSUM", "ODP", "AP0_RC"]
```
mfisherlevine, line 293: Doesn't this need to come from the config now as well?

python/lsst/ip/isr/calibType.py
```diff
@@ -261,11 +270,44 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
         self._metadata["DET_SER"] = self._detectorSerial if self._detectorSerial else None
         self._metadata["FILTER"] = self._filter if self._filter else None
         self._metadata["CALIB_ID"] = self._calibId if self._calibId else None
+        self._metadata["SEQFILE"] = self._seqfile if self._seqfile else None
+        self._metadata["SEQNAME"] = self._seqname if self._seqname else None
+        self._metadata["SEQCKSUM"] = self._seqcksum if self._seqcksum else None
```
mfisherlevine, line 275: How come these are hard-coded still, even though this is now a config option? Doesn't this now need to be done dynamically somehow from the list in the config?

python/lsst/ip/isr/calibType.py
```diff
@@ -318,6 +360,9 @@ def search(haystack, needles):
         self._detectorSerial = search(dictionary, ["DET_SER", "DETECTOR_SERIAL", "detectorSerial"])
         self._filter = search(dictionary, ["FILTER", "filterName"])
         self._calibId = search(dictionary, ["CALIB_ID"])
+        self._seqfile = search(dictionary, ["SEQFILE"])
+        self._seqname = search(dictionary, ["SEQNAME"])
+        self._seqcksum = search(dictionary, ["SEQCKSUM"])
```
mfisherlevine, line 365: And again here.

python/lsst/ip/isr/calibType.py
```diff
@@ -72,15 +72,24 @@ def __init__(self, camera=None, detector=None, log=None, **kwargs):
         self._detectorId = None
         self._filter = None
         self._calibId = None
+        self._seqfile = None
+        self._seqname = None
+        self._seqcksum = None
```
mfisherlevine, line 77: See next comment, but shouldn't this now be dynamic?

python/lsst/ip/isr/calibType.py
```diff
@@ -72,15 +72,24 @@ def __init__(self, camera=None, detector=None, log=None, **kwargs):
...
+        # are entries that are automatically filled and propagated
+        # from the metadata.  The existence of the attribute is
+        # required (they're checked for equivalence), but they do not
+        # necessarily need to have a value (None == None in this
+        # case).
         self.requiredAttributes = set(["_OBSTYPE", "_SCHEMA", "_VERSION"])
         self.requiredAttributes.update(["_instrument", "_raftName", "_slotName",
                                         "_detectorName", "_detectorSerial", "_detectorId",
-                                        "_filter", "_calibId", "_metadata"])
+                                        "_filter", "_calibId", "_seqfile", "_seqname", "_seqcksum",
```
mfisherlevine, line 91: And here.

python/lsst/ip/isr/calibType.py
```diff
@@ -261,11 +270,44 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
...
+        # not a failure, as they may not be defined for the exposures
+        # being used.
+        keywords = ["SEQNAME", "SEQFILE", "SEQCKSUM", "ODP", "AP0_RC"]
+        metadata = {}
+
+        for exp in exposures:
+            try:
+                expMeta = exp.getMetadata()
+            except AttributeError:
+                continue
```
mfisherlevine, line 300: Out of interest, if calibs have this and exposures have this, what do you expect to not have a metadata item? And I take it you want them all to compare correctly then without log warning or anything?

python/lsst/ip/isr/calibType.py
```diff
@@ -261,11 +270,44 @@ def updateMetadata(self, camera=None, detector=None, filterName=None,
...
+            for key in keywords:
+                if key in expMeta:
+                    if key in metadata:
+                        if metadata[key] != expMeta[key]:
+                            self.log.warning("Metadata mismatch! Have: %s Found %s",
+                                             metadata[key], expMeta[key])
```
mfisherlevine, line 306: _Kinda_ wonder if this should raise... This is lossy, as we just end up taking the first value, so if you don't spot this warning you lose the info, right? We could make the item a list item in these cases, and yes, that would cause problems downstream, but that would be half the point - to let you know you have mixed inputs to a calib. The latter was just a passing thought and probably not a good idea, but the former point remains - should this raise or warn?

python/lsst/ip/isr/calibType.py
czwa: Reply in next comment.

python/lsst/ip/isr/calibType.py
czwa: I've been saving and skimming the calib construction log files, so I left this as a warning as it'd be something I'd look for.  There are also situations where I didn't think a raise was appropriate (crosstalk was my first thought, and I think flats would be another as the flux levels should swamp out any voltage related few-ADU differences).

python/lsst/ip/isr/isrTask.py
```diff
@@ -1728,6 +1752,36 @@ def extractCalibDate(calib):
...
+                    if self.config.doRaiseOnCalibMismatch:
+                        raise RuntimeError("Sequencer mismatch for %s [%s]: exposure: %s calib: %s",
+                                           calibName, keyword,
+                                           exposureMetadata[keyword], calibMetadata[keyword])
+                    else:
+                        self.log.warning("Sequencer mismatch for %s [%s]: exposure: %s calib: %s",
+                                         calibName, keyword,
+                                         exposureMetadata[keyword], calibMetadata[keyword])
+            else:
+                self.log.debug("Sequencer keyword %s not found.", keyword)
```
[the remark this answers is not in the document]
czwa: This is debug as it will just be noise for other cameras.

tests/test_calibType.py
czwa: This doesn't test that due to the same issue with calibs not having an `IsrTaskConfig` associated with them.

python/lsst/ip/isr/isrTask.py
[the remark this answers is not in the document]
mfisherlevine: But my point was that here you're inside a `for keyword in self.config.cameraKeywordsToCompare` block, and that will be an empty list for all these cameras for which it would be noise. So given that to be here all, that config list is being set, doesn't that showing up constitute something quite bad happening?

tests/test_calibType.py
mfisherlevine: Ah, OK, thanks.

python/lsst/ip/isr/calibType.py
mfisherlevine: Ah, OK, I see, thanks.

python/lsst/ip/isr/calibType.py
mfisherlevine: Interesting... Well I'll leave it to you then, but if it's easy it might be worth seeing what breaks if you remove that and at least popping a comment in there for future-you.

python/lsst/ip/isr/calibType.py
mfisherlevine: Oooh, right. That's a bummer.

python/lsst/ip/isr/calibType.py
mfisherlevine: OK, I see, thanks.

PR #252, opened 2023-02-16
DM-38044: Logging error in isrTask

czwa:
Fix broken log message.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -1341,18 +1341,29 @@ class IsrTask(pipeBase.PipelineTask):
                     self.log.debug("Corrected overscan for amplifier %s.", amp.getName())
                     if overscanResults is not None and \
                        self.config.qa is not None and self.config.qa.saveStats is True:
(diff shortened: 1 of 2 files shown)


mfisherlevine:
Looks good, thanks! General question/comment/request though: could we turn on debug level logging in some/all of the tests in here so that things like this are always being exercised and we don't run afoul of this again?

mfisherlevine (approved):

timj:
I think we should patch sconsUtils to turn in debug logging in all our testing rather than targeting this package

mfisherlevine:
> I think we should patch sconsUtils to turn in debug logging in all our testing rather than targeting this package

Totally agree - I wonder if it'll break the build!

mfisherlevine:
> > I think we should patch sconsUtils to turn in debug logging in all our testing rather than targeting this package
> 
> Totally agree - I wonder if it'll break the build!

When we do that though, we must be careful to set the `raiseExceptions` thing so that it does actually fault though!

timj:
> I wonder if it'll break the build

The answer is it breaks pipe_tasks...

PR #254, opened 2023-03-20
DM-38309: Update PhotonTransferCurveDataset to reduce memory and fix index errors.

erykoff:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -196,65 +197,76 @@ def __init__(self, ampNames=[], ptcFitType=None, covMatrixSide=1, **kwargs):
...
-        self.rawExpTimes[ampName] = rawExpTime
-        self.rawMeans[ampName] = rawMean
-        self.rawVars[ampName] = rawVar
-        self.photoCharge[ampName] = photoCharge
+        if covariance is None:
+            covariance = nanMatrix
+        if covSqrtWeights is None:
+            covSqrtWeights = nanMatrix
+
+        # This first one I'm not sure of the type; list of tuples is okay?
```
plazas, line 253: A list (or NumPy array, if you changed everything to NumPy arrays) of tuples should be fine, I would think. If so, don't forget to delete this comment.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -618,7 +622,31 @@ def getExpIdsUsed(self, ampName):
         pairs = self.inputExpIdPairs[ampName]
         mask = self.expIdMask[ampName]
         # cast to bool required because numpy
-        return [(exp1, exp2) for ((exp1, exp2), m) in zip(pairs, mask) if bool(m) is True]
+        try:
+            expIdsUsed = [(exp1, exp2) for ((exp1, exp2), m) in zip(pairs, mask) if m]
+        except ValueError:
+            warnings.warn("The PTC file was written incorrectly; you should rerun the "
```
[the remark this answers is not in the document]
plazas: I'm trying to decide if more details in this message would be useful for the user that encounters this.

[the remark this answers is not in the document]
plazas: And why would re-running `ptcSolve` fix it? I'm also thinking of the subset of `cpPtc.yaml` that only runs `isr` and `cpPtcExtract`.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -618,7 +622,31 @@ def getExpIdsUsed(self, ampName):
...
+                          "PTC solve task if possible.", RuntimeWarning)
+            expIdsUsed = []
+            for pairList, m in zip(pairs, mask):
+                if m:
+                    expIdsUsed.append(pairList[0])
+
+        return expIdsUsed
 
     def getGoodAmps(self):
```
plazas, line 637: Do we need a docstring here too?

python/lsst/ip/isr/ptcDataset.py
erykoff: Oops, that was a vestigial comment, I'll remove it.

tests/test_ptcDataset.py
```diff
@@ -61,84 +62,177 @@ def setUp(self):
...
+            localDataset.getExpIdsUsed("C:0,0")
+
+    def test_getGoodAmps(self):
+        dataset = self.dataset
+
+        self.assertTrue(dataset.ampNames == self.ampNames)
+        dataset.badAmps.append("C:0,1")
+        self.assertTrue(dataset.getGoodAmps() == [amp for amp in self.ampNames if amp != "C:0,1"])
+
+    def test_ptcDataset_pre_dm38309(self):
```
plazas, line 222: Do we need an analogous test for the partial PTC datasets created by `cpPtcExtract`?

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -64,85 +65,85 @@ class PhotonTransferCurveDataset(IsrCalib):
...
-    finalMeans : `dict`, [`str`, `list`]
+    finalMeans : `dict`, [`str`, `np.ndarray`]
         Dictionary keyed by amp names containing the masked average of the
         means of the exposures in each flat pair. If needed, each array
         will be right-padded with np.nan to match the length of
         rawExpTimes.
-    photoCharge : `dict`, [`str`, `list`]
+    photoCharges : `dict`, [`str`, `np.ndarray`]
          Dictionary keyed by amp names containing the integrated photocharge
          for linearity calibration.
```
plazas, line 149: Blow this line is the version. Does it need to be increased after the changes in this ticket? If so, will it affect backward compatibility?

plazas (approved):

czwa (approved):
I'm a bit concerned that removing all of the padding will cause issues if one amplifier has a different length than the rest, but if this updated version does work in that case without all the mess, then that's fine.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -196,65 +197,76 @@ def __init__(self, ampNames=[], ptcFitType=None, covMatrixSide=1, **kwargs):
...
+            Name of the amp to set the values.
+        inputExpIdPair : `tuple` [`int`]
+            Exposure IDs of input pair.
+        rawExpTime : `float`, optional
+            Exposure time for this exposure pair.
+        rawMean : `float`, optional
+            Average of the means of the exposures in this pair.
+        rawVar : `float`, optional
+            Variance of the difference of the exposures in this pair.
+        photoCharge : `float`, optional
```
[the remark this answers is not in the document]
erykoff: Note that these are all singular (intentionally) since they are single values to be set for a partial PTC dataset.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -519,68 +537,45 @@ def toTable(self):
...
+                'A_MATRIX': self.aMatrix[ampName].ravel(),
+                'B_MATRIX': self.bMatrix[ampName].ravel(),
+                'A_MATRIX_NO_B': self.aMatrixNoB[ampName].ravel(),
+                'BAD_AMPS': badAmps,
+                'PHOTO_CHARGE': self.photoCharges[ampName],
+                'COVARIANCES': self.covariances[ampName].ravel(),
+                'COVARIANCES_MODEL': self.covariancesModel[ampName].ravel(),
+                'COVARIANCES_SQRT_WEIGHTS': self.covariancesSqrtWeights[ampName].ravel(),
+                'COVARIANCES_MODEL_NO_B': self.covariancesModelNoB[ampName].ravel(),
+                'FINAL_VARS': self.finalVars[ampName],
```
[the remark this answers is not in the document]
erykoff: Neither, really.  The construction of the dataset is such that for a given detector, all amps have the same input pairs and therefore are all filled the same (though it may be with `nan`s).  I believe that previously the reason that this did not work is because (a) we were filtering out bad values on read, which messed up the lengths per amp; (b) we were filtering out bad values on read inconsistently because of the astropy bug, which messed up the lengths within an amp.

tests/test_ptcDataset.py
erykoff: Those are fine, in that the partial datasets didn't have the extra dimension.  But it's also not something that is a common use-case anyway as those are not really user-facing.

python/lsst/ip/isr/ptcDataset.py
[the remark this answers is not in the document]
erykoff: Ah yes, I left all of this part the same.  It's only the re-padding in the dataset that I removed.

PR #258, opened 2023-04-24
DM-38777: LinearizeSpline linearity corrections do not anchor the spline at zero flux

czwa:
Fix spline linearity-induced amplifier offsets.

### DIFF
--- a/python/lsst/ip/isr/linearize.py
+++ b/python/lsst/ip/isr/linearize.py
@@ -412,9 +412,12 @@ class Linearizer(IsrCalib):
             if int(self._detectorId) != int(detector.getId()):
                 raise RuntimeError("Detector IDs don't match: %s != %s" %
                                    (int(self._detectorId), int(detector.getId())))
-            if self._detectorSerial != detector.getSerial():
-                raise RuntimeError("Detector serial numbers don't match: %s != %s" %
-                                   (self._detectorSerial, detector.getSerial()))
+            # DM-38778: This check fails on LATISS due to an error in the
+            #           camera configuration.
+            # if self._detectorSerial != detector.getSerial():
+            #      raise RuntimeError(
+            #          "Detector serial numbers don't match: %s != %s" %
+            #          (self._detectorSerial, detector.getSerial()))
             if len(detector.getAmplifiers()) != len(self.linearityCoeffs.keys()):
                 raise RuntimeError("Detector number of amps = %s does not match saved value %s" %
                                    (len(detector.getAmplifiers()),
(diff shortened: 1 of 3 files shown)


python/lsst/ip/isr/linearize.py
```diff
@@ -753,6 +756,14 @@ def __call__(self, image, **kwargs):
         """
         splineCoeff = kwargs['coeffs']
         centers, values = np.split(splineCoeff, 2)
+        if centers[0] != 0.0 or values[0] != 0:
```
taranu, line 759: Is this deliberately `or` rather than a nested `if` as in `cp_pipe`?

tests/test_linearize.py
```diff
@@ -125,8 +125,8 @@ def setUp(self):
                                     [[1e-5, 1e-7], [1.1e-6, 1e-7], [2.1e-6, 1e-7]]], dtype=float)
 
         # Spline coefficients: should match a 1e-6 Squared solution
-        self.splineCoeffs = np.array([-100, 0.0, 1000, 2000, 3000, 4000, 5000,
-                                      0.0, 0.0, 1.0, 4.0, 9.0, 16.0, 25.0])
```
taranu, line 129: Hmm, presumably this would have generated at least slightly negative values between -100 and 0 before, no? Did that cause any issues? But either way, starting it at zero definitely seems to make more sense.

python/lsst/ip/isr/linearize.py
```diff
@@ -412,9 +412,12 @@ def validate(self, detector=None, amplifier=None):
             if int(self._detectorId) != int(detector.getId()):
                 raise RuntimeError("Detector IDs don't match: %s != %s" %
                                    (int(self._detectorId), int(detector.getId())))
-            if self._detectorSerial != detector.getSerial():
-                raise RuntimeError("Detector serial numbers don't match: %s != %s" %
-                                   (self._detectorSerial, detector.getSerial()))
+            # DM-38778: This check fails on LATISS due to an error in the
```
taranu, line 415: Consider prefacing this and below note with `TODO: Uncomment when DM-38778 is complete...`, if you find TODO tags helpful.

python/lsst/ip/isr/linearize.py
czwa: I thought it was correct as there are two issues to correct, and then discovered that giving the interpolation code centers like `[0.0, 0.0, 1.0, 2.0, 3.0, 4.0, 5.0]` causes an error even if the values for the two `0.0` points is identical.  I've split the two issues into separate checks, which should prevent that interpolation error.

tests/test_linearize.py
czwa: I think it was just poor thinking on my part when I wrote the test.  If that does match the `1e-6` squared solution, then that negative point gains flux from a correction entirely derived from the positive data, which doesn't make any sense at all.

taranu (approved):

PR #259, opened 2023-05-01
DM-38973: Call to np.percentile in overscan.py leads to numpy warnings with 1.23

czwa:
Pass only data to np.percentile from overscan maskedArray.

### DIFF
--- a/python/lsst/ip/isr/overscan.py
+++ b/python/lsst/ip/isr/overscan.py
@@ -617,7 +617,8 @@ class OverscanCorrectionTask(pipeBase.Task):
         maskedArray : `numpy.ma.masked_array`
             Masked image marking outliers.
         """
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/overscan.py
```diff
@@ -617,7 +617,8 @@ def maskOutliers(self, imageArray):
         maskedArray : `numpy.ma.masked_array`
             Masked image marking outliers.
         """
-        lq, median, uq = np.percentile(imageArray, [25.0, 50.0, 75.0], axis=1)
+        lq, median, uq = np.percentile(np.ma.getdata(imageArray),
+                                       [25.0, 50.0, 75.0], axis=1)
```
erykoff, line 621: I was worried that this wouldn't work if it wasn't a masked array, but it does, so great!

erykoff (approved):

PR #261, opened 2023-05-09
DM-39048: Add Gaussian histogram fit values to ptc dataset.

erykoff:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -526,6 +561,15 @@ def fromTable(cls, tableList):
                     inDict['ptcTurnoff'][ampName] = np.nan
             else:
                 inDict['ptcTurnoff'][ampName] = record['PTC_TURNOFF']
+            if calibVersion == 1.1:
+                inDict['histVars'][ampName] = np.nan
```
czwa, line 565: Do these need to be an array type?  `fromDict` is casting it into a numpy array above, but it won't have the same length as there are pairs of inputs (except in the partialPtc case).  Leaving this as is I think just means that any future code that wants to plot these values as a function of mean/variance needs to check the version before doing so.

czwa: This also has a different default value for `kspValues` than that set in `cpPtcExtractTask.py:L535`.

czwa (approved):
Minor question that I'm not sure of the correct answer.

python/lsst/ip/isr/ptcDataset.py
erykoff: Both very good points!  I have to fix the docstring as well.

PR #262, opened 2023-05-12
DM-38776: Generalize amp-to-amp offset correction task

enourbakhsh:


python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
         default=5.0,
     )
-    ampEdgeWindow = pexConfig.Field(
-        doc="Pixel size of the sliding window used to generate rolling average amp offset values.",
+    ampEdgeWindow = Field(
+        doc="Pixel size of the sliding window used to generate rolling average amp offset values. It should "
+        "not exceed the shorter dimension of amps.",
         dtype=int,
-        default=512,
+        default=509,  # We need to reconsider this for every instrument (subaru, lsst, etc.)
```
leeskelvin, line 85: "Subaru, LSST, etc."

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+            for ii, (amp, pedestal) in enumerate(zip(amps, pedestals)):
+                ampIm = exposure.image[amp.getBBox()].array
+                ampIm -= pedestal
+                metadata.set(
+                    f"PEDESTAL{ii + 1}", float(pedestal), f"Pedestal level subtracted from amp {ii + 1}"
+                )
+            self.log.info(f"amp pedestal values: {', '.join([f'{x:.4f}' for x in pedestals])}")
+
+        # raise NotImplementedError("Amp offset task should be retargeted
+        # by a camera specific version.")
```
leeskelvin, line 200: This can probably be removed now.

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+
+class AmpOffsetConfig(Config):
+    """Configuration parameters for AmpOffsetTask."""
+
+    def setDefaults(self):
+        # The `binSize`` should be as large as possible, preventing background
+        # subtraction from inadvertently removing the amp offset signature.
+        # Here it's set to the shorter dimension of an LSST amp by default,
+        # which seems reasonable for LSST.
+        self.background.binSize = 509
```
leeskelvin, line 42: Add a comment after 509: 
```
self.background.binSize = 509  # Default for LSSTCam.
```

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+
+            # Determine amplifier interface geometry.
+            ampAreas = {amp.getBBox().getArea() for amp in amps}
+            if len(ampAreas) > 1:
+                raise NotImplementedError(
+                    "Amp offset correction is not yet implemented for detectors with differing amp sizes."
+                )
+
+            # Obtain association and offset matrices for the subsequent
+            # operation.
```
leeskelvin, line 183: A simple one-liner would suffice here:
```
# Obtain association and offset matrices.
```

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+        -------
+        ampAssociations : `numpy.ndarray`
+            A N x N matrix (N = the number of amplifiers) providing the
+            association information for amplifiers, indicating their
+            connections within the detector layout. The elements are defined as
+            below: TODO
+            -1: ...
+            ...: ...
+        ampInterfaces : `numpy.ndarray`
+            A N x N matrix (N = the number of amplifiers) representing the amp
```
leeskelvin, line 228: An N x N...

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+            Exposure to be corrected for amp offsets.
+        """
+
+        # Generate an exposure clone to work on and establish the bit mask.
+        exp = exposure.clone()
+        bitMask = exp.mask.getPlaneBitMask(self.background.config.ignoredPixelMask)
+        self.log.info(
+            "Ignored mask planes for amp offset estimation: [%s].",
+            ", ".join(self.background.config.ignoredPixelMask),
+        )
```
leeskelvin, line 137: The `background` task will already print something like this in the logs if the log level is set to `DEBUG` mode:
```
DEBUG 2023-08-01T12:04:06.874-07:00 lsst.isr.ampOffset.background (isr:{instrument: 'HSC', detector: 42, exposure: 1228, ...})(subtractBackground.py:255) - Ignoring mask planes: BAD, SAT, INTRP, CR, EDGE, DETECTED, DETECTED_NEGATIVE, SUSPECT, NO_DATA
```
I'm not sure we want it repeated at INFO level too, do we?

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+        for ampId in ampIds.ravel():
+            adjacents, interfaces = self.getAdjacents(ampIds, ampId)
+            ampAssociations[ampId, adjacents] = -1
+            ampInterfaces[ampId, adjacents] = interfaces
+            ampAssociations[ampId, ampId] = -ampAssociations[ampId].sum()
+
+        if ampAssociations.sum() != 0:
+            raise RuntimeError("The `ampAssociations` array does not sum to zero.")
+
+        return ampAssociations, ampInterfaces
```
leeskelvin, line 260: Before this return, I think it would be nice to add the option for displaying these two matrices in a debug logger:
```
self.log.debug(f"amp associations:\n{ampAssociations}")
self.log.debug(f"amp interfaces:\n{ampInterfaces}")
```
which produces an output like this:
```
DEBUG 2023-08-01T12:10:55.114-07:00 lsst.isr.ampOffset (isr:{instrument: 'HSC', detector: 42, exposure: 1228, ...})(ampOffset.py:260) - amp associations:
[[ 1 -1  0  0]
 [-1  2 -1  0]
 [ 0 -1  2 -1]
 [ 0  0 -1  1]]
DEBUG 2023-08-01T12:10:55.117-07:00 lsst.isr.ampOffset (isr:{instrument: 'HSC', detector: 42, exposure: 1228, ...})(ampOffset.py:261) - amp interfaces:
[[-1  1 -1 -1]
 [ 3 -1  1 -1]
 [-1  3 -1  1]
 [-1 -1  3 -1]]
```

leeskelvin: Actually, we should probably move away from using f-strings in loggers, as it causes the log messages to be called even when not shown. The alternative which is better here (and every other log message in this file) should look like this instead:
```
self.log.debug("amp associations:\n%s", ampAssociations)
self.log.debug("amp interfaces:\n%s", ampInterfaces)
```

python/lsst/ip/isr/ampOffset.py
leeskelvin: Quite right.
```
"This should be reconfigured for every instrument (HSC, LSSTCam, etc.)"
```

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+
+            # If least-squares minimization fails, convert NaNs to zeroes,
+            # ensuring that no values are erroneously added/subtracted
+            pedestals = np.nan_to_num(np.linalg.lstsq(A, B, rcond=None)[0])
+            metadata = exposure.getMetadata()
+            for ii, (amp, pedestal) in enumerate(zip(amps, pedestals)):
+                ampIm = exposure.image[amp.getBBox()].array
+                ampIm -= pedestal
+                metadata.set(
+                    f"PEDESTAL{ii + 1}", float(pedestal), f"Pedestal level subtracted from amp {ii + 1}"
```
leeskelvin, line 195: This  +1 will be incorrect for other cameras. In HSC, it was true that the zeroth amp had an ID of 1, the first amp had an ID of 2, and so on. 

For this generalized task, we'll need to figure out a way to pass the actual amp ID into this message, and propagate it through.

python/lsst/ip/isr/ampOffset.py
enourbakhsh: Probably not

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,412 @@
...
+            self.log.warning("All pixels masked: cannot calculate any amp offset corrections.")
+        else:
+            # Set up amp offset inputs.
+            im = exp.image
+            im.array[(exp.mask.array & bitMask) > 0] = np.nan
+            amps = exp.getDetector().getAmplifiers()
+
+            min_dimension = np.min([np.min(amp.getBBox().getDimensions()) for amp in amps])
+            if self.config.ampEdgeWindow is None:
+                self.config.ampEdgeWindow = min_dimension
```
czwa, line 163: Setting values in the config won't work here, as the config has been frozen.  Some other parameter needs to be used to hold this size.

czwa: Also, if this check is only for the downstream `np.convolve`, then you can push this check/setting there instead.  If the default setting is to set it to the smaller size, then there's no reason to `raise` here.  You should allow smaller sizes in the config, as there may be circumstances where the smaller size is better.

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,412 @@
...
+        amps : `list` [`lsst.afw.cameraGeom.Amplifier`]
+            List of amplifier objects.
+        ampSides : `numpy.ndarray`
+            An N x N matrix containing amp side information, where N is the
+            number of amplifiers.
+
+        Returns
+        -------
+        ampEdges : `dict` [`int`, `dict` [`int`, `numpy.ndarray`]]
+        A dictionary containing amp edge(s) for each amplifier, corresponding
```
czwa, line 400: Docstring formatting issue.

python/lsst/ip/isr/ampOffset.py
czwa: This should just be `amp.getId()` from the amplifier object.  `.getName()` may also be useful if this is for humans.

python/lsst/ip/isr/ampOffset.py
enourbakhsh: Removed

python/lsst/ip/isr/ampOffset.py
enourbakhsh: amp.getName() worked!

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,412 @@
...
+        """
+
+        # Generate an exposure clone to work on and establish the bit mask.
+        exp = exposure.clone()
+        bitMask = exp.mask.getPlaneBitMask(self.background.config.ignoredPixelMask)
+
+        # Fit and subtract background.
+        if self.config.doBackground:
+            maskedImage = exp.getMaskedImage()
+            bg = self.background.fitBackground(maskedImage)
```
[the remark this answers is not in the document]
enourbakhsh: Introduced a new parameter called `backgroundFractionSample` that allows the user to input the fraction of the shorter side of the amplifier used for background binning through `nx` and `ny`. The default value for this parameter is set to 1.

python/lsst/ip/isr/ampOffset.py
enourbakhsh: Pushed the check to the downstream `np.convolve`. Smaller sizes are allowed in the config.

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,433 @@
...
+            2: (r - 1, c),
+            3: (r, c - 1),
+        }
+        for side, (row, column) in sideLookup.items():
+            if 0 <= row < m and 0 <= column < n:
+                neighbors.append(ampIds[row][column])
+                sides.append(side)
+        return neighbors, sides
+
+    def getInterfaceOffset(self, ampIdA, ampIdB, edgeA, edgeB):
```
czwa, line 306: This is me being picky, but can this method be moved down after `getAmpEdges`?  This orders it after the method that calls it, and after the method that is called prior to it.

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,433 @@
...
+        """
+        sctrl = StatisticsControl()
+        # NOTE: Taking the difference with the order below fixes the sign flip
+        # in the B matrix.
+        edgeDiff = edgeA - edgeB
+        if self.config.ampEdgeWindow == 0:
+            if len(edgeDiff) > self.shortAmpSide:
+                window = self.shortAmpSide
+            else:
+                window = int(self.shortAmpSide**2 / self.longAmpSide)
```
czwa, line 335: I'm sure I'm missing something, but can you explain this choice?  This looks like it's triggered when the side under consideration is parallel to the short axis of the amplifier.  Why should the window used for that be shorter the longer the other axis is?

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,433 @@
...
+        # Perform a couple of do-no-harm safety checks:
+        # a) The fraction of unmasked pixel rows is > ampEdgeMinFrac,
+        # b) The absolute offset ADU value is < ampEdgeMaxOffset.
+        ampEdgeGoodFrac = 1 - (np.sum(np.isnan(edgeDiffAvg)) / len(edgeDiffAvg))
+        minFracFail = ampEdgeGoodFrac < self.config.ampEdgeMinFrac
+        maxOffsetFail = np.abs(interfaceOffset) > self.config.ampEdgeMaxOffset
+        if minFracFail or maxOffsetFail:
+            interfaceOffset = 0
+        interfaceId = f"{ampIdA}{ampIdB}"
+        self.log.debug(
```
czwa, line 359: I think this needs a message at a different level (warning?) if this fails.  Otherwise, no-offset cases and failed measurements result look identical.

tests/test_ampOffset.py
```diff
@@ -0,0 +1,110 @@
...
+    def testAmpOffset(self):
+        if self.camera is None:
+            raise unittest.SkipTest("Skipping due to no camera set.")
+        exp = self.buildExposure(addBackground=False)
+        amps = exp.getDetector().getAmplifiers()
+        config = AmpOffsetConfig()
+        config.doBackground = False
+        config.doDetection = False
+        task = AmpOffsetTask(config=config)
+        task.run(exp)
```
czwa, line 60: If this is changed to return a struct, it can be used here to directly compare the values without needing to go through the metadata object.

tests/test_ampOffset.py
```diff
@@ -0,0 +1,110 @@
...
+        nAmps = len(amps)
+        ampNames = []
+        for amp in amps:
+            ampNames.append(amp.getName())
+        for i in range(nAmps // 2):
+            self.assertAlmostEqual(
+                meta[f"LSST ISR AMPOFFSET PEDESTAL {ampNames[i]}"],
+                -meta[f"LSST ISR AMPOFFSET PEDESTAL {ampNames[nAmps-i-1]}"],
+                7,
+            )
```
czwa, line 87: Can this check that the measured values are only minimally different than the input pedestals?  Suggestions on JIRA.

czwa (changes requested):

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,406 @@
...
+        amps : `list` [`lsst.afw.cameraGeom.Amplifier`]
+            List of amplifier objects used to deduce associations.
+
+        Returns
+        -------
+        ampAssociations : `numpy.ndarray`
+            A N x N matrix (N = the number of amplifiers) providing the
+            association information for amplifiers, indicating their
+            connections within the detector layout. The elements are defined as
+            below: TODO
```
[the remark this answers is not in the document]
enourbakhsh: Done!

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,433 @@
...
-class AmpOffsetTask(pipeBase.Task):
-    """Calculate and apply amp offset corrections to an exposure.
-    """
+class AmpOffsetTask(Task):
+    """Calculate and apply amp offset corrections to an exposure."""
+
     ConfigClass = AmpOffsetConfig
     _DefaultName = "isrAmpOffset"
 
     def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
enourbakhsh: Done! I only found `self.shortAmpSide`.

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,433 @@
...
+            table = SourceTable.make(schema)
+            # Detection sigma, used for smoothing and to grow detections, is
+            # normally measured from the PSF of the exposure. As the PSF hasn't
+            # been measured at this stage of processing, sigma is instead
+            # set to an approximate value here (which should be sufficient).
+            _ = self.detection.run(table=table, exposure=exp, sigma=2)
+
+        # Safety check: do any pixels remain for amp offset estimation?
+        if (exp.mask.array & bitMask).all():
+            self.log.warning("All pixels masked: cannot calculate any amp offset corrections.")
```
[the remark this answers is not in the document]
enourbakhsh: Done!

python/lsst/ip/isr/ampOffset.py
```diff
@@ -18,88 +18,433 @@
...
+            for amp, pedestal in zip(amps, pedestals):
+                ampIm = exposure.image[amp.getBBox()].array
+                ampIm -= pedestal
+                ampName = amp.getName()
+                metadata.set(
+                    f"LSST ISR AMPOFFSET PEDESTAL {ampName}",
+                    float(pedestal),
+                    f"Pedestal level subtracted from amp {ampName}"
+                )
+            self.log.info(f"amp pedestal values: {', '.join([f'{x:.4f}' for x in pedestals])}")
```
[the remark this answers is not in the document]
enourbakhsh: It now returns `Struct(pedestal=pedestals)`.

python/lsst/ip/isr/ampOffset.py
enourbakhsh: Moved it.

python/lsst/ip/isr/ampOffset.py
enourbakhsh: We went with this method to keep things in line with the way it was done before in `obs_subaru`. The main purpose was to make sure the pixel size of the sliding window matches up with what's used in `obs_subaru`, so it works well with the HSC data we've already got. Now, I've simplified it further by replacing `ampEdgeWindow` with `ampEdgeWindowFrac`. With this new config option, we can easily say how much of the amp edge lengths we want as our sliding window for those rolling averages. The default is set to 512/4176, which matches the previous HSC setup.

enourbakhsh: `self.longAmpSide` has been removed as a result of introducing `ampEdgeWindowFrac`.

python/lsst/ip/isr/ampOffset.py
enourbakhsh: Done!

tests/test_ampOffset.py
enourbakhsh: Done!

tests/test_ampOffset.py
enourbakhsh: Done!

czwa (approved):
I think I'm happy now.  The scatter (sigma) in the test values is higher than I would have thought, but it's possible that's an indication that edge-length weighting is needed.  Looking into that is out-of-scope for this ticket, but I think it's worth adding a new ticket to do that work.

PR #263, opened 2023-05-15
DM-38827: Possible inconsistency in indexing in the brighter fatter kernel generation/correction

czwa:
Fix incorrect BF kernel transposes.


mfisherlevine (approved):
Looks good. Big 👍 for the nicely explicit comments too.

PR #265, opened 2023-05-23
DM-39338: flatten arrays in photodiode calibration

czwa:
Flatten arrays, and remove 2d indexing in integrateChargeSum.
Add unit tests for array shape, and for integrateChargeSum return values.

### DIFF
--- a/python/lsst/ip/isr/photodiode.py
+++ b/python/lsst/ip/isr/photodiode.py
@@ -65,11 +65,11 @@ class PhotodiodeCalib(IsrCalib):
             if len(timeSamples) != len(currentSamples):
                 raise RuntimeError(f"Inconsitent vector lengths: {len(timeSamples)} vs {len(currentSamples)}")
             else:
-                self.timeSamples = np.array(timeSamples).flatten()
-                self.currentSamples = np.array(currentSamples).flatten()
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/photodiode.py
```diff
@@ -65,11 +65,11 @@ def __init__(self, timeSamples=None, currentSamples=None, **kwargs):
             if len(timeSamples) != len(currentSamples):
                 raise RuntimeError(f"Inconsitent vector lengths: {len(timeSamples)} vs {len(currentSamples)}")
             else:
-                self.timeSamples = np.array(timeSamples)
-                self.currentSamples = np.array(currentSamples)
+                self.timeSamples = np.array(timeSamples).flatten()
+                self.currentSamples = np.array(currentSamples).flatten()
```
erykoff, line 69: These should be `ravel()` not `flatten()` which is a much faster no-copy operation.  (Also below wherever `flatten()` is used).

erykoff: I'm also confused how this could have ever been more than 1d, some explanation would be helpful!

erykoff (changes requested):

erykoff (approved):

PR #266, opened 2023-05-31
DM-39460: Bug-fix for CHARGE_SUM photodiode integration method

jchiang87:


python/lsst/ip/isr/photodiode.py
```diff
@@ -297,10 +297,17 @@ def integrateChargeSum(self):
...
+        # signal levels > 5% of the maximum (measured relative to the
+        # overall minimum), and extend that selection 2 entries on
+        # either side to avoid otherwise low-valued points that sample
+        # the signal ramp and which should not be included in the
+        # baseline estimate.
         dy = np.max(current) - np.min(current)
-        index = np.where(current < dy/20. + np.min(current))
-        bg_current = np.sum(charge[index])/np.sum(dt[index])
+        signal = np.where(current > dy/20. + np.min(current))
+        imin = signal[0][0] - 2
```
erykoff, line 308: What happens if the first point is at the edge?  Do we need to make sure this doesn't go below 0?  (Similar for the next line at the max).

python/lsst/ip/isr/photodiode.py
```diff
@@ -297,10 +297,17 @@ def integrateChargeSum(self):
...
+        imax = signal[0][-1] + 2
+        bg = (np.array(list(range(0, imin)) + list(range(imax, len(current)))),)
```
erykoff, line 310: If I'm reading this correctly, this is a tuple of an array of a list  .... 
This should be simpler and I think should work the same.  There's no need for the 1d tuple:
```
bg = np.concatenate([np.arange(0, imin), np.arange(imax, len(current))])
```

python/lsst/ip/isr/photodiode.py
```diff
@@ -297,10 +297,17 @@ def integrateChargeSum(self):
...
+        # To determine the baseline current level, exclude points with
+        # signal levels > 5% of the maximum (measured relative to the
+        # overall minimum), and extend that selection 2 entries on
+        # either side to avoid otherwise low-valued points that sample
+        # the signal ramp and which should not be included in the
+        # baseline estimate.
         dy = np.max(current) - np.min(current)
-        index = np.where(current < dy/20. + np.min(current))
-        bg_current = np.sum(charge[index])/np.sum(dt[index])
+        signal = np.where(current > dy/20. + np.min(current))
```
erykoff, line 307: If this is a 1d array, you can do `signal, = np.where(current > dy/20. + np.min(current))` and just get the indices and below you can do `signal[0]` and `signal[-1]`.  And although this is what I still do, the numpy recommendation is instead to do `signal, = (current > dy/20. + np.min(current)).nonzero()`.

erykoff (commented):
Some small code change suggestions.  And for the tests, I'm wondering if there's a way of adding a pathological point like what triggered this ticket to ensure that it still works with and without one of those problematic points at the rolloff?

python/lsst/ip/isr/photodiode.py
jchiang87: Both `list(range(0, imin))` and `list(range( imax, len(current)))` return empty lists when `imin < 0` and `imax > len(current)`, respectively, which is what we want, I think.  Same appears to be true for `np.arange`.

jchiang87:
@erykoff I've implemented your suggestions and added some test code for that failure mode.

erykoff (approved):

PR #269, opened 2023-06-13
DM-39604: Record full noise matrix in PTC dataset

plazas:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -569,6 +597,13 @@ def fromTable(cls, tableList):
                 inDict['histVars'][ampName] = record['HIST_VARS']
                 inDict['histChi2Dofs'][ampName] = record['HIST_CHI2_DOFS']
                 inDict['kspValues'][ampName] = record['KS_PVALUES']
+            if calibVersion == 1.3:
```
erykoff, line 600: This should be `calibVersion >= 1.3`.  And I realized that above it should be `if calibVersion < 1.2`.  Or perhaps for better parallelism, this new code should be reversed, and `calibVersion < 1.3` should be checked to set to nans, and else is to use the noise matrix.

erykoff (changes requested):
A small comment on the version checking to make it forward-compatible.

python/lsst/ip/isr/ptcDataset.py
plazas: I just changed to `calibVersion >= 1.3` and `if calibVersion < 1.2`, as you suggest.  And the docstring was missing the comment for v1.2; I added "Version 1.2 adds the `histVars`, `histChi2Dofs`, and `kspValues`
    attributes.". Please let me know if this is OK.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -561,14 +589,21 @@ def fromTable(cls, tableList):
...
                 inDict['histVars'][ampName] = record['HIST_VARS']
                 inDict['histChi2Dofs'][ampName] = record['HIST_CHI2_DOFS']
                 inDict['kspValues'][ampName] = record['KS_PVALUES']
+            if calibVersion >= 1.3:
+                inDict['noiseMatrix'][ampName] = record['NOISE_MATRIX']
+                inDict['noiseMatrixNoB'][ampName] = record['NOISE_MATRIX_NO_B']
+            else:
+                nanMatrix = np.full_like(inDict['aMatrix'][ampName], np.nan)
+                inDict['noiseMatrix'][ampName] = nanMatrix
+                inDict['noiseMatrixNoB'][ampName] = nanMatrix
```
erykoff, line 606: I know this is a small thing, but I think that it would be better if we used the same construction for both calibVersion checks.  Something like this:
```
            if calibVersion < 1.2:
                inDict['histVars'][ampName] = np.array([np.nan])
                inDict['histChi2Dofs'][ampName] = np.array([np.nan])
                inDict['kspValues'][ampName] = np.array([0.0])
            else:
                inDict['histVars'][ampName] = record['HIST_VARS']
                inDict['histChi2Dofs'][ampName] = record['HIST_CHI2_DOFS']
                inDict['kspValues'][ampName] = record['KS_PVALUES']
            if calibVersion < 1.3:
                nanMatrix = np.full_like(inDict['aMatrix'][ampName], np.nan)
                inDict['noiseMatrix'][ampName] = nanMatrix
                inDict['noiseMatrixNoB'][ampName] = nanMatrix
                inDict['noiseMatrix'][ampName] = record['NOISE_MATRIX']
                inDict['noiseMatrixNoB'][ampName] = record['NOISE_MATRIX_NO_B']
            else:
                inDict['noiseMatrix'][ampName] = record['NOISE_MATRIX']
                inDict['noiseMatrixNoB'][ampName] = record['NOISE_MATRIX_NO_B']
```

erykoff (approved):
Small comment just to make things more readable, but otherwise looks good.

python/lsst/ip/isr/ptcDataset.py
plazas: Sounds good, I changed it to this. Thanks.

PR #271, opened 2023-06-20
DM-39726: Use float64 rather than generic float type

timj:
Modern numpy requires a size be picked.

### DIFF
--- a/python/lsst/ip/isr/deferredCharge.py
+++ b/python/lsst/ip/isr/deferredCharge.py
@@ -81,7 +81,7 @@ class SerialTrap():
             raise ValueError('Unknown trap type: %s', self.trap_type)
 
         if self.trap_type == 'spline':
-            centers, values = np.split(np.array(self.coeffs, dtype=np.float), 2)
+            centers, values = np.split(np.array(self.coeffs, dtype=np.float64), 2)
             # Ensure all NaN values are stripped out
             values = values[~np.isnan(centers)]
             centers = centers[~np.isnan(centers)]


python/lsst/ip/isr/deferredCharge.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, size, emission_time, pixel, trap_type, coeffs):
             raise ValueError('Unknown trap type: %s', self.trap_type)
 
         if self.trap_type == 'spline':
-            centers, values = np.split(np.array(self.coeffs, dtype=np.float), 2)
+            centers, values = np.split(np.array(self.coeffs, dtype=np.float32), 2)
```
timj, line 84: I picked float32 here but let me know if it's float64.

python/lsst/ip/isr/deferredCharge.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, size, emission_time, pixel, trap_type, coeffs):
             raise ValueError('Unknown trap type: %s', self.trap_type)
 
         if self.trap_type == 'spline':
-            centers, values = np.split(np.array(self.coeffs, dtype=np.float), 2)
+            centers, values = np.split(np.array(self.coeffs, dtype=np.float64), 2)
```
czwa, line 84: I think everything that comes from a calibration should be `float64`, so I'm happy with this fix.

czwa (approved):

erykoff (approved):

PR #275, opened 2023-07-16
DM-40049: fix the inconsistency in indexing in the brighter fatter correction

youtsumi:
fix the inconsistency in indexing https://jira.lsstcorp.org/browse/DM-40049


youtsumi:
merged by this change https://github.com/lsst/ip_isr/pull/276

PR #276, opened 2023-07-19
DM-40094  Convolution control edge copy method results in diverging correction in fluxConservingBrighterFatterCorrection

czwa:
I've also folded in the change from youtsumi:u/youtsumi/DM-40049 to fix the max iteration log message.


erykoff (approved):
I wish these things were tested...

PR #279, opened 2023-08-29
DM-40128: Fix brighter-fatter loading errors

czwa:
Simplify header provenance code and fix BF errors:
* Move calibration date searches to runQuantum with the rest of the header provenance.
* Handle inputs that change name during the header provenance (only brighter-fatter kernels currently).
* Retain header validation checks in the run method.

### DIFF
--- a/python/lsst/ip/isr/isrTask.py
+++ b/python/lsst/ip/isr/isrTask.py
@@ -1027,6 +1027,9 @@ class IsrTask(pipeBase.PipelineTask):
 
         detector = inputs['ccdExposure'].getDetector()
 
+        # This is use for header provenance.
+        additionalInputDates = {}
+
         if self.config.doCrosstalk is True:
             # Crosstalk sources need to be defined by the pipeline
             # yaml if they exist.
@@ -1071,6 +1074,7 @@ class IsrTask(pipeBase.PipelineTask):
 
         # Load the correct style of brighter-fatter kernel, and repack
         # the information as a numpy array.
+        brighterFatterSource = None
         if self.config.doBrighterFatter:
             brighterFatterKernel = inputs.pop('newBFKernel', None)
             if brighterFatterKernel is None:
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTask.py
```diff
@@ -1071,13 +1074,20 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
                 # component of the afwImage kernel.
                 brighterFatterKernel = inputs.get('bfKernel', None)
+                brighterFatterSource = 'bfKernel'
+                additionalInputDates[brighterFatterSource] = self.extractCalibDate(brighterFatterKernel)
+
+            if brighterFatterKernel is None:
+                # This was requested by the config, but none were found.
+                raise RuntimeError("No brighter-fatter kernel was supplied.")
 
             if brighterFatterKernel is not None and not isinstance(brighterFatterKernel, numpy.ndarray):
```
taranu, line 1092: Can't this just be `elif not isinstance(brighterFatterKernel, numpy.ndarray)` now?

python/lsst/ip/isr/isrTask.py
```diff
@@ -1128,16 +1141,30 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         if self.config.doHeaderProvenance:
             # Add calibration provenanace info to header.
             exposureMetadata = inputs['ccdExposure'].getMetadata()
-            for inputName in sorted(inputs.keys()):
+
+            # These inputs change name during this step.  These should
+            # have matching entries in the additionalInputDates dict.
+            additionalInputs = []
```
taranu, line 1147: Not that it matters much but couldn't you just set this to `list(inputs.keys())` to start with and append to/extend it?

taranu (approved):
Looks fine. Feel free to ignore the nitpicky suggestions.

python/lsst/ip/isr/isrTask.py
czwa: That would certainly work the same, but I think it would require some variable name and comment rewrites.  I think in the long term, this can eventually go away if all cameras switch to `IsrCalib` style brighter-fatter kernels.

PR #280, opened 2023-09-19
DM-36647: First draft of new isr task 

aferte:
Here is a first draft of the new isr task implementing calibpalooza ISR steps order.
The new task is called isrTaskLSST.
Note that there are a lot of things missing: upstream changes to the ISR task, few docstring, unit test, further testing. These will be in follow-up tickets.
Note also that we are focusing on LSSTCam and not trying to have HSC and DECam use this task. Can be done if needed in follow-up tickets.  
Also, I ran into git issues multiple times so I ended up re-commiting the isrTaskLSST file in big chunks in only a few commits.

### DIFF
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -25,43 +25,6 @@ from .masking import MaskingTask
 from .isrStatistics import IsrStatisticsTask
 from .isr import maskNans
 
-
-def crosstalkSourceLookup(datasetType, registry, quantumDataId, collections):
-    """Lookup function to identify crosstalkSource entries.
-
-    This should return an empty list under most circumstances.  Only
-    when inter-chip crosstalk has been identified should this be
-    populated.
-
-    Parameters
-    ----------
-    datasetType : `str`
-        Dataset to lookup.
-    registry : `lsst.daf.butler.Registry`
-        Butler registry to query.
-    quantumDataId : `lsst.daf.butler.ExpandedDataCoordinate`
-        Data id to transform to identify crosstalkSources.  The
-        ``detector`` entry will be stripped.
-    collections : `lsst.daf.butler.CollectionSearch`
-        Collections to search through.
-
-    Returns
-    -------
-    results : `list` [`lsst.daf.butler.DatasetRef`]
-        List of datasets that match the query that will be used as
-        crosstalkSources.
-    """
-    newDataId = quantumDataId.subset(DimensionGraph(registry.dimensions, names=["instrument", "exposure"]))
-    results = set(registry.queryDatasets(datasetType, collections=collections, dataId=newDataId,
-                                         findFirst=True))
-    # In some contexts, calling `.expanded()` to expand all data IDs in the
-    # query results can be a lot faster because it vectorizes lookups.  But in
-    # this case, expandDataId shouldn't need to hit the database at all in the
-    # steady state, because only the detector record is unknown and those are
-    # cached in the registry.
-    return [ref.expanded(registry.expandDataId(ref.dataId, records=newDataId.records)) for ref in results]
-
-
 class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
                              dimensions={"instrument", "exposure", "detector"},
                              defaultTemplates={}):
@@ -120,16 +83,6 @@ class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
         storageClass="CrosstalkCalib",
         dimensions=["instrument", "detector"],
         isCalibration=True,
-    )
-    crosstalkSources = cT.PrerequisiteInput(
-        name="isrOverscanCorrected",
-        doc="Overscan corrected input images.",
-        storageClass="Exposure",
-        dimensions=["instrument", "exposure", "detector"],
-        deferLoad=True,
-        multiple=True,
-        lookupFunction=crosstalkSourceLookup,
-        minimum=0,  # not needed for all instruments, no config to control this
     )
     defects = cT.PrerequisiteInput(
         name='defects',
@@ -1040,8 +993,7 @@ class IsrTaskLSST(pipeBase.PipelineTask):
         if self.config.doCrosstalk:
             # Input units: electrons
             self.log.info("Applying crosstalk correction.")
-            self.crosstalk.run(ccdExposure, crosstalk=crosstalk,
-                               crosstalkSources=crosstalkSources)
+            self.crosstalk.run(ccdExposure, crosstalk=crosstalk)
 
         # Masking block (defects, NAN pixels and trails).
         # Saturated and suspect pixels have already been masked.
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -395,9 +395,10 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         "when brighter-fatter correction is applied.",
         default=0,
     )
-    fwhm = pexConfig.Field(
+    fwhmInterBF = pexConfig.Field(
         dtype=float,
-        doc="FWHM of PSF in arcseconds.",
+        doc="FWHM of PSF in arcseconds used for interpolation in brighter-fatter correction "
+        "(currently unused).",
         default=1.0,
     )
     growSaturationFootprintSize = pexConfig.Field(
@@ -788,7 +789,7 @@ class IsrTaskLSST(pipeBase.PipelineTask):
         with self.flatContext(interpExp, flat, dark):
             isrFunctions.interpolateFromMask(
                 maskedImage=interpExp.getMaskedImage(),
-                fwhm=self.config.fwhm,
+                fwhm=self.config.fwhmInterBF,
                 growSaturatedFootprints=self.config.growSaturationFootprintSize,
                 maskNameList=list(self.config.brighterFatterMaskListToInterpolate)
             )
@@ -1022,7 +1023,7 @@ class IsrTaskLSST(pipeBase.PipelineTask):
             self.log.info("Interpolating masked pixels.")
             isrFunctions.interpolateFromMask(
                 maskedImage=ccdExposure.getMaskedImage(),
-                fwhm=self.config.fwhm,
+                fwhm=self.config.fwhmInterBF,
                 growSaturatedFootprints=self.config.growSaturationFootprintSize,
                 maskNameList=list(self.config.maskListToInterpolate)
             )
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -106,7 +106,6 @@ class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
         doc="Linearity correction calibration.",
         dimensions=["instrument", "detector"],
         isCalibration=True,
-        minimum=0,  # can fall back to cameraGeom
     )
     ptc = cT.PrerequisiteInput(
         name="ptc",
@@ -121,7 +120,6 @@ class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
         storageClass="CrosstalkCalib",
         dimensions=["instrument", "detector"],
         isCalibration=True,
-        minimum=0,  # can fall back to cameraGeom
     )
     crosstalkSources = cT.PrerequisiteInput(
         name="isrOverscanCorrected",
@@ -146,7 +144,6 @@ class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
         storageClass="BrighterFatterKernel",
         dimensions=["instrument", "detector"],
         isCalibration=True,
-        minimum=0,
     )
     dark = cT.PrerequisiteInput(
         name='dark',
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -306,7 +306,7 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
     usePtcGains = pexConfig.Field(
         dtype=bool,
         doc="Use the gain values from the Photon Transfer Curve?",
-        default=True,
+        default=False,
     )
     readNoise = pexConfig.Field(
         dtype=float,
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -151,7 +151,6 @@ class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
         if config.usePtcGains is not True and config.usePtcReadNoise is not True:
             del self.ptc
         if config.doCrosstalk is not True:
-            del self.crosstalkSources
             del self.crosstalk
         if config.doDefect is not True:
             del self.defects
@@ -915,7 +914,7 @@ class IsrTaskLSST(pipeBase.PipelineTask):
 
     def run(self, *, ccdExposure, dnlLUT=None, bias=None, deferredChargeCalib=None, linearizer=None,
             ptc=None, crosstalk=None, defects=None, bfKernel=None, bfGains=None, dark=None,
-            crosstalkSources=None, flat=None, **kwargs
+            flat=None, **kwargs
             ):
 
         detector = ccdExposure.getDetector()
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -723,75 +723,497 @@ class IsrTaskLSST(pipeBase.PipelineTask):
             if ccdExposure.getBBox().contains(amp.getBBox()):
                 self.log.debug("Constructing variance map for amplifer %s.", amp.getName())
                 ampExposure = ccdExposure.Factory(ccdExposure, amp.getBBox())
+                if overscanResults is not None:
+                    self.updateVariance(ampExposure, amp,
+                                        overscanImage=overscanResults.overscanImage,
+                                        ptcDataset=ptc)
+                else:
+                    self.updateVariance(ampExposure, amp,
+                                        overscanImage=None,
+                                        ptcDataset=ptc)
+                if self.config.qa is not None and self.config.qa.saveStats is True:
+                    qaStats = afwMath.makeStatistics(ampExposure.getVariance(),
+                                                     afwMath.MEDIAN | afwMath.STDEVCLIP)
+                    self.log.debug("  Variance stats for amplifer %s: %f +/- %f.",
+                                   amp.getName(), qaStats.getValue(afwMath.MEDIAN),
+                                   qaStats.getValue(afwMath.STDEVCLIP))
+        if self.config.maskNegativeVariance:
+            self.maskNegativeVariance(ccdExposure)
 
-                isrFunctions.updateVariance(
-                        maskedImage=ampExposure.getMaskedImage(),
-                        gain=gain,
-                        readNoise=readNoise,
-                        )
+    def maskDefect(self, exposure, defectBaseList):
+        """Mask defects using mask plane "BAD", in place.
 
-#                if self.config.maskNegativeVariance:
-#                    self.maskNegativeVariance(ccdExposure)
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure to process.
 
-    def run(self,**kwargs):
+        defectBaseList : defect-type
+            List of defects to mask. Can be of type  `lsst.ip.isr.Defects`
+            or `list` of `lsst.afw.image.DefectBase`.
+        """
+        maskedImage = exposure.getMaskedImage()
+        if not isinstance(defectBaseList, Defects):
+            # Promotes DefectBase to Defect
+            defectList = Defects(defectBaseList)
+        else:
+            defectList = defectBaseList
+        defectList.maskPixels(maskedImage, maskName="BAD")
 
+    def maskEdges(self, exposure, numEdgePixels=0, maskPlane="SUSPECT", level='DETECTOR'):
+        """Mask edge pixels with applicable mask plane.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            Exposure to process.
+        numEdgePixels : `int`, optional
+            Number of edge pixels to mask.
+        maskPlane : `str`, optional
+            Mask plane name to use.
+        level : `str`, optional
+            Level at which to mask edges.
+        """
+        maskedImage = exposure.getMaskedImage()
+        maskBitMask = maskedImage.getMask().getPlaneBitMask(maskPlane)
+
+        if numEdgePixels > 0:
+            if level == 'DETECTOR':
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        dimensions=["instrument", "detector"],
+        isCalibration=True,
+    )
+    linearizer = cT.PrerequisiteInput(
+        name='linearizer',
+        storageClass="Linearizer",
+        doc="Linearity correction calibration.",
+        dimensions=["instrument", "detector"],
+        isCalibration=True,
+        minimum=0,  # can fall back to cameraGeom
```
erykoff, line 109: I am not sure what this means; we should either use a linearizer calibration or not.  I think that automated fallbacks can be dangerous and confusing.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        doc="Output ISR processed exposure, with pixels left uninterpolated.",
+        storageClass="ExposureF",
+        dimensions=["instrument", "exposure", "detector"],
+    )
+    outputStatistics = cT.Output(
+        name="isrStatistics",
+        doc="Output of additional statistics table.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "exposure", "detector"],
+    )
```
erykoff, line 175: I think we also want the rebinned exposure output. https://github.com/lsst/ip_isr/blob/ba5f90d8bcf72f7eb35098846ed0dda4999b39e7/python/lsst/ip/isr/isrTask.py#L263-L274

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+            self.prerequisiteInputs.remove("crosstalkSources")
+            self.prerequisiteInputs.remove("crosstalk")
+        if config.doDefect is not True:
+            self.prerequisiteInputs.remove("defects")
+        if config.doBrighterFatter is not True:
+            self.prerequisiteInputs.remove("bfKernel")
+        if config.doDark is not True:
+            self.prerequisiteInputs.remove("dark")
+
+        if config.doWrite is not True:
```
erykoff, line 200: Is `doWrite` actually something we want to support, or is this a hangover from the way tasks were written in the gen2 era?  (I'm pretty sure this is not a config we want).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+                        pipelineConnections=IsrTaskLSSTConnections):
+    """Configuration parameters for IsrTaskLSST.
+
+    Items are grouped in the order in which they are executed by the task.
+    """
+    datasetType = pexConfig.Field(
+        dtype=str,
+        doc="Dataset type for input data; users will typically leave this alone, "
+        "but camera-specific ISR tasks will override it.",
+        default="raw",
```
erykoff, line 221: I'm pretty sure this is a hangover from trying to support certain types of pre-processed decam images from gen2.  We do not want this configuration at all.  If, strangely, you need a different dataset type then the `ccdExposure` connection name would be the thing to override).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        dtype=bool,
+        doc="Apply bias frame correction?",
+        default=True,
+    )
+
+    # Deferred charge correction.
+    doDeferredCharge = pexConfig.Field(
+        dtype=bool,
+        doc="Apply deferred charge correction?",
+        default=False,
```
erykoff, line 284: Shouldn't this be default True (this is where we want to be, even if we have to override it now).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    )
+    gain = pexConfig.Field(
+        dtype=float,
+        doc="The gain to use if no Detector is present in the Exposure (ignored if NaN).",
+        default=float("NaN"),
+    )
+    usePtcGains = pexConfig.Field(
+        dtype=bool,
+        doc="Use the gain values from the Photon Transfer Curve?",
+        default=False,
```
erykoff, line 319: This should default to True, this is how we want the default run of the task to go.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    )
+    doEmpiricalReadNoise = pexConfig.Field(
+        dtype=bool,
+        doc="Calculate empirical read noise instead of value from AmpInfo data?",
+        default=False,
+    )
+    usePtcReadNoise = pexConfig.Field(
+        dtype=bool,
+        doc="Use readnoise values from the Photon Transfer Curve?",
+        default=False,
```
erykoff, line 334: This default should be True.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        default=float("NaN"),
+    )
+    usePtcGains = pexConfig.Field(
+        dtype=bool,
+        doc="Use the gain values from the Photon Transfer Curve?",
+        default=False,
+    )
+    readNoise = pexConfig.Field(
+        dtype=float,
+        doc="The read noise to use if no Detector is present in the Exposure.",
```
erykoff, line 323: I don't understand what this is for.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        dtype=str,
+        doc="Name of mask plane to use for suspect pixels.",
+        default="SUSPECT",
+    )
+
+    # Crosstalk.
+    doCrosstalk = pexConfig.Field(
+        dtype=bool,
+        doc="Apply intra-CCD crosstalk correction?",
+        default=False,
```
erykoff, line 363: This should default to True.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    )
+    doCrosstalkBeforeAssemble = pexConfig.Field(
+        dtype=bool,
+        doc="Apply crosstalk correction before CCD assembly?",
+        default=False,
```
erykoff, line 368: What is the correct way of doing this for LSST calibpalooza?  There shouldn't be an option for the other way.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    )
+    doWidenSaturationTrails = pexConfig.Field(
+        dtype=bool,
+        doc="Widen bleed trails based on their width.",
+        default=True,
+    )
+    doCameraSpecificMasking = pexConfig.Field(
+        dtype=bool,
+        doc="Mask camera-specific bad regions?",
+        default=False,
```
erykoff, line 394: What does this config do?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        dtype=bool,
+        doc="Should we set the level of all BAD patches of the chip to the chip's average value?",
+        default=True,
+    )
+
+    # Brighter-Fatter correction.
+    doBrighterFatter = pexConfig.Field(
+        dtype=bool,
+        doc="Apply the brighter-fatter correction?",
+        default=False,
```
erykoff, line 429: This should be True.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    brighterFatterMaskGrowSize = pexConfig.Field(
+        dtype=int,
+        doc="Number of pixels to grow the masks listed in config.brighterFatterMaskListToInterpolate "
+        "when brighter-fatter correction is applied.",
+        default=0,
+    )
+    fwhm = pexConfig.Field(
+        dtype=float,
+        doc="FWHM of PSF in arcseconds.",
+        default=1.0,
```
erykoff, line 472: What is this used for?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    doStandardStatistics = pexConfig.Field(
+        dtype=bool,
+        doc="Should standard image quality statistics be calculated?",
+        default=True,
+    )
+    # Calculate additional statistics?
+    doCalculateStatistics = pexConfig.Field(
+        dtype=bool,
+        doc="Should additional ISR statistics be calculated?",
+        default=False,
```
erykoff, line 518: Should this really be False?  Do they take too long or something?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        lsst.ip.isr.isrFunctions.updateVariance
+        """
+        maskPlanes = [self.config.saturatedMaskName, self.config.suspectMaskName]
+        if self.config.usePtcGains:
+            if ptcDataset is None:
+                raise RuntimeError("No ptcDataset provided to use PTC gains.")
+            else:
+                gain = ptcDataset.gain[amp.getName()]
+                self.log.info("Using gain from Photon Transfer Curve.")
+        else:
```
erykoff, line 653: This should maybe be a warning that you are using a questionable gain value.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+                                               afwMath.STDEVCLIP, stats).getValue()
+            self.log.info("Calculated empirical read noise for amp %s: %f.",
+                          amp.getName(), readNoise)
+        elif self.config.usePtcReadNoise:
+            if ptcDataset is None:
+                raise RuntimeError("No ptcDataset provided to use PTC readnoise.")
+            else:
+                readNoise = ptcDataset.noise[amp.getName()]
+            self.log.info("Using read noise from Photon Transfer Curve.")
+        else:
```
erykoff, line 691: This may also be a warning since we really only want to use certified read noises.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+
+        Returns
+        -------
+        doLinearize : `Bool`
+            If True, linearization should be performed.
+        """
+        return self.config.doLinearize and \
+            detector.getAmplifiers()[0].getLinearityType() != NullLinearityType
+
+    def run(self, ccdExposure, dnlLUT=None, bias=None, deferredChargeCalib=None, linearizer=None,
```
erykoff, line 1044: This should all be keywords (this is not quite in the dev guide yet, but it's the way we're going); if you don't have a default you can do `def run(self, *, ccdExposure, dnlLUT=None ...)`.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+            ptc=None, crosstalk=None, defects=None, bfKernel=None, bfGains=None, dark=None,
+            crosstalkSources=None, flat=None, **kwargs
+            ):
+
+        # TODO rename ccd to detector?
```
erykoff, line 1049: Yes please!

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+from .isrStatistics import IsrStatisticsTask
+from .isr import maskNans
+
+
+def crosstalkSourceLookup(datasetType, registry, quantumDataId, collections):
+    """Lookup function to identify crosstalkSource entries.
+
+    This should return an empty list under most circumstances.  Only
+    when inter-chip crosstalk has been identified should this be
+    populated.
```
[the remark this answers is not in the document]
aferte: this is copied from the original task so I won't address it in this ticket.

python/lsst/ip/isr/isrTaskLSST.py
aferte: this is copied from the original task so I won't address it in this ticket.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        dimensions=["instrument", "detector"],
+        isCalibration=True,
+    )
+    crosstalk = cT.PrerequisiteInput(
+        name="crosstalk",
+        doc="Input crosstalk object",
+        storageClass="CrosstalkCalib",
+        dimensions=["instrument", "detector"],
+        isCalibration=True,
+        minimum=0,  # can fall back to cameraGeom
```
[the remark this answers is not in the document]
aferte: this is copied from the original task so I won't address it in this ticket.

python/lsst/ip/isr/isrTaskLSST.py
aferte: This was updated in the old isr task while I was working on this new one - is it ok if I keep this for DM-40903?

python/lsst/ip/isr/isrTaskLSST.py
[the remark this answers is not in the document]
erykoff: I believe this first draft should only copy the parts that are needed in the new task.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        doc="Output of additional statistics table.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "exposure", "detector"],
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+
+        if config.doDiffNonLinearCorrection is not True:
+            self.prerequisiteInputs.remove("dnlLUT")
```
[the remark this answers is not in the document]
aferte: This was updated in the old isr task (I think?) while I was working on this new one - is it ok if I keep this for DM-40903?

python/lsst/ip/isr/isrTaskLSST.py
aferte: and usePtcGains set to False - right?

python/lsst/ip/isr/isrTaskLSST.py
aferte: Cross talk is done after assembly so happy to remove this.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: I think you can just remove `minimum=0`.

python/lsst/ip/isr/isrTaskLSST.py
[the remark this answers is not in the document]
erykoff: Same here; remove the `minimum=0`.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        dimensions=["instrument", "detector"],
+        isCalibration=True,
+    )
+    bfKernel = cT.PrerequisiteInput(
+        name='brighterFatterKernel',
+        doc="Newer complete kernel + gain solutions.",
+        storageClass="BrighterFatterKernel",
+        dimensions=["instrument", "detector"],
+        isCalibration=True,
+        minimum=0,  # can use either bfKernel or newBFKernel
```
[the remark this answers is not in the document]
erykoff: Also here, remove `minimum=0`.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: If you can copy over the binned exposure stuff that would be great.  But it can also be punted.

python/lsst/ip/isr/isrTaskLSST.py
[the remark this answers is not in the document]
erykoff: Sure, but it's easy to just change all these to `del blah` now, isn't it?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        if config.doBias is not True:
+            self.prerequisiteInputs.remove("bias")
+        if config.doDeferredCharge is not True:
+            self.prerequisiteInputs.remove("deferredChargeCalib")
+        if config.doLinearize is not True:
+            self.prerequisiteInputs.remove("linearizer")
+        if config.usePtcGains is not True and config.usePtcReadNoise is not True:
+            self.prerequisiteInputs.remove("ptc")
+        if config.doCrosstalk is not True:
+            self.prerequisiteInputs.remove("crosstalkSources")
```
erykoff, line 191: We don't need `crosstalkSources` at all (here or if they're defined above)

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        target=AssembleCcdTask,
+        doc="CCD assembly task.",
+    )
+
+    # Combine snaps.
+    doSnapCombine = pexConfig.Field(
+        dtype=bool,
+        doc="Combine snaps?",
+        default=False,
+    )
```
erykoff, line 271: I think actually remove this; I think it's not going to be part of the isrTask if we have to do it, but a separate task.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: I think this should be removed.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: usePtcGains should default to True, that's where we want to get our gains by default.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    readNoise = pexConfig.Field(
+        dtype=float,
+        doc="The read noise to use if no Detector is present in the Exposure.",
+        default=0.0,
+    )
+    doEmpiricalReadNoise = pexConfig.Field(
+        dtype=bool,
+        doc="Calculate empirical read noise instead of value from AmpInfo data?",
+        default=False,
+    )
```
erykoff, line 330: This should probably be removed.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+    doDark = pexConfig.Field(
+        dtype=bool,
+        doc="Apply dark frame correction.",
+        default=True,
+    )
+
+    # Post-ISR operations to make fluence image.
+    doFluence = pexConfig.Field(
+        dtype=bool,
+        doc="Apply post-ISR operations?",
```
erykoff, line 496: I think this docstring is wrong.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        default=True,
+    )
+
+    # Write the outputs to disk. If ISR is run as a subtask, this may not
+    # be needed.
+    doWrite = pexConfig.Field(
```
erykoff, line 502: Remove this for now.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -0,0 +1,1219 @@
...
+        isrFunctions.darkCorrection(
+            maskedImage=exposure.getMaskedImage(),
+            darkMaskedImage=darkExposure.getMaskedImage(),
+            expScale=expScale,
+            darkScale=darkScale,
+            invert=invert,
+        )
+
+    def applyPostIsr(self, **kwargs):
+        # TODO add post-ISR (4th column of Eli's calibration boxes)
```
erykoff, line 997: Oh, this should not be in ISR, this would be a different task _after_ ISR.

python/lsst/ip/isr/isrTaskLSST.py
aferte: right, got confused - thanks.

python/lsst/ip/isr/isrTaskLSST.py
aferte: It is used in the interpolation of masked pixels in the brighter-fatter correction. This is also from the original task, there are probably historical reasons to use this.

python/lsst/ip/isr/isrTaskLSST.py
aferte: Here as well, there are probably historical reasons for needing to do camera-specific masking, happy to get rid of it if we are doing such masking in another way.

erykoff: I think we should remove this from the LSST task, it's not relevant.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: I think we can/should change the name then to something that specifies that it's part of the BFE correction.

aferte:
All comments are addressed and jenkins run was successful. We have a dedicated ticket to develop tests.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: This should have a name consistent with the others, like `brighterFatterFwhm` or `brighterFatterFwhmForInterpolation`.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: Sorry, the warning should be below for `amp.getGain()` and there should _not_ be a warning for the usual gain from the PTC.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: I don't think this needs to be an info, it should be debug since it's the only option.  Also if it isn't supported for other read noise below why do we have the config?

erykoff (approved):
There are a couple more things that I want to make sure are cleaned up, but looks good and I hope it works when we start testing it!

python/lsst/ip/isr/isrTaskLSST.py
aferte: much better - thanks!

python/lsst/ip/isr/isrTaskLSST.py
aferte: Agreed.... equally usePtcGains is also useless? I think I will get rid of them for now.

plazas: I'm also currently on DM-37554 which will have the gain and noise defined with a function in a single point, although this is being done in `isrTask.py`. Just something to keep in mind, especially potentially going out of synch as I see this is a new file (`isrTaskLSST.py`).

erykoff: Thanks for pointing this out, as we move to the next step we'll want to keep it in sync.  But the hope with this new task is that there will be much less confusion because there will be many fewer options.  Though we probably need some way of filling a gain just for initial processing to bootstrap the PTC.  But I'd like to start with the "correct" flow and add from there, rather than holding onto legacy.

PR #283, opened 2023-11-09
DM-41663: DM-37554 breaks ci_imsim

plazas:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1826,7 +1826,7 @@ def defineEffectivePtc(self, ptcDataset, detector, bfGains, overScans, metadata)
 
             # Check if the gain up to this point differs from the
             # gain in bfGains. If so, raise or warn, accordingly.
-            if not boolGainMismatch and bfGains is not None:
+            if not boolGainMismatch and bfGains is not None and len(bfGains) != 0:
```
erykoff, line 1829: Is `bfGains` guaranteed to be `None` or something that has a `__len__()`?

erykoff: I guess below we are assuming it does have indexing, but I want to make sure we know all the options.

plazas: Do you have any other options in mind?

erykoff: Well, what formats can it be?  If it's None or a dict, then actually should be `bfGains is not None and ampName in bfGains`

erykoff: But if there's another possibility then we'd need to check that.  But it may be that None/dict really are the only options (and this seems to be the case given ci_cpp and ci_imsim)

plazas: Yeah, I can't think of other options, so that's why I was asking. @czwa ?

erykoff: At least it should be checking the ampName and not the length.

erykoff (approved):

PR #284, opened 2023-11-16
DM-34340: Adapt to daf_butler API changes.

TallJimbo:


python/lsst/ip/isr/isrTask.py
```diff
@@ -84,15 +83,16 @@ def crosstalkSourceLookup(datasetType, registry, quantumDataId, collections):
         List of datasets that match the query that will be used as
         crosstalkSources.
     """
-    newDataId = quantumDataId.subset(DimensionGraph(registry.dimensions, names=["instrument", "exposure"]))
+    newDataId = quantumDataId.subset(registry.dimensions.conform(["instrument", "exposure"]))
```
timj, line 86: I can't comment on the line directly but I think the docstring for `quantumDataId` above might no longer want to say `ExpandedDataCoordinate`.

PR #290, opened 2023-12-08
DM-41316: Refactor overscan task for IsrTaskLSST into serial/parallel tasks.

erykoff:
This also adds crosstalk correction in the parallel overscan region.

### DIFF
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -210,7 +210,7 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         doc="Do parallel overscan subtraction?",
         default=True,
     )
-    serialOverscan = pexConfig.ConfigurableField(
+    parallelOverscan = pexConfig.ConfigurableField(
         target=ParallelOverscanCorrectionTask,
         doc="Parallel overscan subtraction task for image segments.",
     )
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -467,6 +467,12 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
             #      self.isrStats.doApplyGainsForCtiStatistics:
             raise ValueError("doApplyGains must match isrStats.applyGainForCtiStatistics.")
 
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.serialOverscan.fitType = "MEDIAN_PER_ROW"
+        self.parallelOverscan.fitType = "MEDIAN_PER_ROW"
+
 
 class IsrTaskLSST(pipeBase.PipelineTask):
     ConfigClass = IsrTaskLSSTConfig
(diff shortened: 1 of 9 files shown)


python/lsst/ip/isr/overscan.py
```diff
@@ -974,3 +873,473 @@ def debugView(self, image, model, amp=None, isTransposed=True):
...
+
+        # Do Parallel Overscan
+        parallelResults = None
+        if self.config.doParallelOverscan:
+            # This does not need any extensions, as we'll only
+            # subtract it from the data region.
+            parallelOverscanBBox = amp.getRawParallelOverscanBBox()
+            imageBBox = amp.getRawDataBBox()
+
+            maskIm = exposure.getMaskedImage()
```
czwa, line 1004: I think this `maskIm` can be dropped here and below.  It's almost certainly a leftover from a previous attempt at fixing this.

python/lsst/ip/isr/overscan.py
```diff
@@ -974,3 +873,473 @@ def debugView(self, image, model, amp=None, isTransposed=True):
...
+        ------
+        RuntimeError
+            Raised if an invalid overscan type is set.
+        """
+        # This does not need any extending, as we only subtract
+        # from the data region.
+        parallelOverscanBBox = amp.getRawParallelOverscanBBox()
+        imageBBox = amp.getRawDataBBox()
+
+        maskIm = exposure.getMaskedImage()
```
czwa, line 1274: Also here.

python/lsst/ip/isr/overscan.py
```diff
@@ -974,3 +873,473 @@ def debugView(self, image, model, amp=None, isTransposed=True):
...
+                               overscanMean=overscanMean,
+                               overscanMedian=overscanMedian,
+                               overscanSigma=overscanSigma,
+                               residualMean=residualMean,
+                               residualMedian=residualMedian,
+                               residualSigma=residualSigma)
+
+
+class SerialOverscanCorrectionTaskConfig(OverscanCorrectionTaskConfigBase):
+    leadingToSkip = pexConfig.Field(
```
czwa, line 1051: Can these be moved to the base config?  I see they're both here and in the associated parallel correction config.

czwa (approved):
A few comments.  The only potential issue is the handling of crosstalk units in the future.

python/lsst/ip/isr/overscan.py
erykoff: These are new names (notice it doesn't specify `rows` or `columns` which are subjective), and I don't want to deprecate the old names for no obvious gain.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1014,7 +1268,7 @@ def run(self, *, ccdExposure, dnlLUT=None, bias=None, deferredChargeCalib=None,
 
         if self.config.doVariance:
             # Input units: electrons
-            self.variancePlane(ccdExposure, detector, overscans, ptc)
+            self.variancePlane(ccdExposure, detector, serialOverscans, ptc)
 
         if self.config.doCrosstalk:
             # Input units: electrons
```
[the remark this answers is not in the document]
erykoff: The order here is not correct according to our latest calibration boxes, that's correct.  I don't think it's good for this ticket to reorder things (that's out of scope) but we should make sure that the ordering is fixed prior to/with the testing updates.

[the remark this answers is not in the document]
erykoff: DM-42459 is the ticket.

PR #291, opened 2023-12-09
DM-41681: Implement function from P. Astier's code to subtract covariances at large lags in the PTC code of the LSST Science Pipelines.

plazas:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -53,11 +53,15 @@ class PhotonTransferCurveDataset(IsrCalib):
...
+    ptcFitType : `str`, optional
         Type of model fitted to the PTC: "POLYNOMIAL", "EXPAPPROXIMATION",
         or "FULLCOVARIANCE".
-    covMatrixSide : `int`
-        Maximum lag of covariances (size of square covariance matrices).
+    covMatrixSide : `int`, optional
+        Maximum lag of measured covariances (size of square covariance
+        matrices).
+    covMatrixSideFullCovFit : `int, optional
+        Maximum covariances lag for FULLCOVARIANCE fit. It shoudl be less or
```
erykoff, line 63: Typo for should

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -171,16 +175,19 @@ class PhotonTransferCurveDataset(IsrCalib):
...
-    _VERSION = 1.4
+    _VERSION = 1.5
 
-    def __init__(self, ampNames=[], ptcFitType=None, covMatrixSide=1, **kwargs):
+    def __init__(self, ampNames=[], ptcFitType=None, covMatrixSide=1,
+                 covMatrixSideFullCovFit=1, **kwargs):
         self.ptcFitType = ptcFitType
         self.ampNames = ampNames
         self.covMatrixSide = covMatrixSide
+        self.covMatrixSideFullCovFit = covMatrixSideFullCovFit
```
erykoff, line 190: For maximum compatibility, could this default to `None` and use the value of `covMatrixSide` if it's none?  (It may not matter in usage).

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -815,3 +835,12 @@ def validateGainNoiseTurnoffValues(self, ampName, doWarn=False):
...
         self.ptcTurnoff[ampName] = ptcTurnoff
+
+    def _validateCovarianceMatrizSizes(self):
+        """Ensure  covMatrixSideFullCovFit <= covMatrixSide."""
+        if self.covMatrixSideFullCovFit > self.covMatrixSide:
+            self.log.warning("covMatrixSideFullCovFit > self.covMatrixSide "
+                             f"({self.covMatrixSideFullCovFit} > {self.covMatrixSide})."
+                             "Setting the former to the latter.")
+
+        self.covMatrixSideFullCovFit = self.covMatrixSide
```
erykoff, line 846: This is at the wrong indentation level.

tests/test_ptcDataset.py
```diff
@@ -202,11 +204,13 @@ def test_ptcDatset(self):
         # Fill the dataset with made up data.
         nSignalPoints = 5
         nSideCovMatrix = 2
+        nSideCovMatrixFullCovFit = 2
```
erykoff, line 207: We need to test where `nSideCovMatrixFullCovFit` is less than `nSideCovMatrix`.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -626,6 +641,10 @@ def fromTable(cls, tableList):
             else:
                 inDict['noiseMatrix'][ampName] = record['NOISE_MATRIX']
                 inDict['noiseMatrixNoB'][ampName] = record['NOISE_MATRIX_NO_B']
+            if calibVersion < 1.5:
+                inDict['covMatrixSideFullCovFit'] = inDict['covMatrixSide']
+            else:
+                inDict['covMatrixSideFullCovFit'] = record['COV_MATRIX_SIDE_FULL_COV_FIT']
```
erykoff, line 647: Okay, I see why I originally thought this wouldn't work.  This is matched to that `COV_MATRIX_SIDE` above, and actually will get overwritten through the loop of amps (but should be the same).  Maybe worth adding a comment here about this.

erykoff (commented):
The tests need a bit more coverage.  We need to test what happens when there is a mismatch between the size and fit size.  Both when it's okay (fit size < size; fit size == size) and when bad (fit size > size, should log a warning).  In fact, maybe that should raise instead of warning.

czwa (approved):

PR #293, opened 2024-01-12
DM-42459: ISR boxes version 2.0

aferte:
Update order of ISR steps in the LSST ISR task.

### DIFF
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -1236,7 +1236,7 @@ class IsrTaskLSST(pipeBase.PipelineTask):
             )
 
         if self.config.doAssembleCcd:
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1236,44 +1236,44 @@ def run(self, *, ccdExposure, dnlLUT=None, bias=None, deferredChargeCalib=None,
             )
 
         if self.config.doAssembleCcd:
-            # Input units: ADU
+            # Input units: electrons
```
czwa, line 1239: With this here, this needs to state that we're in ADU still.

aferte: thanks - updated

erykoff (approved):

PR #295, opened 2024-01-17
DM-42238: Propagate flat pair read noises to ptcDataset

Alex-Broughton:


youtsumi (approved):

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -176,11 +179,12 @@ class PhotonTransferCurveDataset(IsrCalib):
     Version 1.3 adds the `noiseMatrix` and `noiseMatrixNoB` attributes.
     Version 1.4 adds the `auxValues` attribute.
     Version 1.5 adds the `covMatrixSideFullCovFit` attribute.
+    Version 1.6 adds the `noiseList` attribute.
     """
 
     _OBSTYPE = 'PTC'
     _SCHEMA = 'Gen3 Photon Transfer Curve'
-    _VERSION = 1.5
+    _VERSION = 1.6
```
czwa, line 187: This may be 1.7 now.  I know there's another ticket in flight that has changed the version to 1.6: DM-42163.

czwa (changes requested):
I think DM-42163 will merge first, so this should increment the version again.

czwa:
General comment applicable here and elsewhere: PR names should start with the ticket name, so they're easier to track: https://developer.lsst.io/work/flow.html#make-a-pull-request

czwa (approved):
Can you add an entry to the unit tests to check existence and type?  This line is the similar change on the previous ticket, and as `ci_cpp`'s testing is intentionally broken for the sprint, this will at least provide something for Jenkins to check:
https://github.com/lsst/ip_isr/blob/main/tests/test_ptcDataset.py#L90
Approving this on condition a matching line is added for `noiseList`.

PR #298, opened 2024-01-25
DM-42666: Propagate MJD

arunkannawadi:


python/lsst/ip/isr/isrStatistics.py
```diff
@@ -215,11 +215,13 @@ def run(self, inputExp, ptc=None, overscanResults=None, **kwargs):
         if self.config.doCopyCalibDistributionStatistics:
             calibDistributionResults = self.copyCalibDistributionStatistics(inputExp, **kwargs)
 
+        mjd = inputExp.getMetadata().get("MJD", None)
```
arunkannawadi, line 218: I'm not entirely sure if I can expect `MJD` to always exist. If not, can the code downstream handle `None`?

czwa: It looks like `None` is cast to 0 by matplotlib, so that's probably safe.

czwa (approved):

PR #305, opened 2024-02-01
DM-42708: Avoid printing debug logs directly to stdout

kfindeisen:
This PR fixes a log that was printed using `print` instead of `debug` (added on #116). It is the only non-UI `print` statement in the package code.

### DIFF
--- a/python/lsst/ip/isr/fringe.py
+++ b/python/lsst/ip/isr/fringe.py
@@ -115,7 +115,8 @@ class FringeTask(Task):
         if expId is None:
             seed = self.config.stats.rngSeedOffset
         else:
-            print(f"{self.config.stats.rngSeedOffset} {expId}")
+            self.log.debug("Seeding with offset %d and ccdExposureId %d.",
+                           self.config.stats.rngSeedOffset, expId)
             seed = self.config.stats.rngSeedOffset + expId
 
         # Seed for numpy.random.RandomState must be convertable to a 32 bit


python/lsst/ip/isr/fringe.py
```diff
@@ -115,7 +115,7 @@ def loadFringes(self, fringeExp, expId=None, assembler=None):
         if expId is None:
             seed = self.config.stats.rngSeedOffset
         else:
-            print(f"{self.config.stats.rngSeedOffset} {expId}")
+            self.log.debug("Seeding with offset %d and expId %d.", self.config.stats.rngSeedOffset, expId)
```
parejkoj, line 118: One minor suggestion to help parsing the log output.

```suggestion
            self.log.debug("Seeding with offset %d and exposureId %d.", self.config.stats.rngSeedOffset, expId)
```

parejkoj (approved):
Good fix, and thanks for checking the rest of the package for errant `print`s: I was going to ask about that!

python/lsst/ip/isr/fringe.py
parejkoj: Oh, yes, definitely!

parejkoj: Ugh... I've just bounced back and forth on this twice. I think `ccdExposureId` is correct, since that's what it comes from, and ISR runs per-detector, and we should be explicit in the logs.

PR #308, opened 2024-02-06
DM-42795: Fix incorrect NOISE_LIST access.

czwa:
Remove incorrect NOISE_LIST access.

As this is a newly added field, it should only be set in a version controlled block.  This ticket removes the extra assignment.


fred3m (approved):
LGTM

PR #309, opened 2024-02-12
DM-42863: Add method to ptcDataset that evaluates PTC model

Alex-Broughton:


erykoff (commented):
This looks fine, but needs tests.

erykoff (approved):

PR #315, opened 2024-03-13
DM-42868 : creation of new ISR mocks for the new ISR task for LSSTCam

aferte:
Jenkins ran successfully.


python/lsst/ip/isr/isrMock.py
```diff
@@ -556,14 +557,60 @@ def getExposure(self):
...
+            'UL': ReadoutCorner.UL,
+        }
         for amp in ccd:
             newAmp = amp.rebuild()
             newAmp.setLinearityCoeffs((0., 1., 0., 0.))
             newAmp.setLinearityType("Polynomial")
             newAmp.setGain(self.config.gain)
             newAmp.setSuspectLevel(25000.0)
             newAmp.setSaturation(32000.0)
+            readcorner = 'LL'
```
erykoff, line 573: Maybe `readoutCorner`?

erykoff: But does this need to be set here?  It's overridden below.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+            Simulated ISR image data.
+
+        Notes
+        -----
+        This method constructs a "raw" data image.
+        """
+        exposure = self.getExposure()
+
+        for idx, amp in enumerate(exposure.getDetector()):
+
```
erykoff, line 132: I would suggest adding some comments here about what we're doing... adding in effects "forward" as they come into the telescope, and then ISR will reverse these.  That's why we start in electrons and end in ADU.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+                                               * self.config.darkTime))
+
+            # 2. Gain normalize  (from e- to ADU)
+            # TODO: DM-??? gain from PTC per amplifier
+            # TODO: DM-??? gain with temperature dependence
+            if self.config.doAddGain:
+                self.addGain(ampData, self.config.gain)
+
+            # 3. Add bias frame to the image region in ADU
+            if self.config.doAddBias:
```
erykoff, line 177: This the pre-existing config name from @czwa, but I think it really is adding noise associated with a bias frame, and not a bias frame, and should be commented accordingly.  (Assuming Chris agrees).

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+                self.amplifierAddNoise(ampData,
+                                       0., self.config.readNoise / self.config.gain)
+
+        # 4. Apply cross-talk in ADU
+        if self.config.doAddCrosstalk:
```
erykoff, line 182: I have more thoughts on crosstalk, will discuss not on this PR.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+            ctCalib = CrosstalkCalib()
+            for idxS, ampS in enumerate(exposure.getDetector()):
+                for idxT, ampT in enumerate(exposure.getDetector()):
+                    ampDataT = exposure.image[ampT.getRawDataBBox().shiftedBy(ampT.getRawXYOffset())]
+                    outAmp = ctCalib.extractAmp(exposure.getImage(), ampS, ampT,
+                                                isTrimmed=False)
+                    self.amplifierAddCT(outAmp, ampDataT, self.crosstalkCoeffs[idxS][idxT])
+
```
erykoff, line 190: Add comment here that we're now adding overscan and bias level.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+        """
+        if "_filename" in dataType:
+            self.expectData()
+            return tempfile.mktemp(), "mock"
+        elif 'transmission_' in dataType:
+            self.expectData()
+            return TransmissionMockLSST(config=self.config).run()
+        elif dataType == 'ccdExposureId':
+            self.expectData()
+            return 20090913
```
erykoff, line 499: Can we put a date in the middle of the survey?  (It doesn't matter but would be nice)

erykoff: (And for all the dates in this file)

erykoff (commented):
This looks fine, and I'll want @czwa to also take a look.  Have you checked that this is suitable to use as a test mock for the LSST ISR task?  Though it may be okay to make updates to this when standing up those tests as more possible issues are found, I don't think that should be split out into yet another ticket.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+                self.amplifierAddYGradient(ampData, -1.0 * self.config.overscanScale,
+                                           1.0 * self.config.overscanScale)
+                self.amplifierAddYGradient(parallelOscanData, -1.0 * self.config.overscanScale,
+                                           1.0 * self.config.overscanScale)
+
+        # 6. Add Parallel overscan xtalk.
+        # TODO: DM-43286
+
+            if self.config.doAddSerialOverscan:
+                # 1. We grow the image to the parallel oversan region
```
czwa, line 219: typo

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+                serialOscanBBox = geom.Box2I(
+                    geom.Point2I(serialOscanBBox.getMinX(),
+                                 grownImageBBox.getMinY()),
+                    geom.Extent2I(serialOscanBBox.getWidth(),
+                                  grownImageBBox.getHeight()),
+                )
+                serialOscanData = exposure.image[serialOscanBBox]
+
+                # 7. Add bias level to the whole region of amplifiers in ADU
+                self.amplifierAddNoise(ampData, self.config.biasLevel,
```
czwa, line 234: This (combined with the above call with zero mean) has double-added the read noise, at least in the imaging region.  The bias signal contributes to the image by adding:

- A DC offset (`self.config.biasLevel`) which has a known dispersion (`self.config.readNoise`)
- Two 1-D contributions (the serial and parallel overscan signals)
- A final 2-D residual pattern (what the combined bias will correct).

The `doAddBias` step above should now only be adding this third residual pattern, as the read noise and bias level additions have been moved here to ensure that the two overscan regions are consistent.  If we're not adding a residual pattern, that step should add 0.0 there.

I was confused on units at this point, but `self.config.readNoise` claims to be in electrons, the comments here indicate we're working in ADU, so the factor of `1 / self.config.gain` is correct, which I'm stating here in case I get confused again in the future.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+        if self.config.doGenerateAmpDict:
+            expDict = dict()
+            for amp in exposure.getDetector():
+                expDict[amp.getName()] = exposure
+            return expDict
+        else:
+            return exposure
+
+    def addGain(self, ampData, gain):
+        ampArr = ampData.array
```
czwa, line 261: Docstring here, indicating that despite the name, this is a multiplicative effect (I chose bad names when I wrote the original class).

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+
+        self.config.doAddFlat = True
+
+
+class CalibratedRawMockLSST(RawMockLSST):
+    """Generate a trimmed raw exposure.
+    """
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+#        self.config.isTrimmed = True
```
czwa, line 315: If this is not trimmed, then the overscan regions should be filled with read noise.  I also think that `doAddFringe` should be `False`, as the calibration should have corrected that.  I'm happy to be wrong here.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+        self.config.biasLevel = 0.0
+        self.config.readNoise = 10.0
+
+
+class ReferenceMockLSST(IsrMockLSST):
+    """Parent class for those that make reference calibrations.
+    """
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.config.isTrimmed = False
```
czwa, line 340: Similar question here:  if these are to be the reference calibrations, I think those need to be trimmed (we don't retain the overscan after combination).  I think there also needs to be a factor applied to the readNoise for this class of mocks:  a combined bias is made by adding multiple calibrated bias frames, so the readNoise should drop by a factor of sqrt(N).  I'm not sure if we want to make N a config option, but having this parent class set `self.config.readNoise *= 0.25  # Assume combined calibrations are made with 16 inputs.` would be good.  I doubt this is done right in the original task.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+
+
+class BiasMockLSST(ReferenceMockLSST):
+    """Simulated combined bias calibration.
+    """
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.config.doAddBias = True
+        self.config.doAddGain = True
+        self.config.readNoise = 10.0
```
czwa, line 374: See above comment.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+        self.config.doDefects = False
+        self.config.doCrosstalkCoeffs = False
+        self.config.doTransmissionCurve = True
+
+
+class MockLSSTDataContainer(object):
+    """Container for holding ISR LSST mock objects.
+    """
+    dataId = "isrMockLSST Fake Data"
+    darkval = 2.  # e-/sec
```
czwa, line 454: These should derive from the config options, but I also don't recall how they're used, so it's probably fine for now.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+        elif dataType == 'crosstalkSources':
+            return None
+        else:
+            raise RuntimeError("ISR DataRefMock cannot return %s.", dataType)
+
+
+class MockFringeLSSTContainer(object):
+    """Container for mock fringe data.
+    """
+    dataId = "isrMockLSST Fake Data"
```
czwa, line 535: Same here, but this entire object is probably not useful until fringing is working better.

czwa (approved):
I think I'm mostly happy.  I have a few questions on trim states for Calibrated and Reference Mocks, and there's an error in applying the read noise, but nothing critical.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,597 @@
...
+
+            if self.config.doAddSource:
+                for sourceAmp, sourceFlux, sourceX, sourceY in zip(self.config.sourceAmp,
+                                                                   self.config.sourceFlux,
+                                                                   self.config.sourceX,
+                                                                   self.config.sourceY):
+                    if idx == sourceAmp:
+                        self.amplifierAddSource(ampData, sourceFlux, sourceX, sourceY)
+
+            # Post ISR effects in e-
```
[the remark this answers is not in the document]
aferte: I used that term because they are effects corrected in the post-ISR image but I can think of something else.

python/lsst/ip/isr/isrMockLSST.py
aferte: Done and changed the name to applyGain.

python/lsst/ip/isr/isrMockLSST.py
aferte: I am removing the containers as they are out of date and I don't see their use at the moment.

python/lsst/ip/isr/isrMockLSST.py
aferte: I am now adding the bias frame (level and scatter) at this stage when no serial overscan is applied, and doing that step in the serial overscan if is applied.

python/lsst/ip/isr/isrMock.py
aferte: I confirm `isLsstLike = False` gives the same on the branch as on main. I don't see why we need to set `rawFlipX,Y` to False but happy to do so. 

isrMock_TrimmedRawMock_branch
isrMock_TrimmedRawMock_main

python/lsst/ip/isr/isrMockLSST.py
aferte: I would agree but the fringe correction is done after the ISR corrections so I am leaving them here. Happy to change if needed later on.

python/lsst/ip/isr/isrMockLSST.py
aferte: I am removing the containers as they are out of date and I don't see their use at the moment.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,482 @@
...
+                                       np.sqrt(self.config.darkRate * self.config.darkTime))
+
+            # 2. Gain normalize  (from e- to ADU)
+            # TODO: DM-??? gain from PTC per amplifier
+            # TODO: DM-??? gain with temperature dependence
+            if self.config.doApplyGain:
+                self.applyGain(ampData, self.config.gain)
+
+            # 3. We either add the bias frame or only the read noise
+            # to the image region in ADU.
```
erykoff, line 181: I don't think this comment is correct.  The "bias frame" as we use it as a calibration product is fixed pattern noise, with a mean of ~0.  So I think the adding of noise below is fine, but this is not a bias frame.

aferte: fixed.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -0,0 +1,482 @@
...
+
+        Parameters
+        ----------
+        ampData : `lsst.afw.image.ImageF`
+            Amplifier image to operate on.
+        gain : `float`
+            Gain value in e^-/DN.
+        """
+        ampArr = ampData.array
+        ampArr[:] = ampArr[:] / gain
```
czwa, line 285: I'm adding this comment solely to state that this is dividing by the gain, which is correct as part of the mocks, because the mocks need to convert the electron image to ADU (which is the opposite sense than how it's done in IsrTask).  It might be worth adding a note in the docstring, so a code search that finds this will know why it's different than the `isrFunctions.applyGain` method.

czwa (approved):
I think I'm still happy/happy again/will be happy/etc. with this change.  A few comment/docstring suggestions, but I think this is fine to merge if Jenkins passes.

python/lsst/ip/isr/isrMockLSST.py
erykoff: My comment is that we need to be extra careful about distinguishing between a bias *frame* which is mean zero fixed pattern noise and the bias *level* which is like 30k and we remove by looking at the overscan levels. (But could be removed with a bias frame I suppose but that’s not how we are going to run lsstcam)

aferte: Thank you for clarifying - I updated the comment, it also means the biasMockLSST class should actually have a bias level of 0 which I will update.

python/lsst/ip/isr/isrMock.py
```diff
@@ -556,14 +557,63 @@ def getExposure(self):
...
         for amp in ccd:
             newAmp = amp.rebuild()
             newAmp.setLinearityCoeffs((0., 1., 0., 0.))
             newAmp.setLinearityType("Polynomial")
             newAmp.setGain(self.config.gain)
             newAmp.setSuspectLevel(25000.0)
             newAmp.setSaturation(32000.0)
+            readoutCorner = 'LL'
+
+            # Apply flips to bbox where needed
```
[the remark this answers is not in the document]
aferte: yes, I think that's right.

PR #316, opened 2024-03-28
DM-43563: Add flat correction placeholder.

aferte:


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -404,6 +404,13 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         default=True,
     )
 
+    # Flat correction.
+    doFlat = pexConfig.Field(
```
plazas, line 408: @erykoff has mentioned that we are going to have multiple types of flats, not just one single flat. Does this placeholder refer to a particular flat?  is this the dome flat? Should we add comments to the code?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1320,6 +1329,11 @@ def run(self, *, ccdExposure, dnlLUT=None, bias=None, deferredChargeCalib=None,
             bfKernelOut, bfGains = self.getBrighterFatterKernel(detector, bfKernel)
             ccdExposure = self.applyBrighterFatterCorrection(ccdExposure, flat, dark, bfKernelOut, bfGains)
 
+        if self.config.doFlat:
+            # Input units: electrons
+            self.log.info("Applying flat correction.")
+            self.flatCorrection(ccdExposure, flat)
```
plazas, line 1335: This `flatCorrection` method has not been implemented yet? Will there be future tickets to do so, and to add testing? If so, perhaps we could add their numbers here as comments.

plazas (approved):
I think it looks good as long as this is a "transition" ticket, i.e., there will be follow-up tickets for testing and making sure that things run properly. I think that's the case, so I'm just asking for some clarification in the comments.

python/lsst/ip/isr/isrTaskLSST.py
aferte: Ah good catch - thanks. For the moment we will use the same as for the original task, which I meant to copy over.

python/lsst/ip/isr/isrTaskLSST.py
aferte: I am adding a comment to say it would correspond to a background flat in this placeholder.

PR #320, opened 2024-04-12
DM-42927: Update cp_verify connections/classes/outputs for analysis_tools

czwa:
Ensure defects are passes to the IsrStatisticsTask so header keywords can be saved.


erykoff (approved):
Does this also need to go into the isrTaskLSST?

PR #322, opened 2024-04-29
DM-44049: Speed up variance plane tests in ip_isr

enourbakhsh:


tests/test_variance_plane.py
```diff
@@ -400,122 +510,23 @@ def buildExposure(
...
+            image = self.raw_image + self.background
+        else:
+            image = self.signal_free_raw_image + self.background
 
         # Add noise to the image which is in electrons. Note that we won't
         # specify a `sky_level` here to avoid double-counting it, as it's
-        # already included as the background.
-        image.addNoise(galsim.PoissonNoise(rng))
+        # already included in the image as the background.
+        image.addNoise(galsim.PoissonNoise(self.rng))
```
plazas, line 522: Is this meant to represent shot noise from which sources? Or read noise? (which should be Gaussian, although perhaps it doesn't matter that much). In any case, I thought Galsim had a CCD noise model (perhaps it doesn't matter that much for all this, I'm just curious).

tests/test_variance_plane.py
```diff
@@ -606,7 +617,7 @@ def test_variance_signal_removal(
         else:
             # Tighten tolerance for the 'predefined_gain_type' of 'per-amp' or
             # for a more accurate per-amp gain estimation strategy.
-            rtol = 2e-7
+            rtol = 3e-7
```
plazas, line 620: why was this raised?

plazas (approved):
LGTM, I just asked a couple of clarifying questions, but in Jira it is stated that the new changes that take the galsim profile creation out of the loop speed things up.

tests/test_variance_plane.py
enourbakhsh: We decided to only consider the shot noise of the image for this idealized test, but I recall receiving a similar comment from @czwa regarding adding the read noise. This would probably be beyond the scope of this ticket, but perhaps in the future, we can make this test more realistic by adding the read noise as well, and GalSim has the capability to do so.

tests/test_variance_plane.py
enourbakhsh: In the current version, it actually passes with `rtol=2e-7` without any issues. Initially, when I moved the GalSim generations to `setUp`, I forgot to also move the random seed assignment, causing a small number of pixels to exceed the tolerance level. As a result, I had to increase the `rtol` slightly. However, after moving the random seed lines to `setUp` as well, this issue was resolved, and it began working with the previous `rtol=2e-7` once again. I decided to keep the slightly larger `rtol` to mitigate sensitivity to future changes in the libraries and cross-platform variations.

PR #323, opened 2024-05-09
DM-41520: Add unit test for ISR LSST task

aferte:


python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
+                                       np.sqrt(self.config.skyLevel * self.config.gain))
 
             if self.config.doAddSource:
                 for sourceAmp, sourceFlux, sourceX, sourceY in zip(self.config.sourceAmp,
                                                                    self.config.sourceFlux,
                                                                    self.config.sourceX,
                                                                    self.config.sourceY):
                     if idx == sourceAmp:
-                        self.amplifierAddSource(ampData, sourceFlux, sourceX, sourceY)
+                        self.amplifierAddSource(ampData, sourceFlux * self.config.gain, sourceX, sourceY)
```
erykoff, line 148: Similar to above, we should note that the source config is in ADU.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
 
             # Other effects in e-
             if self.config.doAddFringe:
-                self.amplifierAddFringe(amp, ampData, np.array(self.config.fringeScale),
+                self.amplifierAddFringe(amp, ampData, np.array(self.config.fringeScale) * self.config.gain,
```
erykoff, line 152: And here that the fringe scale config is in ADU.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
                                         x0=np.array(self.config.fringeX0),
                                         y0=np.array(self.config.fringeY0))
 
             if self.config.doAddFlat:
                 if ampData.getArray().sum() == 0.0:
+                    # Add noise
```
erykoff, line 158: What is happening here?  The call below isn't adding noise, but is adding 1 electron and no noise.  @czwa?

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
                     self.amplifierAddNoise(ampData, 1.0, 0.0)
-                u0 = exposure.getDimensions().getX()
-                v0 = exposure.getDimensions().getY()
+                # Multiply each amplifiers by a Gaussian centered on u0 and v0
```
erykoff, line 160: -> `each amplifier`

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
+                u0 = exposure.getDetector().getBBox().getDimensions().getX()/2.
+                v0 = exposure.getDetector().getBBox().getDimensions().getY()/2.
                 self.amplifierMultiplyFlat(amp, ampData, self.config.flatDrop, u0=u0, v0=v0)
 
             # ISR effects
-            # 1. Add dark in e- (different from isrMock which does it in ADU)
+            # 1. Add dark in e-
```
erykoff, line 166: Add a comment that the dark current is specified in electrons per second in the config.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
-            # 3. Add read noise with or without a bias level
-            # to the image region in ADU.
-            self.amplifierAddNoise(ampData, self.config.biasLevel if self.config.doAddBias else 0.0,
-                                   self.config.readNoise / self.config.gain)
+            # 3. Add read noise to the image region in ADU.
+            if not self.config.calibMode:
+                self.amplifierAddNoise(ampData, 0.0,
+                                       self.config.readNoise / self.config.gain)
 
         # 4. Apply cross-talk in ADU
```
erykoff, line 184: I believe the crosstalk is defined in electrons (including the new non-linear crosstalk matrices that we'll want to test in the mocks in the near future).  So I think this might be in the wrong place.  (Unless we haven't updated the lsst isr task, in which case we need to do that as well, but maybe a separate ticket but we need that before we cut over to the new task).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -273,11 +273,28 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
...
+        doc="Mask saturated pixels? NB: this is totally independent of the"
+        " interpolation option - this is ONLY setting the bits in the mask."
+        " To have them interpolated make sure doSaturationInterpolation=True",
+        default=True,
+    )
+    saturation = pexConfig.Field(
+        dtype=float,
+        doc="The saturation level to use if no Detector is present in the Exposure (ignored if NaN)",
+        default=float("NaN"),
+    )
```
erykoff, line 287: The whole way that saturation is used/overridden in the current isr task is a mess, and I'm wondering if we can figure out how to do something better before this goes live.  We should discuss.

python/lsst/ip/isr/isrMock.py
```diff
@@ -274,7 +274,16 @@ def __init__(self, **kwargs):
                                          [1e-2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
                                          [1e-2, 0.0, 0.0, 2.2e-2, 0.0, 0.0, 0.0, 0.0],
                                          [1e-2, 5e-3, 5e-4, 3e-3, 4e-2, 5e-3, 5e-3, 0.0]])
-
+        # For debugging
```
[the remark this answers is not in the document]
czwa: I'd suggest making it a debug option:
```
from lsstDebug import getDebugFrame
# [...]
if getDebugFrame(self._display, "mockCrosstalkCoeffs"):
    self.crosstalkCoeffs = [...]
```

python/lsst/ip/isr/isrMockLSST.py
czwa: I think this was my hack for what is now better controlled by the `calibMode`.  If you're making a combined flat, you don't want any of the previous signals included, so mock that reaches this point that is universally 0.0 (which this isn't the best check for, but it's unlikely that a populated mock will sum to zero) and has `doAddFlat=True` is probably making the combined flat.  In this case, the mock should be filled with 1.0 so that the flat signal can be included.
This should probably instead be a check for `if self.config.calibMode:`.

python/lsst/ip/isr/isrMockLSST.py
czwa: There is a new units field in the crosstalk code now, and there are gain ratio parameters, but neither of these are currently included in the correction.  I haven't thought about how that should be included, and it depends on where the crosstalk is happening.

tests/test_isrTaskLSST.py
```diff
@@ -0,0 +1,426 @@
...
+        self.mockConfig.doAddCrosstalk = False
+        self.mockConfig.doAddParallelOverscan = False
+        self.mockConfig.doAddSerialOverscan = False
+
+    def setIsrConfig(self):
+        """Set all configs corresponding to ISR steps to False.
+        """
+        self.defaultAmpConfig = self.config.overscanCamera.\
+            getOverscanDetectorConfig(self.detector).defaultAmpConfig
+        # overscanDetectorConfig = self.config.overscanCamera.\
```
czwa, line 115: Is this comment still needed?

tests/test_isrTaskLSST.py
```diff
@@ -0,0 +1,426 @@
...
+        """
+        self.task = IsrTaskLSST(config=self.config)
+
+        mockMean = 0.
+        for amp in self.inputExp.getDetector():
+            bbox = amp.getRawDataBBox()
+            ampData = self.inputExp.image[bbox]
+            mockMean += np.nanmean(ampData.array)
+
+        # Not testing dnlLUT (not existant yet), deferred Charge,
```
czwa, line 156: `crosstalk` is being tested here.

tests/test_isrTaskLSST.py
```diff
@@ -0,0 +1,426 @@
...
+        self.setIsrConfig()
+        self.defaultAmpConfig.doSerialOverscan = True
+        self.defaultAmpConfig.doParallelOverscan = True
+
+        self.validateIsrResults()
+
+    def test_run_linearize(self):
+        """Test up to linearizer.
+        """
+        # TODO DM-???
```
czwa, line 218: Please file a ticket for this work, and update the comment here.

tests/test_isrTaskLSST.py
```diff
@@ -0,0 +1,426 @@
...
+        self.config.doDefect = True
+        self.config.doNanMasking = True
+        self.config.doWidenSaturationTrails = True
+        self.config.doDark = True
+
+        self.validateIsrResults()
+
+    def test_run_doBFcorrection(self):
+        """Test up to do BF correction
+        # TODO DM-??
```
czwa, line 345: And here, as above.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -135,63 +130,68 @@ def makeImage(self):
...
-                bbox = amp.getRawDataBBox().shiftedBy(amp.getRawXYOffset())
+                bbox = amp.getRawDataBBox()
 
             ampData = exposure.image[bbox]
 
             # Sky effects in e-
             if self.config.doAddSky:
-                self.amplifierAddNoise(ampData, self.config.skyLevel, np.sqrt(self.config.skyLevel))
+                self.amplifierAddNoise(ampData, self.config.skyLevel * self.config.gain,
+                                       np.sqrt(self.config.skyLevel * self.config.gain))
```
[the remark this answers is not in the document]
aferte: Agreed! I had filed [DM-42880](https://rubinobs.atlassian.net/browse/DM-42880) which was about update the configs to values that are LSSTCam-like, I added a bit about updating config parameters to the correct units.

erykoff (approved):
This is fine, but is it true that we don't have any tests of the quality of the ISR application?  Is there anything else we can do?  Or if we need to visualize, I don't know how hard it is to put in those visualize-on-config kind of things in the tests.

aferte:
I have been checking the ISR correction worked well by visually inspection and posting them in [here](https://docs.google.com/presentation/d/13GEH6BKwPd1APkfD-Fgk9Qd8Wm2O8AFj0IeU7-fPrbE/edit#slide=id.g2d22ee7ac28_0_9). Chris suggested I add these to the doc, I'll file a ticket about it. 
We could also check that the stats from the corrected image are similar to the one from a mock with just a source and sky background.

PR #324, opened 2024-05-10
DM-42766: Add fitResidualsSigmaMad to linearizer object.

erykoff:


czwa (approved):
I'm thinking of a future `IsrCalib` refactor that should make these kind of changes trivial.

PR #328, opened 2024-06-19
DM-44305: Implement Gaussian Process interpolation over bad pixels

PFLeget:
See this ticket --> https://rubinobs.atlassian.net/browse/DM-44305

### DIFF
--- a/tests/test_isrFunctions.py
+++ b/tests/test_isrFunctions.py
@@ -110,11 +110,22 @@ class IsrFunctionsCases(lsst.utils.tests.TestCase):
         ipIsr.makeThresholdMask(self.mi, 200, growFootprints=2,
                                 maskName='SAT')
         for growFootprints in range(0, 3):
-            interpMaskedImage = ipIsr.interpolateFromMask(self.mi, 2.0,
-                                                          growSaturatedFootprints=growFootprints,
-                                                          maskNameList=['SAT'])
-            numBit = ipIsr.countMaskedPixels(interpMaskedImage, "INTRP")
-            self.assertEqual(numBit, 40800, msg=f"interpolateFromMask with growFootprints={growFootprints}")
+            for useLegacyInterp in (False, True):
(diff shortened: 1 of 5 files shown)


tests/test_isrFunctions.py
```diff
@@ -110,11 +110,21 @@ def test_interpolateFromMask(self):
...
+                interpMaskedImage = ipIsr.interpolateFromMask(self.mi, 2.0,
+                                                              growSaturatedFootprints=growFootprints,
+                                                              maskNameList=['SAT'],
+                                                              useLegacyInterp=useLegacyInterp)
+                numBit = ipIsr.countMaskedPixels(interpMaskedImage, "INTRP")
+                if useLegacyInterp:
+                    self.assertEqual(numBit, 40800,
+                                     msg=f"interpolateFromMask with growFootprints={growFootprints}")
+                elif growFootprints == 0 and not useLegacyInterp:
+                    self.assertEqual(numBit, 0,
```
czwa, line 123: I'm not sure this makes sense to me (it's entirely possible it's a bug elsewhere).  With growFootprints=0 and using the new interpolator returns 0 interpolated pixels?  The fact that the old interpolator returns the same number for all values also doesn't make sense to me.  If you think it's doing the right thing in this case, that's fine, but I don't think I'm following what's happening here.

czwa (approved):
No major concerns.  The docstring needs updating, but that's trivial enough.

tests/test_isrFunctions.py
PFLeget: In this specific test, the full image is saturated and there is no growFootprints to use. So there is no information to use to interpolate over. On the GP algorithm side I took the choice of not doing anything in that specific case. Let me know if it sounds good or if I should do something else in that specific case.

czwa: I think this is fine as written, it's just been a while since I thought through these tests.

PR #330, opened 2024-06-26
DM-40903: upstream changes of isrTask to isrTaskLSST

aferte:
This PR should update isrTaskLSST with all the relevant changes done to isrTask since last year. I might have missed important updates or made unnecessary updates, let me know if this is the case. 
Jenkins is currently running.

### DIFF
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -202,6 +202,13 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         default=[],
     )
 
+    # Image conversion configuration
+    doConvertIntToFloat = pexConfig.Field(
+        dtype=bool,
+        doc="Convert integer raw images to floating point values?",
+        default=True,
+    )
+
     # Differential non-linearity correction.
     doDiffNonLinearCorrection = pexConfig.Field(
         dtype=bool,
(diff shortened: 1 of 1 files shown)


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1238,10 +1330,28 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+        if self.config.doBrighterFatter:
+            self.compareCameraKeywords(exposureMetadata, bfKernel, "brighter-fatter")
+        if self.config.doFlat:
+            self.compareCameraKeywords(exposureMetadata, flat, "flat")
+
         if self.config.doHeaderProvenance:
-            # Inputs have been validated, so we can add their date
-            # information to the output header.
-            exposureMetadata = ccdExposure.getMetadata()
+            # Add calibration date information to the output header.
```
czwa, line 1354: We'll need to add the other `doHeaderProvenance` code, from `isrTask.py:L1146`.  I think I had assumed this could be added to the `run` method, but looking at that code now, I think it needs to be in the `runQuantum` method for this task as well, as it gets some of the information that gets inserted into the header from the butler reference.  I think that will replace the `extractCalibDate` calls here.

czwa: I don't think you need to worry about the `additionalInputs`, though.  Since we don't mangle any of the inputs in the new `runQuantum`, everything should hopefully just work without any special cases.

czwa (approved):
A few comments on other things that should get copied over.

czwa (approved):
Looks good to me.

PR #331, opened 2024-07-05
DM-45131: Correct logger type in docstring

timj:


python/lsst/ip/isr/photodiode.py
```diff
@@ -40,7 +40,7 @@ class PhotodiodeCalib(IsrCalib):
         List of samples the photodiode was measured at.
     currentSamples : `list` or `numpy.ndarray`
         List of current measurements at each time sample.
-    log : `lsst.log.Log`, optional
+    log : `logging.Logger`, optional
```
timj, line 43: @czwa as a general comment I think it might be clearer if we included the base class parameters in the docstring and the `__init__`.

mwittgen (approved):

PR #332, opened 2024-07-08
DM-45151: Fix use of overscanIsInt and add appropriate tests.

erykoff:
This PR also removes some vestigial code that was not used or tested (from before we implemented the C++ row-by-row MEDIANs).


czwa (approved):
This all looks reasonable to me.

PR #334, opened 2024-07-18
Update IsrMockLSST and IsrTaskLSST for new calibration boxes

erykoff:
This is a major refactoring of the mock task and the isr task, following the detector and readout model that has been discussed on the ticket.  This also updates mock unit documentation, and crosstalk and linearity corrections now handle adu or electron units with appropriate checks/conversions.  CTI and BF are still not supported by the mocks or the new task (on the mock side because we need a forward model; on the task side because of unit questions).  The hope is that the new mock + testing here will ensure that we will not have unit confusion in the future.


python/lsst/ip/isr/crosstalk.py
```diff
@@ -675,18 +699,44 @@ def subtractCrosstalk(self, thisExposure, sourceExposure=None, crosstalkCoeffs=N
...
+                    fullAmplifier=fullAmplifier,
+                    parallelOverscan=parallelOverscan,
+                )
                 tImage.getMask().getArray()[:] &= crosstalk  # Remove all other masks
                 tImage -= backgrounds[tt]
                 sImage.scaledPlus(coeffs[ss, tt], tImage)
                 # Add the nonlinear term
                 if doSqrCrosstalk:
-                    tImageSqr = self.extractAmp(mi2, tAmp, sAmp, isTrimmed)
+                    tImageSqr = self.extractAmp(
```
czwa, line 732: Can you add a comment here pointing out that `mi2` has already been squared, so this is correct as written?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -926,6 +892,13 @@ def run(self,
             Do subtraction in parallel overscan region (only)?
         detectorConfig : `lsst.ip.isr.OverscanDetectorConfig`, optional
             Per-amplifier configs used when parallelOverscanRegion=True.
+        fullAmplifier : `bool`, optional
+            Use full amplifier and not just imaging region.
+        doSqrCrosstalk : `bool` or `None`, optional
+            If this is set to False, override the configuration.
```
czwa, line 898: Why do we need the ability to override the config here?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -941,77 +914,117 @@ def run(self,
...
+            if crosstalkSources:
+                # Parse crosstalkSources: Identify which detectors we have
+                # available
+                if isinstance(crosstalkSources[0], lsst.afw.image.Exposure):
+                    # Received afwImage.Exposure
+                    sourceNames = [exp.getDetector().getName() for exp in crosstalkSources]
+                elif isinstance(crosstalkSources[0], lsst.daf.butler.DeferredDatasetHandle):
+                    # Received dafButler.DeferredDatasetHandle
+                    detectorList = [source.dataId['detector'] for source in crosstalkSources]
+                    sourceNames = [camera[detector].getName() for detector in detectorList]
```
czwa, line 996: Should this perhaps be `detectorId in detectorList`?  Just so it's clear that this isn't a `detector` object.

python/lsst/ip/isr/crosstalk.py
```diff
@@ -941,77 +914,117 @@ def run(self,
...
+                        # Dereference the dafButler.DeferredDatasetHandle.
+                        sourceExposure = sourceExposure.get()
+                    if not isinstance(sourceExposure, lsst.afw.image.Exposure):
+                        raise RuntimeError("Unknown object passed as crosstalk sources.",
+                                           type(sourceExposure))
+
+                    self.log.info("Correcting detector %s with ctSource %s",
+                                  exposure.getDetector().getName(),
+                                  sourceExposure.getDetector().getName())
+                    crosstalk.subtractCrosstalk(exposure, sourceExposure=sourceExposure,
```
czwa, line 1020: This should have a check that `exposure.getBBox()` and `sourceExposure.getBBox()` match.  I think the DECam strategy has been to pre-process the `crosstalkSource` exposures, and those may not have the same trimmed state, and if they're in a different collection, I don't think this can be made consistent in the pipeline.

czwa (commented):
First pass on things up to `isrMockLSST.py` (exclusive).

python/lsst/ip/isr/crosstalk.py
erykoff: So in the case that we don't have proper gains in the new task, then we can't use the non-linear crosstalk term.  But if you're unhappy with this code then the configuration consistency could be checked in the validation in the new ISR task config, and we could make sure that we set it in the pipeline.  I should probably do that.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -129,139 +223,227 @@ def makeImage(self):
...
+                if self.config.calibMode:
+                    darkNoise = 0.0
+                else:
+                    darkNoise = np.sqrt(self.config.darkRate * self.config.darkTime)
+
+                self.amplifierAddNoise(ampImageData, darkLevel, darkNoise, rng=rngDark)
+
+            # 3. Add 2D bias residual (e-) to imaging portion of the amplifier.
+            if self.config.doAdd2DBias:
+                self.amplifierAddNoise(
```
czwa, line 296: Should this be structured?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1266,6 +1283,46 @@ def convertIntToFloat(self, exposure):
...
+        # This ID is a unique combination of {exposure, detector} for a raw
+        # image as we have here.
+        seed = exposure.info.id
+        if seed is None:
+            seed = fallbackSeed
+            self.log.warning("No exposure ID found; using fallback random seed %d.", fallbackSeed)
+
+        rng = numpy.random.RandomState(seed=seed)
+
+        if detectorConfig.integerDitherMode == "POSITIVE":
```
czwa, line 1312: Do we expect these will differ between detectors?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1346,77 +1403,190 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+            if not math.isnan(gain := overscanDetectorConfig.getOverscanAmpConfig(amp).gain):
+                gains[amp.getName()] = gain
+                self.log.warning(
+                    "Overriding gain for amp %s with configured value of %.3f.",
+                    amp.getName(),
+                    gain,
+                )
+
+        # Serial overscan correction.
+        # Units: ADU
```
czwa, line 1510: If the header keyword is lower case, can we make these guides lower case as well?  I'm close to suggesting an isrFunction to add the header keywords so we can enforce the set of possible values it can take, and correct "ADU"->"adu" and "electrons"->"electron".

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1346,77 +1403,190 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+            # This requires the PTC (tbd) with the temperature dependence.
+            self.log.info("Apply temperature dependence to the gains.")
+            gains, readNoise = self.correctGains(ccdExposure, ptc, gains)
+
+        # Do gain normalization; this may be the nominal gains.
+        if self.config.doApplyGains:
             # Input units: ADU
-            # Make sure that the units here are consistent with later
-            # application.
+            # Output units: electrons
```
czwa, line 1547: "electron" to keep guides and keywords consistent.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1346,77 +1403,190 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+            self.log.info("Using gain values to convert from ADU to electrons.")
+            isrFunctions.applyGains(ccdExposure, normalizeGains=False, ptcGains=gains, isTrimmed=False)
```
czwa, line 1549: This matches the Mock code, so I'm extra sure the boxes need an update.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1425,59 +1595,64 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
                 ccdExposure,
             )
 
+        # Linearity correction
+        # Units: electrons
+        if self.config.doLinearize:
+            self.log.info("Applying linearizer.")
+            # The linearizer is in units of ADU.
```
czwa, line 1602: Can't the new linearizer have whichever units, so shouldn't this check to see if there's a mismatch based on the linearizer header value?

python/lsst/ip/isr/overscanAmpConfig.py
```diff
@@ -36,16 +36,36 @@ class OverscanAmpConfig(pexConfig.Config):
         dtype=ParallelOverscanCorrectionTaskConfig,
         doc="Parallel overscan configuration.",
     )
+    saturation = pexConfig.Field(
```
czwa, line 39: If I've tracked through the code correctly, `saturation` and `suspectLevel` have their secondary definition here, with the `Amplifier.getSaturation()` being the primary source.  Will those be migrated to the PTC in the future?
If `gain` is overridable here (I think these are also their secondary definition location), shouldn't `readNoise` be here too?

tests/test_isrTaskLSST.py
```diff
@@ -20,396 +20,801 @@
...
+        # straight image.
+        self.assertLess(np.abs(np.median(delta[good_pixels])), 0.5)
+
+        # And overall where the interpolation is a bit worse but
+        # the statistics are still fine.  Note that this is worse than
+        # the defect case because of the widening of the saturation
+        # trail.
+        self.assertLess(np.std(delta), 7.0)
+
+    def get_mock_config_no_signal(self):
```
czwa, line 670: Are these configurations things we might expect to be useful outside of the tests, and if so, should they be Mock classes?

czwa (commented):
From my notes, my major questions are:
1) 2D bias and noise properties.
2) Boxes seem to mismatch with code; I've assumed the code is correct.
3) Linearity treatment around the high signal threshold.
4) Making units consistent.

I didn't look over the tests as thoroughly as the task, but the signs of the `asserts` looked reasonable.  I anticipate more discussion on Monday.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -129,139 +223,227 @@ def makeImage(self):
...
+                    darkLevel = self.config.darkRate * self.config.darkTime
+                if self.config.calibMode:
+                    darkNoise = 0.0
+                else:
+                    darkNoise = np.sqrt(self.config.darkRate * self.config.darkTime)
+
+                self.amplifierAddNoise(ampImageData, darkLevel, darkNoise, rng=rngDark)
+
+            # 3. Add 2D bias residual (e-) to imaging portion of the amplifier.
+            if self.config.doAdd2DBias:
```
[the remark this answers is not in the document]
erykoff: Yes that's true, it's an unstructured noise field so that we can have something that is a "2D bias residual pattern" even if it isn't an accurate one, so that we can test that it get removes (and not just subtracting zero).

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -129,139 +223,227 @@ def makeImage(self):
...
+            if self.config.doAddHighSignalNonlinearity:
+                if linearizer.linearityType[amp.getName()] != "Spline":
+                    raise RuntimeError("IsrMockLSST only supports spline non-linearity.")
+
+                coeffs = linearizer.linearityCoeffs[amp.getName()]
+                centers, values = np.split(coeffs, 2)
+
+                # This is an application of high signal non-linearity, so we
+                # set the lower values to 0.0 (this cut is arbitrary).
+                values[centers < self.config.highSignalNonlinearityThreshold] = 0.0
```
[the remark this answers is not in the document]
erykoff: It took me a moment to see the problem here.  I think that there is a bug in my current implementation that the linearizer from `makeLinearizer` has the low signal terms, but the application does not.  This needs to be fixed.

And more comments on where this is coming from will help.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -129,139 +223,227 @@ def makeImage(self):
...
+                    serialOverscanBBox = self.getFullSerialOverscanBBox(amp)
+                    serialOverscanData = exposure.image[serialOverscanBBox]
 
-        # 4. Apply cross-talk in ADU
+                    # Add read noise of mean 0
+                    # to the parallel and serial overscan regions.
+                    self.amplifierAddNoise(
+                        parallelOverscanData,
+                        0.0,
+                        self.config.readNoise,
```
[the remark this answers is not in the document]
erykoff: Not if we can't subtract it.  Again, this is not "noise", it's just a fixed field that I used a gaussian distribution to approximate.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -129,139 +223,227 @@ def makeImage(self):
...
-
-                    grownImageBBoxAll = grownImageBBox.expandedTo(serialOscanBBox)
-                    allData = exposure.image[grownImageBBoxAll]
-
-                    if self.config.doAddParallelOverscan:
-                        # Apply gradient along the Y axis
-                        self.amplifierAddXGradient(allData, -1.0 * self.config.overscanScale,
-                                                   1.0 * self.config.overscanScale)
+            # 11. Gain un-normalize (from e- to floating point ADU)
+            if self.config.doApplyGain:
```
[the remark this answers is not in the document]
erykoff: How does this disagree with the boxes?  Gain (ASPIC) comes right after crosstalk (cables).  (I'm not decomposing the different gains).

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -380,10 +700,11 @@ class TrimmedRawMockLSST(RawMockLSST):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.config.isTrimmed = True
-        self.config.doAddParallelOverscan = False
-        self.config.doAddSerialOverscan = False
+        self.config.doAddParallelOverscanRamp = False
+        self.config.doAddSerialOverscanRamp = False
 
 
+# Question: what is this used for?
```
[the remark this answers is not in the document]
erykoff: Ah okay, I have done this a different way in the test code, but I think it comes out the same way.  (Except I guess this is untrimmed?)

python/lsst/ip/isr/isrTaskLSST.py
erykoff: No, I guess not.  If you think it should be moved out of the overscan config to the regular config I can do that.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: Interesting point!  I can add that function and fix the units in the comments.  (The lower case is because this is what astropy units wants).

python/lsst/ip/isr/isrTaskLSST.py
erykoff: Because of the order in which we construct calibrations, it would be hard/impossible/circular to have an electron linearizer.  We should discuss.

python/lsst/ip/isr/overscanAmpConfig.py
erykoff: I have not yet tackled the "where does saturation live?" question, and I think that should be a different ticket, so yes it can still come from the amp.
For readnoise I thought we could get it from the overscan if we don't know it, and wouldn't need to be a config.  But I can add the additional option if you think it could be useful.

aferte (approved):
Thanks a lot for letting me look as well, looks great - there is indeed a lot in this ticket which was hard for me to parse so I don't have many comments. I am good with it, as we will test it on real data which will be very informative.

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -120,6 +202,18 @@ def makeImage(self):
         """
         exposure = self.getExposure()
 
+        # Set up random number generators for consistency of components,
+        # no matter the group that are configured.
+        rngSky = np.random.RandomState(seed=self.config.rngSeed + 1)
```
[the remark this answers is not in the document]
erykoff: The default levels are defined in the config `setDefaults` above and I don't think they're too crazy.  However, the serial and parallel ramps are greatly exaggerated (which I don't think is a problem; it can highlight problems with our approach).  Also the crosstalk matrix is unrealistic which is also useful to highlight problems.

python/lsst/ip/isr/isrMockLSST.py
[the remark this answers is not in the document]
erykoff: I think this confusion is because I had failed to update the ticket with the correct boxes.

python/lsst/ip/isr/isrTaskLSST.py
czwa: I'm fine with it here for now.  We can simplify in the future if needed.

czwa:
I think I'm happy, and hopefully we can get this into the weekly this week.

czwa (approved):

PR #337, opened 2024-07-31
DM-45507: Split amp offset config into measurement and application components

enourbakhsh:


python/lsst/ip/isr/isrTask.py
```diff
@@ -768,10 +768,22 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
     )
 
     # Amp offset correction.
+    # TODO: Remove this config field on DM-45526.
     doAmpOffset = pexConfig.Field(
         doc="Calculate and apply amp offset corrections?",
         dtype=bool,
         default=False,
+        deprecated="This field is no longer used. Will be removed after v28."
```
czwa, line 776: I think we want to keep this, and get rid of the two new configs.  `IsrTaskConfig.doAmpOffset` then controls whether to pass the exposure to the `IsrTask.ampOffset` task for the offsets to be measured, and the `IsrTask.ampOffset.doApplyAmpOffset` then controls whether those measured values are subtracted.  This reduces the number of config options, and I think ensures everything is consistent without needing extra validation.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1702,10 +1714,34 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
...
         self.roughZeroPoint(ccdExposure)
 
+        # An extra config validation logic that checks for compatibility
+        # among the config options for amp offset correction.
+        if self.config.doApplyAmpOffset and not self.config.doMeasureAmpOffset:
+            raise RuntimeError("doApplyAmpOffset requires doMeasureAmpOffset to be True.")
+
+        # Measure amp offset corrections within the CCD.
+        if self.config.doMeasureAmpOffset:
+            ampOffsetConfig = AmpOffsetConfig()
```
czwa, line 1724: This should be done automatically if `self.ampOffset` is a subtask.  I think this also confuses the config that's written to disk, as it doesn't know about these changes.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1702,10 +1714,34 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
...
+                self.log.info("Calculating amp offset corrections without applying them.")
+            ampOffsetTask = self.ampOffset(config=ampOffsetConfig)
+            ampOffsetTask.run(ccdExposure)
+
         # correct for amp offsets within the CCD
+        # TODO: Remove this whole ``if`` clause on DM-45526.
         if self.config.doAmpOffset:
-            self.log.info("Correcting amp offsets.")
-            self.ampOffset.run(ccdExposure)
+            if self.config.doMeasureBackground or self.config.doApplyAmpOffset:
```
czwa, line 1739: This check should be in the `IsrTaskConfig.validate`, as that will cause the task to raise before execution (it "fails fast" instead of doing a bunch of work before failing).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -369,6 +370,22 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         default=False,
     )
 
+    # Amp offset correction.
+    doMeasureAmpOffset = pexConfig.Field(
```
czwa, line 374: As above, `doAmpOffset` would control whether the exposure is passed to `AmpOffsetTask`, and that task's `doApplyAmpOffset` would control if the measured values are subtracted or not.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1523,6 +1540,25 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
                 maskNameList=list(self.config.maskListToInterpolate)
             )
 
+        # An extra config validation logic that checks for compatibility
+        # among the config options for amp offset correction.
+        if self.config.doApplyAmpOffset and not self.config.doMeasureAmpOffset:
+            raise RuntimeError("doApplyAmpOffset requires doMeasureAmpOffset to be True.")
+
+        # Measure amp offset corrections within the CCD.
+        if self.config.doMeasureAmpOffset:
```
czwa, line 1549: And this would just be `if self.config.doAmpOffset: self.ampOffset.run(ccdExposure)`.

tests/test_ampOffset.py
```diff
@@ -372,6 +373,7 @@ def testAmpOffset(self, valueType):
             config.doBackground = False
             config.doDetection = False
             config.ampEdgeWidth = 12  # Given 100x51 amps in our mock detector.
+            config.doApplyAmpOffset = True  # Updates the exposure in place.
```
czwa, line 376: Can this have a trivial test that with `doApplyAmpOffset=False` the image after being measured is the same as it was before?

enourbakhsh: Done!

enourbakhsh (commented):
Thanks, @czwa! I just addressed the review comments!

czwa (approved):
I like this version much better.  Thanks!

PR #339, opened 2024-08-13
DM-42438: Add realistic brighter-fatter to IsrMockLSST and add it to IsrTaskLSST

Alex-Broughton:


python/lsst/ip/isr/isrMock.py
```diff
@@ -295,11 +295,109 @@ def __init__(self, **kwargs):
                                              [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
                                              [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
                                              [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]])
-        self.bfKernel = np.array([[1., 4., 7., 4., 1.],
-                                  [4., 16., 26., 16., 4.],
-                                  [7., 26., 41., 26., 7.],
-                                  [4., 16., 26., 16., 4.],
-                                  [1., 4., 7., 4., 1.]]) / 273.0
+
```
erykoff, line 298: Please put a comment in on where these values came from.  Also, are these specific for the mock LSST camera layout?

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -68,6 +70,26 @@ class IsrMockLSSTConfig(IsrMockConfig):
...
+    )
+    diffusionStrength = pexConfig.Field(
+        dtype=float,
+        default=1.0,
+        doc="The diffusion effect scaling parameter."
+    )
+    nRecalc = pexConfig.Field(
+        dtype=int,
+        default=10000,
+        doc="Number of e- to accumulate before recalculating pixel shapes.",
```
erykoff, line 91: We have been standardizing on using the proper unit name `electron` in these places.  (And `adu` where appropriate).

python/lsst/ip/isr/isrMockLSST.py
```diff
@@ -322,7 +344,12 @@ def makeImage(self):
                 self.amplifierAddNoise(ampImageData, darkLevel, darkNoise, rng=rngDark)
 
             # 3. Add BF effect (electron) to imaging portion of the amp.
-            # TODO
+            if self.config.doAddSiliconPhysics is True:
+                self.amplifierAddSiliconPhysics(ampImageData,
+                                                self.config.rngSeed,
```
erykoff, line 349: So we don't want to use the seed directly if at all possible.  Instead we want to create the random number generator as above (see line 244), which in this case may be the galsim generator.  This ensures that we don't use the same seed for every amp (as is currently the case), but also ensures repeatability from run to run no matter what else is doing random numbers.

python/lsst/ip/isr/isrMock.py
erykoff: In fact, I think the new bfKernel should probably be put into the init for `isrMockLSST` instead.

python/lsst/ip/isr/isrMockLSST.py
erykoff: Is this to do with the silicon physics?  It maybe should be renamed to be clear.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -415,7 +415,7 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
     brighterFatterApplyGain = pexConfig.Field(
         dtype=bool,
         doc="Should the gain be applied when applying the brighter-fatter correction?",
-        default=True,
+        default=False,
```
erykoff, line 418: Should this config be removed entirely?  And instead if the image is in `adu` units then the gain is applied, and if it is in `electron` units the gain is not applied?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1061,32 +1061,21 @@ def getBrighterFatterKernel(self, detector, bfKernel):
             else:
                 raise RuntimeError("Failed to extract kernel from new-style BF kernel.")
         elif bfKernel.level == 'AMP':
-            self.log.warning("Making DETECTOR level kernel from AMP based brighter "
-                             "fatter kernels.")
+            self.log.info("Making DETECTOR level kernel from AMP based brighter "
+                          "fatter kernels.")
```
erykoff, line 1065: Thank you for making this not a warning.  Is this even an `info` or should it be a `debug`?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1061,32 +1061,21 @@ def getBrighterFatterKernel(self, detector, bfKernel):
...
-        # the flats and darks applied so we can work in units of electrons
-        # or holes. This context manager applies and then removes the darks
-        # and flats.
-        #
-        # We also do not want to interpolate values here, so operate on
-        # temporary images so we can apply only the BF-correction and roll
-        # back the interpolation.
-        # This won't be necessary once the gain normalization
-        # is done appropriately.
+        # We need to interpolate before we do B-F.
```
erykoff, line 1071: Should probably add a comment that the fwhm parameter is currently unused by the code that is called.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1061,32 +1061,21 @@ def getBrighterFatterKernel(self, detector, bfKernel):
...
-            )
+        isrFunctions.interpolateFromMask(
+            maskedImage=interpExp.getMaskedImage(),
+            fwhm=self.config.brighterFatterFwhmForInterpolation,
+            growSaturatedFootprints=self.config.growSaturationFootprintSize,
+            maskNameList=list(self.config.brighterFatterMaskListToInterpolate)
+        )
         bfExp = interpExp.clone()
         self.log.info("Applying brighter-fatter correction using kernel type %s / gains %s.",
                       type(bfKernel), type(bfGains))
```
erykoff, line 1081: Can you update this log message?  There's no point in logging a python `type` in general.  Can just be `Applying brighter-fatter correction.`

erykoff: Also, below here I think that `ApplyGain` could be derived from the units.

tests/test_isrTaskLSST.py
```diff
@@ -49,7 +49,7 @@ def setUp(self):
         self.bias = isrMockLSST.BiasMockLSST().run()
         self.dark = isrMockLSST.DarkMockLSST().run()
         self.flat = isrMockLSST.FlatMockLSST().run()
-        self.bf_kernel = isrMockLSST.BfKernelMockLSST().run()
+        self.bfKernel = isrMockLSST.BfKernelMockLSST().run()
```
erykoff, line 52: This test code is `snake_case` so I think `bf_kernel` is correct. 😄

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +476,79 @@ def test_isrFlat(self):
...
+
+        mock = isrMockLSST.IsrMockLSST(config=mock_config)
+        input_exp = mock.run()
+
+        isr_config = self.get_isr_config_electronic_corrections()
+        isr_config.doBias = True
+        isr_config.doDark = True
+        isr_config.doFlat = True
+        isr_config.doDefect = True
+        isr_config.doBrighterFatter = False
```
erykoff, line 499: We are not running the bf correction here, which seems incorrect.

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +476,79 @@ def test_isrFlat(self):
...
+    def test_isrBrighterFatter(self):
+        """Test processing of a flat frame."""
+        mock_config = self.get_mock_config_no_signal()
+        mock_config.doAddDark = True
+        mock_config.doAddFlat = True
+        # The doAddSky option adds the equivalent of flat-field flux.
+        mock_config.doAddSky = True
+        mock_config.doAddSiliconPhysics = True
+        mock_config.bfStrength = 1.0
+        mock_config.diffusionStrength = 0.0
```
erykoff, line 488: If we're going to override this, this should be the default.

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +476,79 @@ def test_isrFlat(self):
         ratio = result2.exposure.image.array / result.exposure.image.array
         self.assertFloatsAlmostEqual(ratio, flat_nodefects.image.array, atol=1e-4)
 
+    def test_isrBrighterFatter(self):
+        """Test processing of a flat frame."""
+        mock_config = self.get_mock_config_no_signal()
+        mock_config.doAddDark = True
+        mock_config.doAddFlat = True
+        # The doAddSky option adds the equivalent of flat-field flux.
+        mock_config.doAddSky = True
```
erykoff, line 485: In this test we aren't adding a source; is that intentional?

erykoff (changes requested):

python/lsst/ip/isr/isrMockLSST.py
Alex-Broughton: Took away the diffusion part since it is unused anyways; renamed to `doAddBrighterFatter`.

python/lsst/ip/isr/isrTaskLSST.py
Alex-Broughton: This honestly should become a default, but might be out of scope of the mock BF ticket

Alex-Broughton: Should we just get rid of the brighter fatter kernel `bfk.level` parameter entirely in this ticket? This would extend this ticket beyond ip_isr.

tests/test_isrTaskLSST.py
Alex-Broughton: I found that the BF correction (even in it's best state) isn't correct perfect enough to pass the test cases in the skyImage test without massively opening up the assert validity range to the point that they really aren't testing anything. So I formed this new test (which is also very fast to run < 10s), which sees if the BF correction can reconstruct poisson variations in a flat image.

python/lsst/ip/isr/isrTaskLSST.py
Alex-Broughton: Sure. I'll need to review it more carefully to make sure it doesn't break anything in isrFunctions. I'll save that for the morning after I get some sleep :)

tests/test_isrTaskLSST.py
Alex-Broughton: This is the default actually :)

tests/test_isrTaskLSST.py
Alex-Broughton: Oops remnant from a test. Fixed. 
Made sure the same test cases passes ✅

python/lsst/ip/isr/isrTaskLSST.py
erykoff: We shouldn't need to touch anything in isrFunctions.  This is only used internally by this task, and can be inferred from the units directly.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: I wouldn't change the bfk format, that does seem like a bridge too far.  I'm just saying that because this is such a standard operation I'm not sure if it needs to be logged at all at info level.  Logging below that BF is being applied is important, but this just seems ... extraneous.

tests/test_isrTaskLSST.py
erykoff: I haven't run the test myself, but is it passing whether or not we apply the BF correction?  If so, then I fear it's too trivial.  It may be worth (a) adding the source; (b) doing the comparison below away from the source; (c) at least doing a simple specific test of the size of the source before or after BF correction and confirming that it is closer to the correct/true width after correction.

python/lsst/ip/isr/isrTaskLSST.py
Alex-Broughton: For clarity, we I think we should keep it in, until we make this the only option.

tests/test_isrTaskLSST.py
Alex-Broughton: The test failed without the BF correction turned on. But the test you describe seems more stable. I will move it towards that.

python/lsst/ip/isr/isrMock.py
Alex-Broughton: Ok I left the original Gaussian kernel in isrMock, and now override it with the empirical one for the simulated LSSTCam in isrMockLSST

python/lsst/ip/isr/isrTaskLSST.py
Alex-Broughton: I've added a check of the current exposure units and determine the application of the gains internally. I then deleted the user config for` brighterFatterApplyGain`. Everything should always be applied in the correct units now.

python/lsst/ip/isr/isrTaskLSST.py
Alex-Broughton: I added the check in the main body of the run() method, this way someone just following along with the overall structure of main() can easily follow the unit progression of each step.

tests/test_isrTaskLSST.py
Alex-Broughton: Updated the test.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1655,12 +1640,33 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
             isrFunctions.widenSaturationTrails(ccdExposure.getMaskedImage().getMask())
 
         # Brighter/Fatter
-        # FIXME: watch out (again) for gain units.
         # Units: electron
         if self.config.doBrighterFatter:
-            self.log.info("Applying Bright-Fatter kernels.")
+            self.log.info("Applying brighter-fatter correction.")
```
erykoff, line 1645: This does not need to be logged here and in the helper function.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1655,12 +1640,33 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+
+            # Needs to be done in electrons; applyBrighterFatterCorrection
+            # will convert the image if necessary.
+            if exposureMetadata["LSST ISR UNITS"] == "electron":
+                brighterFatterApplyGain = False
+            else:
+                brighterFatterApplyGain = True
+                exposureMetadata["LSST ISR UNITS"] == "electron"
+
+            if brighterFatterApplyGain and (ptc is not None) and (not bfGains == gains):
```
erykoff, line 1657: I don't love `not bfGains == gains`.  I guess these are dictionaries?  So this will work, but `bfGains != gains` is clearer.  And I think you should comment below that these are amp-keyed dictionaries of gains.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1655,12 +1640,33 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
             bfKernelOut, bfGains = self.getBrighterFatterKernel(detector, bfKernel)
-            ccdExposure = self.applyBrighterFatterCorrection(ccdExposure, flat, dark, bfKernelOut, bfGains)
+
+            # Needs to be done in electrons; applyBrighterFatterCorrection
+            # will convert the image if necessary.
+            if exposureMetadata["LSST ISR UNITS"] == "electron":
+                brighterFatterApplyGain = False
+            else:
+                brighterFatterApplyGain = True
+                exposureMetadata["LSST ISR UNITS"] == "electron"
```
erykoff, line 1655: This is not correct.  The gain is applied in a context, so the application is temporary.  We don't want to output `electron` units if `doApplyGain = False`.

erykoff: (And to be fair, this will be a very unusual code path, but I want to make sure it's correct).

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +477,38 @@ def test_isrFlat(self):
...
+            )
+
+        # Measure the source size in the BF-corrected image.
+        # The injected source is a Gaussian with 3.0px
+        image = galsim.ImageF(result.exposure.image.array)
+        hsm_result = galsim.hsm.FindAdaptiveMom(image, guess_centroid=galsim.PositionD(50., 25.),
+                                                strict=False)
+        measured_sigma = hsm_result.moments_sigma
+
+        # With the BF correction, the flats should have the same variation.
```
erykoff, line 509: This isn't a flat, it's an image.

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +477,38 @@ def test_isrFlat(self):
         ratio = result2.exposure.image.array / result.exposure.image.array
         self.assertFloatsAlmostEqual(ratio, flat_nodefects.image.array, atol=1e-4)
 
+    def test_isrBrighterFatter(self):
+        """Test processing of a flat frame."""
+        mock_config = self.get_mock_config_clean()
+        mock_config.doAddSource = True
+        mock_config.sourceFlux = [75000.0]
+        mock_config.isTrimmed = False
+        mock_config.doAddBrighterFatter = True
```
erykoff, line 486: Are we not putting sky, flat, everything else?  I think we want the whole thing (or a second test that does the whole thing).

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +477,38 @@ def test_isrFlat(self):
...
+        # Measure the source size in the BF-corrected image.
+        # The injected source is a Gaussian with 3.0px
+        image = galsim.ImageF(result.exposure.image.array)
+        hsm_result = galsim.hsm.FindAdaptiveMom(image, guess_centroid=galsim.PositionD(50., 25.),
+                                                strict=False)
+        measured_sigma = hsm_result.moments_sigma
+
+        # With the BF correction, the flats should have the same variation.
+        self.assertFloatsAlmostEqual(measured_sigma, 3.0, rtol=1.0e-3)
+
```
erykoff, line 511: If we add in the sky and everything else (which we should on this or a second test) we should also check the equality everywhere except around the source (how far out is up to you/the tests).  It'll also be important to add in the defects, etc, to make sure that all the interpolation code is working correction in the BF context.

tests/test_isrTaskLSST.py
Alex-Broughton: We already have the skyImage test which tests all of those. But I realized that to properly test the BF component, we need a separate test, so I only draw a source onto an blank image.

python/lsst/ip/isr/isrTaskLSST.py
Alex-Broughton: Ah! I forgot that the application of gain in the helper function applies gain through gainContext

tests/test_isrTaskLSST.py
erykoff: The skyImage test does not (currently) have BF turned on for either the mock or the ISR.  I would like to see a test that has all of everything, even if we can't do pixel-perfect comparisons.

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +477,107 @@ def test_isrFlat(self):
...
+        mock_config.isTrimmed = False
+        mock_config.doAddDark = True
+        mock_config.doAddFlat = True
+        mock_config.doAddSky = True
+        mock_config.doAddSource = True
+        mock_config.sourceFlux = [75000.0]
+        mock_config.doAddBrighterFatter = False
+
+        mock = isrMockLSST.IsrMockLSST(config=mock_config)
+        input_exp = mock.run()
```
erykoff, line 525: Please update this to `input_exp2` and below so that if we add any relevant tests we have both the raw with BF and the raw without.

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +477,107 @@ def test_isrFlat(self):
...
+        hsm_result_truth = galsim.hsm.FindAdaptiveMom(image_truth, guess_centroid=source_centroid,
+                                                      strict=False)
+        measured_sigma = hsm_result.moments_sigma
+        true_sigma = hsm_result_truth.moments_sigma
+        self.assertFloatsAlmostEqual(measured_sigma, true_sigma, rtol=3e-3)
+
+        # Check that the variance in an amp far away from the
+        # source is expected. The source is in amp 0; this will
+        # check the variation in neighboring amp 1
+        # np.std(result_truth.exposure[test_amp_bbox].image.array)/
```
erykoff, line 562: Please remove this debugging (?) comment.

tests/test_isrTaskLSST.py
```diff
@@ -476,6 +477,107 @@ def test_isrFlat(self):
...
+        stdev = np.std(result.exposure[test_amp_bbox].image.array)
+        stdev_truth = np.std(result_truth.exposure[test_amp_bbox].image.array)
+        self.assertFloatsAlmostEqual(stdev, stdev_truth, atol=3*stdev_truth/np.sqrt(n_pixels))
+
+        # Check that the variance in the amp with a defect is
+        # unchanged as a result of applying the BF correction after
+        # interpolating. The defect was added to amplifier 2.
+        test_amp_bbox = result.exposure.detector.getAmplifiers()[2].getBBox()
+        dummy = isrMockLSST.IsrMockLSST()
+        defect_bbox = dummy.makeDefectList(isTrimmed=False)[0].getBBox()
```
erykoff, line 574: I'm surprised this works with the `isTrimmed=False`; isn't that backwards?  But I think that the test utility function `self.get_non_defect_pixels` will get you want you want here.

tests/test_isrTaskLSST.py
```diff
@@ -766,17 +866,17 @@ def get_isr_config_electronic_corrections(self):
         isr_config.doBias = False
         isr_config.doDark = False
         isr_config.doFlat = False
+        isr_config.doBrighterFatter = False
```
erykoff, line 869: Please add a comment here about why we have this `False` here.  Also I think that this should be moved to line 875 (electronic effects that the tests support), but it is `False` because it is only used in a specialized tests.

erykoff (approved):
Some small remaining comments; otherwise looks great!

Alex-Broughton:
Jenkins passed;

https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/2203/pipeline

PR #340, opened 2024-08-16
DM-45815: Add utility functions to access gain and readnoise

kadrlica:
The preferred values of the gain and read noise are currently being set in header keys. We'd like to provide access functions to avoid making the user know the precise header keyword prefixes (which are changing with the new isrTask).


kadrlica:
Jenkins build [#2490](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/2490/pipeline).

kadrlica:
@Alex-Broughton would you be able to review this?

Alex-Broughton:
Sure let me take a look.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -1222,3 +1224,65 @@ def countMaskedPixels(maskedIm, maskPlane):
...
+        The exposure to find gains for.
+
+    Returns
+    -------
+    gains : `dict` [`str` `float`]
+        Dictionary of gain values, keyed by amplifier name.
+    """
+    det = exposure.getDetector()
+    if det is None:
+        return dict()
```
Alex-Broughton, line 1244: In keeping this consistent with other parts of DM, it might be better to return None here or a dict of NaNs rather than an empty dict.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -1222,3 +1224,65 @@ def countMaskedPixels(maskedIm, maskPlane):
...
+
+    metadata = exposure.getMetadata()
+    gains = {}
+    for amp in det:
+        ampName = amp.getName()
+        # The key may use the new LSST ISR or the old LSST prefix
+        if (key1 := f"LSST ISR GAIN {ampName}") in metadata:
+            gains[ampName] = metadata[key1]
+        elif (key2 := f"LSST GAIN {ampName}") in metadata:
```
Alex-Broughton, line 1253: Both the old and new tasks have the same keys for "LSST ISR GAIN"

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -1222,3 +1224,65 @@ def countMaskedPixels(maskedIm, maskPlane):
...
+        The exposure to find read noise for.
+
+    Returns
+    -------
+    readnoises : `dict` [`str` `float`]
+        Dictionary of read noise values, keyed by amplifier name.
+    """
+    det = exposure.getDetector()
+    if det is None:
+        return dict()
```
Alex-Broughton, line 1275: Same as above

python/lsst/ip/isr/isrFunctions.py
kadrlica: Thanks. Isn't it possible that this code will run on data that has been processed with older versions of ISR where only one keyword is present?

python/lsst/ip/isr/isrFunctions.py
kadrlica: Oh, interesting. I asked @erykoff about this and he suggested the empty dict.

kadrlica: Eli confirmed using an empty dict.

python/lsst/ip/isr/isrFunctions.py
kadrlica: Eli confirmed using an empty dict.

python/lsst/ip/isr/isrFunctions.py
kadrlica: Eli suggests:
```I would leave in the backup check for backwards compatibility.  It’s the practical choice.```

Comment (Alex-Broughton) [reply python/lsst/ip/isr/isrFunctions.py:1253]: You and Eli are correct, keeping the original keys is fine.

File: python/lsst/ip/isr/isrFunctions.py [hunk 4]
```diff
@@ -1222,3 +1224,65 @@ def countMaskedPixels(maskedIm, maskPlane):
...
+
+    Parameters
+    ----------
+    exposure : `lsst.afw.image.Exposure`
+        The exposure to find gains for.
+
+    Returns
+    -------
+    gains : `dict` [`str` `float`]
+        Dictionary of gain values, keyed by amplifier name.
```
Alex-Broughton, line 1240: Please add a note to the doc string that it will return an empty dict if the detector is None.

Alex-Broughton (approved):
Looks good 👍

PR #342, opened 2024-08-22
DM-45871: Fix confusing log message in ampOffset code

enourbakhsh:


python/lsst/ip/isr/ampOffset.py
```diff
@@ -518,19 +519,20 @@ def getInterfaceOffset(self, ampIdA, ampIdB, edgeA, edgeB):
...
+                        f"The fraction of unmasked pixels for amp interface {interfaceId} is below the "
+                        f"threshold ({ampEdgeGoodFrac:.2f} < {self.config.ampEdgeMinFrac}). Setting the "
+                        f"interface offset to {interfaceOffset}."
+                    )
+                if maxOffsetFail:
+                    self.log.warning(
+                        "The absolute offset value exceeds the limit "
+                        f"({np.abs(interfaceOffset):.2f} > {self.config.ampEdgeMaxOffset} ADU). Setting the "
+                        f"interface offset to {interfaceOffset}."
+                    )
```
erykoff, line 534: I think this log needs more info.  Because it's actually setting the interfaceOffset to 0 (line below).  I think these logs should be `resetting interface offset from {interfaceOffset} to 0.` or something to that effect.

Also, the reset to 0 should be inside the check of `doApplyAmpOffset`.  In the measurement case I don't care about maxima, I want to know what the offset was.

erykoff (changes requested):

erykoff (approved):

python/lsst/ip/isr/ampOffset.py
```diff
@@ -518,19 +519,20 @@ def getInterfaceOffset(self, ampIdA, ampIdB, edgeA, edgeB):
         minFracFail = ampEdgeGoodFrac < self.config.ampEdgeMinFrac
         maxOffsetFail = np.abs(interfaceOffset) > self.config.ampEdgeMaxOffset
         if minFracFail or maxOffsetFail:
```
enourbakhsh, line 521: Hmm, maybe we should also move `if minFracFail or maxOffsetFail` inside the `doApplyAmpOffset` check?

python/lsst/ip/isr/ampOffset.py
leeskelvin: I think it's important for the `measure=True`, `apply=False` scenario to also apply the safety checks. For example, in a scenario where my 4000-long pixel amp interface only has 1 valid pixel (the rest being masked), I could certainly calculate an `interfaceOffset` value here, and then feed that into my fitting code to generate my amp pedestals, but the value of these pedestals is obviously highly noisy.

erykoff: Safety check in terms of number of pixels, sure.  But in terms of the absolute value of the offset, this is something that I want to know whatever it is in measurement mode.  (I can also turn off the safety check of course if that's what you prefer)

leeskelvin: Which value are you interested in exactly? The `interfaceOffset` != `ampPedestal` value. We're currently only reporting the latter.

erykoff: Ah I guess I care about the `ampPedestal`, but I do _not_ want any of the `interfaceOffset` set to 0 if they are not 0.

leeskelvin: I see your point, and I agree it would be nice to know the raw `interfaceOffset` values after the fact. However, I thought the purpose of this `measure=True`/`apply=False` exercise was to facilitate producing metrics that would show what amp offset values would be applied if we did indeed have `apply=True`? To that end, wouldn't precisely replicating what happens in an `apply=True` be what we need here?

Perhaps we're instead talking about tracking a new metric (which would be beyond the scope of this ticket). Something like `ampOffsetOriginal`, or, even better, `interfaceOffsetOriginal`?

erykoff: There is the subtlety of the difference between `interfaceOffset` and `ampOffset` which might make my thoughts incorrect.  But at the same time, in measurement mode I would like to know what the full set of offsets are, even without safety checks.  (Well ... safety checks in terms of number of valid pixels are 👍 but safety checks in terms of how big can the offset be are 😕 ).  Because the metric will be wrong if it says "there is no amp offset here" when in actual fact the amp offset is very large.
But for these purposes (which is one of the ways I intend to use this) the offset max can be turned off (or set to a very large value).  So this is just a convenience.  If you think it should be applied even if apply is False, I won't scream too loud and can just do the config override to get the functionality that I need.

enourbakhsh: Based on our discussions, we now always enforce do-no-harm safety checks. If someone wants to bypass safety checks on the maximum interface offset size, they can configure `ampEdgeMaxOffset` to a very large value. Logging is set to warning if `doApplyAmpOffset` is true; otherwise, it logs as debug. Additionally, I filed a ticket to track a new metric for monitoring `interfaceOffset` statistics.

PR #343, opened 2024-09-03
DM-45976: Fix units bugs with camera read noise / ptc read noise / effective ptc / variance plane creation

Alex-Broughton:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -389,20 +389,23 @@ def darkCorrection(maskedImage, darkMaskedImage, expScale, darkScale, invert=Fal
 
 def updateVariance(maskedImage, gain, readNoise):
     """Set the variance plane based on the image plane.
+    Expect maskedImage to have units of adu or electron
+    if gain is 1.0. Will always produce a variance plane
+    int the same units as the image.
```
erykoff, line 394: I think this needs more punctuation and a blank line after the one-line summary.  Something like:
```
"""Set the variance plane based on the image plane.

The maskedImage will have units of `adu` (if gain != 1.0) or
electron (if gain == 1.0). This routine will always produce a
variance plane in the same units as the image.

Parameters
------------
```

python/lsst/ip/isr/isrTask.py
```diff
@@ -1853,6 +1853,17 @@ def defineEffectivePtc(self, ptcDataset, detector, bfGains, overScans, metadata)
                                          gain, bfGain)
                     boolGainMismatch = True
 
+            # Gain:
+            if math.isnan(gain):
+                gain = 1.0
+                self.log.warning("Gain for amp %s set to NAN! Updating to"
```
erykoff, line 1859: NAN -> NaN.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1865,28 +1876,42 @@ def defineEffectivePtc(self, ptcDataset, detector, bfGains, overScans, metadata)
...
+                    # After version 1.9, we standardized the
+                    # PhotonTransferCurveDataset noise units to electron to fix
+                    # this bug. If a user tries to use an earlier version of
+                    # PTC with this fit type, we must be sure to do the
+                    # calculations properly. More information about this noise
+                    # issue can be found in DM-45976.
+                    self.log.info(f"Input PTC VERSION ({ptcVersion}) < 1.9 and"
+                                  "ptcFitType == FULLCOVARIANCE. Applying fix"
+                                  f"for DM-45976 noise issue for AMP {ampName}.")
+                    noise = numpy.sqrt(noise)
```
erykoff, line 1903: This is the wrong place for this version check and info log.  It should be done when the PTC is read in, similar to the other PTC version patches.  Also note that the log is missing spaces so it will look like `andptcFitType ... fixfor ...`.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1865,28 +1876,42 @@ def defineEffectivePtc(self, ptcDataset, detector, bfGains, overScans, metadata)
...
             else:
                 # Finally, try with the amplifier noise.
-                # We already have a detector at this point. If there
-                # was no detector to begin with, one would have
-                # been created with self.config.gain and
-                # self.config.noise.
+                # We already have a detector at this point. If there was no
+                # detector to begin with, onewould have been created with
```
erykoff, line 1907: onewould -> one would.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1865,28 +1876,42 @@ def defineEffectivePtc(self, ptcDataset, detector, bfGains, overScans, metadata)
...
-            if math.isnan(gain):
-                gain = 1.0
-                self.log.warning("Gain for amp %s set to NAN! Updating to"
-                                 " 1.0 to generate Poisson variance.", ampName)
-            elif gain <= 0:
-                patchedGain = 1.0
-                self.log.warning("Gain for amp %s == %g <= 0; setting to %f.",
-                                 ampName, gain, patchedGain)
-                gain = patchedGain
+                # Amplifier object noise is always in adu; we must
```
erykoff, line 1910: The read noise is always in `electron` in the amplifier object.  See https://github.com/lsst/afw/blob/7f102674c438d58292d6800526c800bf78397141/include/lsst/afw/cameraGeom/Amplifier.h#L120 and ticket description.

python/lsst/ip/isr/isrTask.py
```diff
@@ -2261,6 +2286,9 @@ def updateVariance(self, ampExposure, amp, ptcDataset):
         gain = ptcDataset.gain[ampName]
         readNoise = ptcDataset.noise[ampName]
 
+        # The image will always be in adu and the noise
```
erykoff, line 2289: I think it's important in these comments to say "read noise" (in fact `ptc.noise` was probably a bad name for the variable but we're stuck with it).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1496,6 +1503,26 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+                # use an earlier version of PTC with this fit type, we must be
+                # sure to do the calculations properly. More information about
+                # this noise issue can be found in DM-45976.
+                self.log.info(f"Input PTC VERSION ({ptcVersion}) < 1.9 and"
+                              "ptcFitType == FULLCOVARIANCE. Applying fix"
+                              f"for DM-45976 noise issue.")
+                for amp in detector:
+                    ptc.noise[amp.getName()] = numpy.sqrt(ptc.noise[amp.getName()])
+
+        # Set which gains to use
```
erykoff, line 1525: Similar to above, if this code is in the `ptcDataset` then it won't need to be repeated (with the same formatting errors 😄 ).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1526,7 +1554,11 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
                     if serialOverscan is None:
                         ptc.noise[amp.getName()] = 0.0
                     else:
-                        ptc.noise[amp.getName()] = serialOverscan.residualSigma
+                        # All PhotonTransferCurveDataset objects should contain
+                        # noise attributes in units of electrons. The read
```
erykoff, line 1558: Probably should say "read noise" here as well.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -189,11 +190,12 @@ class PhotonTransferCurveDataset(IsrCalib):
     Version 1.6 adds the `rowMeanVariance` attribute.
     Version 1.7 adds the `noiseList` attribute.
     Version 1.8 adds the `ptcTurnoffSamplingError` attribute.
+    Version 1.9 standardizes PTC noise units to electron.
     """
 
     _OBSTYPE = 'PTC'
     _SCHEMA = 'Gen3 Photon Transfer Curve'
-    _VERSION = 1.8
+    _VERSION = 1.9
```
erykoff, line 198: Look for the version checks on ptc load below and that is the place for the unit conversion.

tests/test_isrTaskLSST.py
```diff
@@ -65,10 +65,10 @@ def setUp(self):
                                               ptcFitType='DUMMY_PTC',
                                               covMatrixSide=1)
 
-        # PTC records noise units in adu.
+        # PTC records noise units in electrons.
```
erykoff, line 68: "... in electron, same as the configuration parameter."

tests/test_isrTaskLSST.py
```diff
@@ -99,6 +99,11 @@ def test_isrBootstrapBias(self):
         isr_config.doApplyGains = False
         isr_config.doBias = True
 
+        # Need to make sure we are not masking the negative variance
```
erykoff, line 102: The key here is that we don't want to mask negative pixels, because these are legal in calibration-corrected calibrations.  And `maskNegativeVariance` inadvertently does that.

tests/test_isrTaskLSST.py
```diff
@@ -134,17 +139,15 @@ def test_isrBootstrapBias(self):
...
+        self.assertIn(key, metadata)
+        self.assertEqual(metadata[key], "electron")
+
         for amp in self.detector:
             amp_name = amp.getName()
             key = f"LSST ISR GAIN {amp_name}"
             self.assertIn(key, metadata)
             self.assertEqual(metadata[key], isr_config.nominalGain)
-            key = f"LSST ISR READNOISE {amp_name}"
-            self.assertIn(key, metadata)
```
erykoff, line 143: Upon second thought, checking the existence of this keyword is a good idea so this part should be restored.

erykoff: I'm not sure what to do with the read noise units; they are not really electron in this case since we don't know the gain.  But don't change this part, I think when I change to the better "bootstrap" terminology on another branch it'll be clearer.

tests/test_isrTaskLSST.py
```diff
@@ -257,19 +270,19 @@ def test_isrBootstrapFlat(self):
...
+        # And add the gain and read noise (in electron) per amp.
         for amp in self.detector:
             # We need to use the gain and read noise from the header
             # because these are bootstraps.
             gain = result.exposure.metadata[f"LSST ISR GAIN {amp.getName()}"]
             read_noise = result.exposure.metadata[f"LSST ISR READNOISE {amp.getName()}"]
 
             expected_variance[amp.getBBox()].array /= gain
-            expected_variance[amp.getBBox()].array += read_noise**2.
+            expected_variance[amp.getBBox()].array += (read_noise/gain)**2.
```
erykoff, line 285: I think a comment should be added here that the conversion here is necessary because this is a bootstrap version of the variance plane just to be extra sure.

erykoff, line 285: I think a comment should be added here that the conversion here is necessary because this is a bootstrap version of the variance plane just to be extra sure.

tests/test_isrTaskLSST.py
```diff
@@ -423,6 +454,11 @@ def test_isrFlat(self):
         # processing flats, this is a good test of the interpolation.
         isr_config.doDefect = True
 
+        # Need to make sure we are not masking the negative variance
```
erykoff, line 457: I don't think the identical comment is needed on all of these, actually.  Can probably just be on the first one.

tests/test_isrTaskLSST.py
```diff
@@ -477,6 +513,61 @@ def test_isrFlat(self):
...
+                crosstalk=self.crosstalk,
+                ptc=self.ptc,
+                linearizer=self.linearizer,
+            )
+
+        metadata = result.exposure.metadata
+
+        for amp in self.detector:
+            # The overscan noise is always in adu and the readnoise is always
+            # in electron
```
erykoff, line 550: Trailing period here and below.

tests/test_isrTaskLSST.py
```diff
@@ -477,6 +513,61 @@ def test_isrFlat(self):
...
+            # values stored in the PTC
+            self.assertEqual(gain, self.ptc.gain[amp.getName()])
+            self.assertEqual(read_noise, self.ptc.noise[amp.getName()])
+
+            key = f"LSST ISR OVERSCAN RESIDUAL SERIAL STDEV {amp.getName()}"
+            self.assertIn(key, metadata)
+
+            # Determine if the residual serial overscan stddev is consistent
+            # with the PTC readnoise within 3xstandard error
+            serial_overscan_area = amp.getRawHorizontalOverscanBBox().area
```
erykoff, line 564: If we're doing MEDIAN_BY_ROW it'll be noisier than this, but this is a reasonable scaling.

erykoff (commented):
Overall looks good, with some minor formatting.  The largest change is that the PTC updates should be done in `ptcDataset` and not in the tasks.

python/lsst/ip/isr/isrTask.py
Alex-Broughton: Correct, I removed this application of gain

python/lsst/ip/isr/isrTask.py
```diff
@@ -574,7 +574,7 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
     usePtcReadNoise = pexConfig.Field(
         dtype=bool,
         default=False,
-        doc="Use readnoise values from the Photon Transfer Curve?"
+        doc="Use read values from the Photon Transfer Curve?"
```
erykoff, line 577: -> "use read noise values"

python/lsst/ip/isr/isrTask.py
```diff
@@ -2307,7 +2318,7 @@ def darkCorrection(self, exposure, darkExposure, invert=False):
         """
         expScale = exposure.getInfo().getVisitInfo().getDarkTime()
         if math.isnan(expScale):
-            raise RuntimeError("Exposure darktime is NAN.")
+            raise RuntimeError("Exposure darktime isNaN.")
```
erykoff, line 2321: -> "is NaN."

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -693,6 +695,21 @@ def fromTable(cls, tableList):
...
+                # units of electron^2 only if ptcFitType == FULLCOVARIANCE.
+                # After version 1.9, we standardized the
+                # PhotonTransferCurveDataset noise units to electron to fix
+                # this bug. If a user tries to use an earlier version of
+                # PTC with this fit type, we must be sure to do the
+                # calculations properly. More information about this noise
+                # issue can be found in DM-45976.
+                cls().log.info(f"Input PTC VERSION ({calibVersion}) < 1.9 and"
+                               " ptcFitType == FULLCOVARIANCE. Applying fix"
+                               f" for DM-45976 noise issue for AMP {ampName}.")
```
erykoff, line 709: I think that we should put extra logic in here because we don't need 16x this message, only once not per amp.  Or make this a `debug` log.  Up to you.

tests/test_isrTaskLSST.py
```diff
@@ -337,7 +347,14 @@ def test_isrBias(self):
         self.assertLess(np.std(result.exposure.image.array[good_pixels]), 2.0*mock_config.readNoise)
 
         delta = result2.exposure.image.array - result.exposure.image.array
-        self.assertFloatsAlmostEqual(delta[good_pixels], self.bias.image.array[good_pixels], atol=1e-5)
+
+        # Note that the bias is made with bias noise + read noise, and
+        # the image contains read noise.
```
erykoff, line 352: I don't know why this was changed.  The delta here is between the processing where we applied the bias, and where we didn't apply the bias, but with the same read noise pattern.

erykoff (approved):
A couple of minor comments here, but otherwise looks good.

python/lsst/ip/isr/ptcDataset.py
Alex-Broughton: I'll keep it an info log, but I'll only have it output once

PR #344, opened 2024-09-06
DM-46116: Add gains used for crosstalk fit to crosstalk dataset.

erykoff:


python/lsst/ip/isr/crosstalk.py
```diff
@@ -206,6 +211,7 @@ def fromDetector(self, detector, coeffVector=None, coeffSqrVector=None):
         self.coeffValid = np.ones(self.crosstalkShape, dtype=bool)
         self.coeffErrSqr = np.zeros(self.crosstalkShape)
         self.ampGainRatios = np.zeros(self.crosstalkShape)
+        self.fitGains = np.zeros(self.nAmp)
```
czwa, line 214: I don't know if this was an intentional choice, but should this be `np.ones` here instead?  The default casts this to zeros as well, but I can't think of any gain operations that can be performed where zero yields a usable answer.

czwa (approved):
Just the one comment.

python/lsst/ip/isr/crosstalk.py
erykoff: This was intentional in that I wanted zero to mean "this calib did not have a fitGain so don't try to use it".  (And it is not in use anywhere).  But in use I think that would warn and set to 1.0; I could set here to 1.0 and not worry about it downstream if you prefer.

PR #346, opened 2024-09-16
DM-46308: Update PTC dataset to 2.0 and add more fields and methods.

erykoff:
As discovered in development of DM-45856 the PTC dataset can be improved (now version 2.0 because that is the number after 1.9, not for any semantic versioning reasons).

* We should have the `gainList` to go with the `noiseList` which tracks the gain from flat pairs in the combined PTC
* We can now store string-based auxiliary metadata, including observation reason and science program.
* The PTC `sort` and `appendPartialPtc` methods are added so that the PTC solver in `cp_pipe` doesn't need to monkey with internals (and risk missing appending/sorting a field)
* A comment has been added to explicitly list which functions should be updated when a new field is added with a new PTC version.

### DIFF
--- a/python/lsst/ip/isr/ptcDataset.py
+++ b/python/lsst/ip/isr/ptcDataset.py
@@ -874,6 +874,64 @@ class PhotonTransferCurveDataset(IsrCalib):
 
         pass
 
+    def sort(self, sortIndex):
+        """Sort the components of the PTC by a given sort index.
+
+        The PTC is sorted in-place.
+
+        Parameters
+        ----------
+        sortIndex : `list` or `np.ndarray`
+            The sorting index, which must be the same length as
+            the number of elements of the PTC.
+        """
+        index = np.atleast_1d(sortIndex)
+
+        # First confirm everything matches.
+        for ampName in self.ampNames:
+            if len(index) != len(self.rawExpTimes[ampName]):
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


tests/test_ptcDataset.py
```diff
@@ -136,7 +136,12 @@ def _checkTypes(self, ptcDataset):
 
         for key, value in ptcDataset.auxValues.items():
             self.assertIsInstance(value, np.ndarray)
-            self.assertEqual(value.dtype, np.float64)
+            if key == "intVal":
```
Alex-Broughton, line 139: Perhaps make this "INTVAL" for consistency?

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -105,6 +106,9 @@ class PhotonTransferCurveDataset(IsrCalib):
     gainErr : `dict`, [`str`, `float`]
         Dictionary keyed by amp names containing the errors on the
         fitted gains.
+    gainList : `dict`, [`str`, `np.ndarray`]
```
Alex-Broughton, line 109: There is no `gainUnadjusted` added to this doc string. Also, may want to adjust the `gain`description to mention that it is now the amp-offset "adjusted" values, and specify that the `gainUnadjusted` values are the fitted values.

Alex-Broughton (approved):

PR #347, opened 2024-09-17
DM-46366: Fix ISR metadata key inconsistencies between isrTask and isrTaskLSST

Alex-Broughton:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1916,14 +1916,17 @@ def defineEffectivePtc(self, ptcDataset, detector, bfGains, overScans, metadata)
             doWarningPtcValidation = False
 
             metadata[f"LSST GAIN {amp.getName()}"] = effectivePtc.gain[ampName]
+            metadata[f"LSST ISR GAIN {amp.getName()}"] = effectivePtc.gain[ampName]
```
erykoff, line 1919: Probably should reorganize these to pair the old header values (LSST GAIN / LSST READNOISE) with a comment that these are here for compatibility.

erykoff (approved):
Just a minor comment for the old task.

PR #349, opened 2024-09-24
DM-46287: Make image binning a subtask for IsrTask

jrmullaney:


python/lsst/ip/isr/binExposureTask.py
```diff
@@ -0,0 +1,202 @@
...
+                storageClass=self.inputExposure.storageClass,
+                dimensions=frozenset(config.exposureDimensions),
+            )
+            self.binnedExposure = cT.Output(
+                name=self.binnedExposure.name,
+                doc=self.binnedExposure.doc,
+                storageClass=self.binnedExposure.storageClass,
+                dimensions=frozenset(config.exposureDimensions),
+            )
+        if config and config.exposureStorageClass != self.inputExposure.storageClass:
```
czwa, line 80: Do you have any expected cases where the `storageClass` would need to change?

python/lsst/ip/isr/binExposureTask.py
```diff
@@ -0,0 +1,202 @@
...
+    """Perform an nxn binning of an Exposure dataset type.
+
+    The binning factor is the same in both spatial dimensions (i.e.,
+    an nxn binning is performed). Each of the input Exposure's image
+    arrays are binned by the same factor.
+    """
+    ConfigClass = BinExposureConfig
+    _DefaultName = "binExposure"
+
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
```
czwa, line 129: This looks like it's just doing the standard `runQuantum` work, so I think you can drop this (and it'll use the parent class version).

python/lsst/ip/isr/binExposureTask.py
```diff
@@ -0,0 +1,202 @@
...
+               sub-types. The type matches that of the inputExposure).
+        """
+        if not binFactor:
+            binFactor = self.config.binFactor
+        return pipeBase.Struct(
+            binnedExposure=binExposure(inputExposure, binFactor)
+        )
+
+
+def binExposure(inputExposure, binFactor=8):
```
czwa, line 163: Can this method have the `'` switched to `"`?  I'm terrible at consistency myself, but doing it now will save some future clean-up ticket a file.

python/lsst/ip/isr/isrTask.py
```diff
@@ -46,6 +46,7 @@
 from .defects import Defects
 
 from .assembleCcdTask import AssembleCcdTask
+from .binExposureTask import BinExposureTask
```
czwa, line 49: Can I have you make the same changes to `isrTaskLSST.py` as well?  It should be doing the same binning operation here: https://github.com/lsst/ip_isr/blob/main/python/lsst/ip/isr/isrTaskLSST.py#L1851

python/lsst/ip/isr/isrTask.py
```diff
@@ -1777,7 +1783,15 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
         outputBin1Exposure = None
         outputBin2Exposure = None
         if self.config.doBinnedExposures:
-            outputBin1Exposure, outputBin2Exposure = self.makeBinnedImages(ccdExposure)
```
czwa, line 1780: Can you also add a `@deprecated` decorator to the `makeBinnedImages` method for both tasks as well?  I don't think anyone else is using it, but this will at least warn people before it goes away.

python/lsst/ip/isr/binExposureTask.py
czwa: I think this also would cause problems with the check on L193, as that will raise if the `storageClass` isn't an `Exposure`.  This does allow `ExposureD`, so maybe a comment here mentioning that although the `storageClass` can be changed, if it isn't some kind of `Exposure`, it's going to cause problems.

tests/test_binExposure.py
```diff
@@ -0,0 +1,88 @@
...
+            binnedExposure.getMaskedImage(),
+        )
+
+    def test_BinExsposureDataTypes(self):
+        '''Test that a TypeError is raised should the input
+        not be of type `lsst.afw.image.Exposure or one of its sub-types.
+        '''
+        with self.assertRaises(TypeError):
+            _ = binExposure(None)
+        with self.assertRaises(TypeError):
```
czwa, line 75: We may want to allow these to pass in the future, but I'm fine with not allowing `Image` or `MaskedImage` alone for now.

czwa (approved):
A few suggested changes, but I'm happy here.  It would be good to have the `pipe_tasks` task call this code as well, but that can be on a follow-up ticket (I'll file that now).

python/lsst/ip/isr/binExposureTask.py
jrmullaney: I added this to ensure that the output storage class would always match the input storage class. While I could have the input and output storage classes as the generic `Exposure`, and it would be happy with any of the sub-types as input, any subtype information (I,F,D etc) would be lost on the output (at least that's my understanding).

I can certainly add the comment.

tests/test_binExposure.py
jrmullaney: I'll make a ticket and add a comment to say that Image and MaskedImage binning will be added with that ticket. Sound ok?

PR #354, opened 2024-10-15
DM-46509: Add FULLCOVRIANCE_NO_B fit option to PTC solver

Alex-Broughton:


erykoff (changes requested):
If we read an old PTC, we still want to fill in the noB parameters so they are available to the user, but any access will issue a deprecation warning.  So I think this went too far in the code removal

erykoff:
My comment above is still a problem.  If the file has these columns we want to read them in; when that is done it shouldn't issue a deprecation warning, but if you try to access them it will (but will have values!)

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -811,6 +817,15 @@ def fromTable(cls, tableList):
...
                 inDict['gainList'][ampName] = record['GAIN_LIST']
+            if calibVersion < 2.1:
+                inDict['covariancesModelNoB'][ampName] = record['COVARIANCES_MODEL']
+                inDict['aMatrixNoB'][ampName] = record['A_MATRIX']
+            else:
+                inDict['covariancesModelNoB'][ampName] = np.full_like(
+                    inDict['covariancesModel'][ampName],
+                    np.nan,
+                )
+                inDict['aMatrixNoB'][ampName] = np.full_like(inDict['aMatrix'][ampName], np.nan)
```
erykoff, line 828: I don't know if this else is necessary.  For newer PTCs we shouldn't be using these fields at all, right?

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -1013,9 +1025,9 @@ def appendPartialPtc(self, partialPtc):
                     self.covMatrixSide,
                 )
             )
-            self.covariancesModelNoB[ampName] = np.append(
-                self.covariancesModelNoB[ampName].ravel(),
-                partialPtc.covariancesModelNoB[ampName].ravel()
+            self._covariancesModelNoB[ampName] = np.append(
+                self._covariancesModelNoB[ampName].ravel(),
+                partialPtc._covariancesModelNoB[ampName].ravel()
```
erykoff, line 1030: We aren't using this at all; I think if we need filler this should be explicitly filled with nans.

erykoff (commented):
A couple more comments.  Also I want to make sure that when you run tests that no warnings from this deprecation are emitted, either here or on cp_pipe.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -1013,16 +1019,16 @@ def appendPartialPtc(self, partialPtc):
...
+            # self._covariancesModelNoB[ampName] = np.append(
+            #     self._covariancesModelNoB[ampName].ravel(),
+            #     partialPtc._covariancesModelNoB[ampName].ravel()
+            # ).reshape(
+            #     (
+            #         len(self.rawExpTimes[ampName]),
+            #         self.covMatrixSide,
+            #         self.covMatrixSide,
+            #     )
+            # )
```
erykoff, line 1031: Remove this?

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -413,18 +417,18 @@ def setAmpValuesPartialDataset(
...
-
         # Filler values.
         self.finalVars[ampName] = np.array([np.nan])
         self.finalModelVars[ampName] = np.array([np.nan])
         self.finalMeans[ampName] = np.array([np.nan])
 
+        self._covariancesModelNoB[ampName] = np.array([nanMatrixFit])
+        self._aMatrixNoB[ampName] = nanMatrixFit
+        self._noiseMatrixNoB[ampName] = nanMatrixFit
+
```
erykoff, line 431: This needs the TODO ticket link

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -191,15 +193,16 @@ class PhotonTransferCurveDataset(IsrCalib):
         the model in Eq. 20 of Astier+19 (units: electron^2).
     covariancesModelNoB : `dict`, [`str`, `np.ndarray`]
         Dictionary keyed by amp names containing covariances model
-        (with 'b'=0 in Eq. 20 of Astier+19)
-        per mean flux (units: adu^2).
+        (with 'b'=0 in Eq. 20 of Astier+19) per mean flux (units:
+        adu^2). Will be deprecated in v28.
```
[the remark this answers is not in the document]
TallJimbo: I think our deprecation messages typically only say "after", not "before", and if this deprecation is going in now then removal is "after v29."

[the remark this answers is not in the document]
erykoff: This still needs to be fixed!

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -701,15 +705,15 @@ def fromTable(cls, tableList):
...
-        inDict['noiseMatrixNoB'] = dict()
         inDict['finalVars'] = dict()
         inDict['finalModelVars'] = dict()
         inDict['finalMeans'] = dict()
         inDict['badAmps'] = []
         inDict['photoCharges'] = dict()
         inDict['ampOffsets'] = dict()
+        inDict['noiseMatrixNoB'] = dict()
+        inDict['covariancesModelNoB'] = dict()
+        inDict['aMatrixNoB'] = dict()
```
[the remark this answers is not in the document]
Alex-Broughton: Hang on... Jenkins is failing when I take these out. I think we still need these to read in older versions of the PTC.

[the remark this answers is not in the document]
erykoff: You either need an `else` after line 822 (current version), filling with nans, or an else after line 566 filling with nans.  Second one is what I recommend as it's more general.

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -191,15 +193,16 @@ class PhotonTransferCurveDataset(IsrCalib):
...
-        (with 'b'=0 in Eq. 20 of Astier+19)
-        per mean flux (units: adu^2).
+        (with 'b'=0 in Eq. 20 of Astier+19) per mean flux (units:
+        adu^2). Will be deprecated after v29.
     aMatrixNoB : `dict`, [`str`, `np.ndarray`]
         Dictionary keyed by amp names containing the "a" parameters from the
-        model in Eq. 20 of Astier+19
-        (and 'b' = 0) (units: 1/electron).
+        model in Eq. 20 of Astier+19 (and 'b' = 0) (units: 1/electron).
+        Will be deprecated after v29.
```
erykoff, line 201: Will be  *removed* (and the others)

erykoff (approved):

PR #355, opened 2024-10-17
DM-46925: Fix fitGains on load when stored incorrectly.

erykoff:


czwa (approved):
That looks reasonable to me.

PR #357, opened 2024-10-29
DM-47195: Fix accidental saturation mask transfer when using non-linear crosstalk terms.

czwa:


tests/test_crosstalk.py
```diff
@@ -282,7 +285,8 @@ def testTaskAPI_NL(self):
         correction.
         """
         for this_isr_task in [IsrTask, IsrTaskLSST]:
-            self.checkTaskAPI_NL(this_isr_task)
+            for subtrahendMasking in [False, True]:
+                self.checkTaskAPI_NL(this_isr_task, subtrahendMasking)
```
mrawls, line 289: Does `checkSubtracted` do enough here? In addition to adding another call to this test, would it be possible to add something that checks the mask planes are being cleared properly along the way (i.e., the original bug that prompted all this)?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -563,7 +563,7 @@ def calculateBackground(mi, badPixels=["BAD"]):
 
     def subtractCrosstalk(self, thisExposure, sourceExposure=None, crosstalkCoeffs=None,
                           crosstalkCoeffsSqr=None, crosstalkCoeffsValid=None,
-                          badPixels=["BAD"], minPixelToMask=45000,
+                          badPixels=["BAD"], minPixelToMask=45000, doSubtrahendMasking=False,
```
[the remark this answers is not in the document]
czwa: Due to choices (I'm not sure of their direction), this method is attached to the calibration object.  The task has the config options to set this parameter.

tests/test_crosstalk.py
czwa: Added a region that is declared "saturated" with matching test to confirm that masks aren't being duplicated.  I can confirm that running the test from this ticket with `ip_isr` on main throws a `AssertionError: 100 != 25` (25 pixels for this saturated region, x4 due to that mask being duplicated across all four amplifiers).

mrawls (approved):

PR #358, opened 2024-10-29
DM-47263: Simplify amp offset correction logging

leeskelvin:


python/lsst/ip/isr/ampOffset.py
```diff
@@ -420,11 +436,59 @@ def getAmpOffsets(self, im, amps, associations, sides):
...
+                    interfaceOffsetOriginals.append(interfaceOffsetOriginal)
+                    ampEdgeGoodFracs.append(ampEdgeGoodFrac)
+                    minFracFails.append(minFracFail)
+                    maxOffsetFails.append(maxOffsetFail)
+                    interfaceOffsetLookup[f"{ampId:02d}:{ampNeighbor:02d}"] = interfaceOffset
                 else:
-                    interfaceOffset = -interfaceOffsetLookup[f"{ampNeighbor}{ampId}"]
+                    interfaceOffset = -interfaceOffsetLookup[f"{ampNeighbor:02d}:{ampId:02d}"]
                 ampsOffsets[ampId] += interfaceWeight * interfaceOffset
+        quartile_summary = np.percentile(interfaceOffsetOriginals, [0, 25, 50, 75, 100])
```
erykoff, line 456: I think you might need a check here to make sure numbers were measured or else this is going to crash on an empty input.  (I don't know if it is possible, but I imagine it might be in the limit of one working amp?)

python/lsst/ip/isr/ampOffset.py
```diff
@@ -474,16 +538,16 @@ def getAmpEdges(self, im, amps, ampSides):
...
         """Calculate the amp offset for a given interface between two
         amplifiers.
 
         Parameters
         ----------
-        ampIdA : int
-            ID of the first amplifier.
-        ampIdB : int
-            ID of the second amplifier.
+        ampNameA : int
```
erykoff, line 547: This is not an int anymore.

python/lsst/ip/isr/ampOffset.py
```diff
@@ -474,16 +538,16 @@ def getAmpEdges(self, im, amps, ampSides):
...
+            Name of the first amplifier.
+        ampNameB : int
```
erykoff, line 549: Neither is this.

erykoff (approved):
Minor comments.

PR #360, opened 2024-11-06
DM-45928: add parallel CTI and add serial/parallel turnoff calculations to DeferredChargeCalib

Alex-Broughton:


python/lsst/ip/isr/deferredCharge.py
```diff
@@ -904,6 +963,41 @@ def fromTable(cls, tableList):
...
+            inDict['parallelCtiTurnoffSamplingErr'] = {amp: np.nan for amp in amps}
+        else:
+            signals = ampTable['SIGNALS']
+            serialEper = ampTable['SERIAL_EPER']
+            parallelEper = ampTable['PARALLEL_EPER']
+            serialCtiTurnoff = ampTable['SERIAL_CTI_TURNOFF']
+            parallelCtiTurnoff = ampTable['PARALLEL_CTI_TURNOFF']
+            serialCtiTurnoffSamplingErr = ampTable['SERIAL_CTI_TURNOFF_SAMPLING_ERR']
+            parallelCtiTurnoffSamplingErr = ampTable['PARALLEL_CTI_TURNOFF_SAMPLING_ERR']
+            inDict['signals'] = {amp: value for amp, value in zip(amps, signals)}
```
czwa, line 989: Another "it must work if the tests pass", but: does this round-trip correctly?

python/lsst/ip/isr/deferredCharge.py
```diff
@@ -904,6 +963,41 @@ def fromTable(cls, tableList):
...
         inDict['globalCti'] = {amp: value for amp, value in zip(amps, globalCti)}
 
+        # Version check
+        if calibVersion < 1.1:
+            # This version might be in the wrong units (not electron),
+            # and does not contain the gain information to convert
+            # into a new calibration version.
+            raise RuntimeError(f"Using old version of CTI calibration (ver. {calibVersion} < 1.1), "
+                               "which is no longer supported.")
+        if calibVersion < 1.2:
```
czwa, line 973: `elif` here?

tests/test_deferredCharge.py
```diff
@@ -98,6 +98,20 @@ def setUp(self):
...
+        self.calib.parallelEper['amp0'] = np.full_like(self.calib.signals['amp0'], 1.0e-6)
+        self.calib.parallelEper['amp1'] = np.full_like(self.calib.signals['amp1'], 1.0e-6)
+        self.calib.serialCtiTurnoff['amp0'] = 1.0e5
+        self.calib.serialCtiTurnoff['amp1'] = 1.0e5
+        self.calib.parallelCtiTurnoff['amp0'] = 1.0e5
+        self.calib.parallelCtiTurnoff['amp1'] = 1.0e5
+        self.calib.serialCtiTurnoffSamplingErr['amp0'] = 1.0e3
+        self.calib.serialCtiTurnoffSamplingErr['amp1'] = 1.0e3
+        self.calib.parallelCtiTurnoffSamplingErr['amp0'] = 1.0e3
+        self.calib.parallelCtiTurnoffSamplingErr['amp1'] = 1.0e3
```
czwa, line 114: After reading through the rest, I think this probably should be fixed here (so `amp1` gets a `logistic` trap).  Adding a `testPartiallyPopulated` might be good to add to test that `serialTraps` can be not populated for `amp1` to keep the accidental test case intact.

czwa (approved):
Minor comments.

czwa:
Talking with @erykoff let me know that this is going back for more development, so feel free to ignore my comments as needed.

python/lsst/ip/isr/deferredCharge.py
```diff
@@ -745,6 +745,29 @@ class DeferredChargeCalib(IsrCalib):
     serialTraps : `dict` [`str`, `lsst.ip.isr.SerialTrap`]
         A dictionary, keyed by amplifier name, containing a single
         serial trap for each amplifier.
+    signals : `dict` [`str`, `np.ndarray`, `float`]
+        A dictionary, keyed by amplifier name, of the mean signal
+        level for each input measurement.
+    serialEper : `dict` [`str`, `np.ndarray`, `float`]
+        A dictionary, keyed by amplifier name, of the serial EPER
```
erykoff, line 752: Can you spell out EPER at least once in the docstring?

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -225,7 +225,8 @@ def __init__(self, statControl=None, **kwargs):
                                                      afwImage.Mask.getPlaneBitMask(self.config.badMask))
         self.statType = afwMath.stringToStatisticsProperty(self.config.stat)
 
-    def run(self, inputExp, ptc=None, overscanResults=None, **kwargs):
+    def run(self, inputExp, untrimmedInputExposure=None, ptc=None, serialOverscanResults=None,
+            parallelOverscanResults=None, **kwargs):
```
erykoff, line 229: Do we use the parallelOverscanResults anywhere?  Is this just adding this in case we need it in the future?

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -350,90 +354,136 @@ def measureCti(self, inputExp, overscans, gains):
         """
         outputStats = {}
 
+        # Debug check
+        assert isTrimmedExposure(inputExp)
+        assert not isTrimmedExposure(untrimmedInputExp)
+
```
erykoff, line 360: Please remove debug checks.

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -350,90 +354,136 @@ def measureCti(self, inputExp, overscans, gains):
...
+        with gainContext(inputExp, image, applyGainToImage, gains, isTrimmed=True):
+            with gainContext(untrimmedInputExp, untrimmedImage, applyGainToUntrimmedImage,
+                             gains, isTrimmed=False):
+                for amp in detector.getAmplifiers():
+                    ampStats = {}
+                    readoutCorner = amp.getReadoutCorner()
+
+                    # Full data region.
+                    dataRegion = image[amp.getBBox()]
```
erykoff, line 382: I don't think this is correct (or could be wrong depending on configs).  The `image` in this case (potentially) has bias/dark/flat/bfk/etc applied (though this is a pipeline choice).  Can't you just use the untrimmed image exclusively and ignore the trimmed image for the CTI statistics?  The untrimmed image has the same operations guaranteed to be applied to both the overscan and imaging regions, and therefore will be matched.  Plus, I think it would make the code simpler.

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -350,90 +354,136 @@ def measureCti(self, inputExp, overscans, gains):
...
+                        ampStats["FIRST_ROW_MEAN"] = rowZ
+                        ampStats["LAST_ROW_MEAN"] = rowA
+                    else:
+                        ampStats["FIRST_ROW_MEAN"] = rowA
+                        ampStats["LAST_ROW_MEAN"] = rowZ
+
+                    # Measure the columns of the serial overscan and the rows
+                    # of the parallel overscan.
+                    nSerialOverscanCols = amp.getRawSerialOverscanBBox().getWidth()
+                    nParallelOverscanRows = amp.getRawParallelOverscanBBox().getHeight()
```
erykoff, line 435: Note that these are only going to work with an "LSST style" camera, which is fine, that's what we support.

erykoff: (The problem is that a different camera could have a different row/col orientation and width/height would therefore be reversed).

python/lsst/ip/isr/isrTask.py
```diff
@@ -1783,7 +1783,8 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
         # calculate additional statistics.
         outputStatistics = None
         if self.config.doCalculateStatistics:
-            outputStatistics = self.isrStats.run(ccdExposure, overscanResults=overscans,
+            outputStatistics = self.isrStats.run(ccdExposure, serialOverscanResults=overscans,
+                                                 parallelOverscanResults=[None for _ in overscans],
```
erykoff, line 1787: The parallelOverscanResults could/should just be None here (but as asked above this seems to be unused anyway).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1785,6 +1778,11 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
                 gains=gains,
             )
 
+        # Save the untrimmed version for later statistics,
+        # which still contains the overscan information
+        if self.config.doCalculateStatistics:
```
erykoff, line 1783: Is this the right check?  It's not if we're calculating statistics at all, it's if we're computing CTI statistics.  As this is written we're going to be doing an expensive clone operation for regular ISR which we do not want.  So we want to check if we're doing cti statistics, `self.config.isrStatistics.doCtiStatistics` (I think?) and if not set `untrimmedCcdExposure = None` to protect it below.

python/lsst/ip/isr/deferredCharge.py
Alex-Broughton: Can you clarify what you mean by round-trip?

erykoff: There does appear to be a test (but is it too trivial?) that you can serialize to fits and read it back.

python/lsst/ip/isr/isrStatistics.py
Alex-Broughton: No, not yet. I wasn't sure if we wanted to include parallel CTI statistics in the ISR statistics output dataset in the future.

python/lsst/ip/isr/isrStatistics.py
Alex-Broughton: We could, but the defect mask would not be applied to the untrimmed image, so we would calculate CTI statistics on an image with potentially large defects (that might affect the last imaging pixels or overscan). When I calculate the statistics I use median of the last imaging column (serial) or row (parallel), so it might be pretty robust against not masking defects. I can go ahead and simplify it, but we should check the calibration or add a test to cp_verify in the future that can make sure there are no defects impacting the CTI solver.

erykoff: Ah!  That is a great point, you are correct.  This should be made clear in the code comments, and we need to be extra careful about our pipeline settings.

czwa (changes requested):
I think I'm generally happy, except for the serial trap handling in `test_deferredCharge.py`.

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -251,6 +268,11 @@ def run(self, inputExp, ptc=None, overscanResults=None, **kwargs):
                 correction applied (`lsst.afw.image.Image`). This
                 quantity is used to estimate the amplifier read noise
                 empirically.
+        doLegacyCtiStatistics : `bool`, optional
+            Use the older version of measureCti (not recommended).
+            This should be True if and only if this task is called
+            from IsrTask. TODO: Deprecate legacy CTI + CTI correction
+            from IsrTask.
```
erykoff, line 275: Can you file this deprecate ticket and link it here?  It would need an RFC to be done, but we can file the plan now.

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -317,15 +348,164 @@ def run(self, inputExp, ptc=None, overscanResults=None, **kwargs):
...
         )
 
-    def measureCti(self, inputExp, overscans, gains):
+    def measureCti(self, inputExp, untrimmedInputExp, gains):
+        """Task to measure CTI statistics.
+
+        Parameters
+        ----------
+        inputExp : `lsst.afw.image.Exposure`
+            The exposure to measure.
```
erykoff, line 357: Can you add a comment here that the full exposure is needed because the data region has the defect mask?

python/lsst/ip/isr/isrStatistics.py
```diff
@@ -348,6 +528,8 @@ def measureCti(self, inputExp, overscans, gains):
             Dictionary of measurements, keyed by amplifier name and
             statistics segment. Everything in units based on electron.
         """
+        self.log.warning("Using the legacy version of CTI statistics may give wrong CTI")
```
erykoff, line 531: Do we need a warning here and above?  (Though heck why not, nobody should be running this...)

erykoff (approved):
A couple of remaining small documentation requests; otherwise looks great!

czwa (approved):

PR #361, opened 2024-11-07
DM-47449: Replace afw akima interpolator with scipy for speed.

erykoff:


czwa (approved):
Is there a reason to keep the AFW interpolator in tests/mocks?  I'm fine with it, but wanted to ask.

erykoff:
I wanted to leave it there as a cheap way of testing consistency.  I thought I put this in the comments.  Anyway, I deliberately said "use afw to alter the image and scipy to correct it" and everything works fine as long as you stay away from extrapolation (which we should).

PR #364, opened 2024-11-20
DM-47636: Switch CROSSTALK masking to use subtrahend masking

czwa:


erykoff (approved):
This is good, but I think that the docstring to `subtractCrosstalk` also needs to be updated to (a) add `doSubtrahendMasking` and explain it; (b) explain that `minPixelToMask` has a different meaning whether or not `doSubtrahendMasking` is `True` or `False`.  Also I notice that the config docstring for `minPixelToMask` at https://github.com/lsst/ip_isr/blob/main/python/lsst/ip/isr/crosstalk.py#L811 needs to be updated.

PR #367, opened 2024-12-08
DM-48024: Modify amp offset metadata setter

leeskelvin:


leeskelvin:
This looks fine. Please squash the commits down into just 1 commit though.

PR #369, opened 2025-01-13
DM-48373: Refactor and optimize crosstalk correction task.

erykoff:


python/lsst/ip/isr/crosstalk.py
```diff
@@ -659,139 +686,171 @@ def subtractCrosstalk(self, thisExposure, sourceExposure=None, crosstalkCoeffs=N
...
+        crosstalk = sourceMaskedImage.mask.getPlaneBitMask(crosstalkStr)
 
-        # Define a subtrahend image to contain all the scaled crosstalk signals
-        subtrahend = source.Factory(source.getBBox())
+        # Define a subtrahend image to contain all the scaled crosstalk
+        # signals. These will be applied to the target image.
+        subtrahend = targetMaskedImage.Factory(targetMaskedImage.getBBox())
         subtrahend.set((0, 0, 0))
 
-        coeffs = coeffs.transpose()
```
czwa, line 738: I know the ticket mentions that things are pixel identical, so I'm guessing this transpose is no longer needed because of the naming swaps below?

python/lsst/ip/isr/crosstalk.py
```diff
@@ -659,139 +686,171 @@ def subtractCrosstalk(self, thisExposure, sourceExposure=None, crosstalkCoeffs=N
...
-                        mi2,
-                        tAmp,
-                        sAmp,
+
+                targetBBox = self._getAppropriateBBox(targetAmp, isTrimmed, fullAmplifier)
+
+                # The extractAmp() method will always make a copy of the source
+                # amplifier data.
+                if imageOnly:
+                    sourceImage = self.extractAmp(
```
czwa, line 816: Does `extractAmp` need to be inside this if/else?  This also seems like it can be merged with the subsequent `if not imageOnly` block.

czwa (approved):
Took me a bit of time to think through everything, but this looks right (one question, one very optional style suggestion).

python/lsst/ip/isr/crosstalk.py
erykoff: In some sense, the transpose was never needed because you could always change `[index1, index2]` to `[index2, index1]`.  But I think it may be that the `s` and `t` confusion led to the addition of this in the first place.

python/lsst/ip/isr/crosstalk.py
erykoff: Partly this is because of the way I wrote the code originally, but the option is to either make a temporary variable that is sent to `extractAmp` or do it this way.  I think it's a wash.

As for whether the following `if not imageOnly` is needed ... no it is not.

PR #373, opened 2025-01-29
DM-45545: Fixes for numpy 2

timj:


tests/test_variance_plane.py
```diff
@@ -333,7 +333,7 @@ def setUp(self):
 
         # Set the random seed for reproducibility.
         random_seed = galsim.BaseDeviate(1905).raw() + 1
-        np.random.seed(random_seed)
+        self.np_rng = np.random.Generator(np.random.MT19937(4))
```
timj, line 336: I did this as a side project but couldn't find a seed that would work exactly like before and so ended up slightly relaxing the test tolerance.

tests/test_ampOffset.py
```diff
@@ -351,7 +351,7 @@ def runAmpOffsetWithBackground(self, valueType, rampBackground=False):
             self.assertAlmostEqual(
                 np.std(pedestals - approximatePedestals),
                 measuredSigma[weightType][valueType],
-                12,
+                4,  # DO NOT MERGE
```
timj, line 354: @enourbakhsh to confirm, you are okay with this being only 4 decimal places?

enourbakhsh: Do we know if it’s just one number causing it or if all of them are affected? I’m just curious, could there be any type conversion happening somewhere that’s introducing precision loss?

timj: ```tests/test_ampOffset.py::AmpOffsetTest::testAmpOffsetWithRampBackground - AssertionError: np.float64(6.08989306958048) != 6.089900455983161 within 8 places (np.float64(7.386402680609194e-06) difference)```

timj: so it's weighted artificial that is changing.

enourbakhsh: I'd make the number of decimal places conditional on the `rampBackground` parameter because the approximations are less reliable when using ramp backgrounds. Does this fix it?
```python
atol = 4 if rampBackground else 12
```

timj:
```diff
diff --git a/tests/test_ampOffset.py b/tests/test_ampOffset.py
index 4303b97..c405ae3 100644
--- a/tests/test_ampOffset.py
+++ b/tests/test_ampOffset.py
@@ -351,7 +351,7 @@ class AmpOffsetTest(lsst.utils.tests.TestCase):
             self.assertAlmostEqual(
                 np.std(pedestals - approximatePedestals),
                 measuredSigma[weightType][valueType],
-                4,  # DO NOT MERGE
+                4 if rampBackground else 12,
             )
             if valueType == "artificial":
                 if not applyWeights:
```
does work.

enourbakhsh: Great! I'd be happy with this change.

timj: Changed. I can merge once the full PR is reviewed.

erykoff (approved):

PR #375, opened 2025-02-05
DM-11270: Please don't warn about zero points for exposure times of 0

czwa:
I've demoted the message in this case to debug, so it can be caught if needed.
This doesn't trigger in `ci_cpp_gen3`, because this is only called in the old `IsrTask` code (`ci_cpp_gen3` has switched to the LSST specific version).


timj (approved):
I did review on Jira.

PR #376, opened 2025-02-11
DM-48878: Set parallelOverscans=None default in IsrTaskLSST

Alex-Broughton:
Fix a bug caused by DM-45928

### DIFF
--- a/tests/test_isrTaskLSST.py
+++ b/tests/test_isrTaskLSST.py
@@ -400,6 +400,77 @@ class IsrTaskLSSTTestCase(lsst.utils.tests.TestCase):
 
         self._check_bad_column_crosstalk_correction(result.exposure)
 
+    def test_isrBiasNoParallelOscanCorrection(self):
+        """Test processing of a bias frame with parallel
+        overscan correction turned off."""
+        mock_config = self.get_mock_config_no_signal()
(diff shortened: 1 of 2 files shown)


erykoff (commented):
Can we test this as well?  Add a test to `test_isrTaskLSST.py` that just runs without parallel overscan.  It can be a bias test, doesn't even need to look good (unless you turn off the parallel overscan slope).

tests/test_isrTaskLSST.py
```diff
@@ -400,6 +400,77 @@ def test_isrBias(self):
...
+        mock_config = self.get_mock_config_no_signal()
+        mock_config.doAddParallelOverscanRamp = False
+
+        mock = isrMockLSST.IsrMockLSST(config=mock_config)
+        input_exp = mock.run()
+
+        isr_config = self.get_isr_config_electronic_corrections()
+
+        # Turn off the parallel overscan correction
+        ampConfig = isr_config.overscanCamera.getOverscanDetectorConfig(self.detector).defaultAmpConfig
```
erykoff, line 415: Small comment: this should be `amp_config` (snake_case) to match the style of the python file.

erykoff (approved):

PR #378, opened 2025-02-12
DM-47681: ITL edge bleed masking

aferte:
The code is linted and passes through ISR tests successfully but the jenkins run failed, I need to look at why.

### DIFF
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -217,29 +217,32 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
         fpSet.setMask(mask, maskValue)
 
 
-def maskITLEdgeBleed(ccdExposure, itlEdgeBleedSatMinArea,
-                     itlEdgeBleedSatFracLevel, itlEdgeBleedModelConstant, saturatedMaskName, badAmpDict):
+def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
+                     itlEdgeBleedSatFracLevel=0.8,
+                     itlEdgeBleedModelConstant=0.03,
+                     saturatedMaskName="SAT"):
     """Mask edge bleeds in ITL detectors.
 
     Parameters
     ----------
     ccdExposure : `lsst.afw.image.Exposure`
         Exposure to apply masking to.
-    itlEdgeBleedSatMinArea : scalar
-        Minimal saturated footprint area where the presence of edge bleeds
-        will be checked.
-    itlEdgeBleedSatFracLevel : scalar
-       Fraction of the saturation level at the detector edge
-       above which there is an edge bleed.
-    itlEdgeBleedModelConstant : scalar
-        Constant in the decaying exponential in the edge bleed masking.
-    saturatedMaskName : `str`
-        Mask name for saturation.
-    badAmpDict : `str`[`bool`]
+    badAmpDict : `dict` [`str`, `bool`]
         Dictionary of amplifiers, keyed by name, value is True if
         amplifier is fully masked.
+    itlEdgeBleedSatMinArea : `int`, optional
+        Minimal saturated footprint area where the presence of edge bleeds
+        will be checked.
+    itlEdgeBleedSatFracLevel : `float`, optional
+       Fraction of the saturation level at the detector edge
+       above which there is an edge bleed.
+    itlEdgeBleedModelConstant : `float`, optional
+        Constant in the decaying exponential in the edge bleed masking.
+    saturatedMaskName : `str`, optional
+        Mask name for saturation.
     """
     maskedImage = ccdExposure.maskedImage
+    xmax = maskedImage.image.array.shape[1]
 
     # Get minimum of amplifier saturation level
     satLevel = numpy.min([ccdExposure.metadata[f"LSST ISR SATURATION LEVEL {amp.getName()}"]
@@ -252,71 +255,96 @@ def maskITLEdgeBleed(ccdExposure, itlEdgeBleedSatMinArea,
     satAreas = numpy.asarray([fp.getArea() for fp in fpList])
     largeAreas, = numpy.where(satAreas >= itlEdgeBleedSatMinArea)
 
+    satMaskBit = maskedImage.mask.getPlaneBitMask(saturatedMaskName)
+
     if len(largeAreas) == 0:
         return
-    else:
-        for largeAreasIndex in largeAreas:
 
-            # Get centroid of saturated core
-            xCore, yCore = fpList[largeAreasIndex].getCentroid()
-            # Turn the Y detector coordinate
-            # into Y footprint coordinate
-            yCoreFP = yCore - fpList[largeAreasIndex].getBBox().getMin().getY()
-            # Get the number of saturated columns around the centroid
-            widthSat = numpy.sum(fpList[largeAreasIndex].getSpans().asArray()[int(yCoreFP), :])
+    for largeAreasIndex in largeAreas:
 
-            for amp in ccdExposure.getDetector():
-                # Select 2 amplifiers around the saturated
-                # core with a potential edge bleed
-                yBox = amp.getBBox().getCenter()[1]
-                if amp.getBBox().contains(xCore, yBox):
+        # Get centroid of saturated core
+        xCore, yCore = fpList[largeAreasIndex].getCentroid()
+        # Turn the Y detector coordinate
+        # into Y footprint coordinate
+        yCoreFP = yCore - fpList[largeAreasIndex].getBBox().getMinY()
+        # Get the number of saturated columns around the centroid
+        widthSat = numpy.sum(fpList[largeAreasIndex].getSpans().asArray()[int(yCoreFP), :])
 
-                    # Get the amp name
-                    ampName = amp.getName()
+        for amp in ccdExposure.getDetector():
+            # Select the 2 top and bottom amplifiers around the saturated
+            # core with a potential edge bleed by selecting the amplifiers
+            # that have the same X coordinate as the saturated core.
+            # As we don't care about the Y coordinate, we set it to the
+            # center of the BBox.
+            yBox = amp.getBBox().getCenter()[1]
+            if amp.getBBox().contains(xCore, yBox):
 
-                    # Because in ITLs the edge bleed happens on both edges
-                    # of the detector, we make a cutout around
-                    # both the top and bottom
-                    # edge bleed candidates around the saturated core.
-                    # We flip the cutout of the top amplifier
-                    # to then work with the same coordinates for both.
-                    # The way of selecting top vs bottom amp
-                    # is very specific to ITL.
-                    if ampName[:2] == 'C1':
-                        sliceImage = maskedImage.image.array[:200, :]
-                        sliceMask = maskedImage.mask.array[:200, :]
-                    elif ampName[:2] == 'C0':
-                        sliceImage = numpy.flipud(maskedImage.image.array[-200:, :])
-                        sliceMask = numpy.flipud(maskedImage.mask.array[-200:, :])
+                # Get the amp name
+                ampName = amp.getName()
 
-                    # The middle columns of edge bleeds are often quite close
-                    # to the saturation level so
-                    # we check there is an edge bleed by looking
-                    # at a small image up to 50 pixels from the edge
-                    # and around the saturated columns
-                    # of the saturated core, and checking its mean is
-                    # above 80 percent of the saturation level.
-                    if numpy.mean(sliceImage[:50, int(xCore)-5:int(xCore)+5]) \
-                            > itlEdgeBleedSatFracLevel*satLevel:
+                # Because in ITLs the edge bleed happens on both edges
+                # of the detector, we make a cutout around
+                # both the top and bottom
+                # edge bleed candidates around the saturated core.
+                # We flip the cutout of the top amplifier
+                # to then work with the same coordinates for both.
+                # The way of selecting top vs bottom amp
+                # is very specific to ITL.
+                if ampName[:2] == 'C1':
+                    sliceImage = maskedImage.image.array[:200, :]
+                    sliceMask = maskedImage.mask.array[:200, :]
+                elif ampName[:2] == 'C0':
+                    sliceImage = numpy.flipud(maskedImage.image.array[-200:, :])
+                    sliceMask = numpy.flipud(maskedImage.mask.array[-200:, :])
 
-                        # We need an estimate of the maximum width
-                        # of the edge bleed for our masking model
-                        # so we now estimate it by measuring the width of
-                        # areas above 60 percent of the saturation level
-                        # close to the edge,
-                        # in a cutout up to 100 pixels from the edge.
-                        subImage = sliceImage[:100, :]
-                        maxWidthEdgeBleed = numpy.max(numpy.sum(subImage > 0.6*satLevel,
-                                                                axis=1))
+                # The middle columns of edge bleeds are often quite close
+                # to the saturation level so
+                # we check there is an edge bleed by looking
+                # at a small image up to 50 pixels from the edge
+                # and around the saturated columns
+                # of the saturated core, and checking its mean is
+                # above 80 percent (which is set as default) of the
+                # saturation level.
 
-                        # Mask edge bleed with a
-                        # decaying exponential model
-                        for y in range(200):
-                            edgeBleedHalfWidth = \
-                                int(((maxWidthEdgeBleed)*numpy.exp(-itlEdgeBleedModelConstant*y)
-                                     + widthSat)/2.)
-                            sliceMask[y, int(xCore)-edgeBleedHalfWidth:int(xCore)+edgeBleedHalfWidth] = \
-                                maskedImage.mask.getPlaneBitMask(saturatedMaskName)
+                # If the centroid is too close to the edge of the detector
+                # (within 5 pixels), we set the limit to the mean check
+                # to the edge of the detector
+                lowerRangeSmall = int(xCore)-5
+                upperRangeSmall = int(xCore)+5
+                if lowerRangeSmall < 0:
+                    lowerRangeSmall = 0
+                if upperRangeSmall > xmax:
+                    upperRangeSmall = xmax
+                if numpy.mean(sliceImage[:50,
+                                         lowerRangeSmall:upperRangeSmall]) \
+                        > itlEdgeBleedSatFracLevel*satLevel:
+
+                    # We need an estimate of the maximum width
+                    # of the edge bleed for our masking model
+                    # so we now estimate it by measuring the width of
+                    # areas above 60 percent of the saturation level
+                    # close to the edge,
+                    # in a cutout up to 100 pixels from the edge.
+                    subImage = sliceImage[:100, :]
+                    maxWidthEdgeBleed = numpy.max(numpy.sum(subImage > 0.6*satLevel,
+                                                            axis=1))
+
+                    # Mask edge bleed with a
+                    # decaying exponential model
+                    for y in range(200):
+                        edgeBleedHalfWidth = \
+                            int(((maxWidthEdgeBleed)*numpy.exp(-itlEdgeBleedModelConstant*y)
+                                 + widthSat)/2.)
+                        lowerRange = int(xCore)-edgeBleedHalfWidth
+                        upperRange = int(xCore)+edgeBleedHalfWidth
+                        # If the edge bleed model goes outside the detector
+                        # we set the limit for the masking
+                        # to the edge of the detector
+                        if lowerRange < 0:
+                            lowerRange = 0
+                        if upperRange > xmax:
+                            upperRange = xmax
+                        sliceMask[y, lowerRange:upperRange] = satMaskBit
 
 
 def interpolateFromMask(maskedImage, fwhm, growSaturatedFootprints=1,
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -218,6 +218,7 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
 
 
 def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
+                     itlEdgeBleedSatMaxArea=100000,
                      itlEdgeBleedSatFracLevel=0.8,
                      itlEdgeBleedModelConstant=0.03,
                      saturatedMaskName="SAT"):
@@ -253,7 +254,8 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
     fpList = afwDetection.FootprintSet(maskedImage, thresh).getFootprints()
 
     satAreas = numpy.asarray([fp.getArea() for fp in fpList])
-    largeAreas, = numpy.where(satAreas >= itlEdgeBleedSatMinArea)
+    largeAreas, = numpy.where((satAreas >= itlEdgeBleedSatMinArea)
+                              & (satAreas < itlEdgeBleedSatMaxArea))
 
     satMaskBit = maskedImage.mask.getPlaneBitMask(saturatedMaskName)
 
--- a/python/lsst/ip/isr/isrFunctions.py
+++ b/python/lsst/ip/isr/isrFunctions.py
@@ -34,6 +34,7 @@ __all__ = [
     "gainContext",
     "getPhysicalFilter",
     "growMasks",
+    "maskITLEdgeBleed",
     "illuminationCorrection",
     "interpolateDefectList",
     "interpolateFromMask",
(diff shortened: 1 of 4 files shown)


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
                 badAmpDict[ampName] = True
                 maskView |= maskView.getPlaneBitMask("BAD")
 
+        # Edge bleed masking
+        if self.config.doITLEdgeBleedMask and detector.getPhysicalType() == 'ITL':
```
erykoff, line 796: Please change to double-quotes (matching the rest of the file).  Though I think this whole function should be moved to `isrFunctions`.  Also, it needs a test.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+
+            # Get footprint of saturated pixels
+            thresh = afwDetection.Threshold(limits[self.config.saturatedMaskName])
+            fpList = afwDetection.FootprintSet(maskedImage, thresh).getFootprints()
+
+            # Go through footprints to select saturated cores
+            for i in range(len(fpList)):
+                if fpList[i].getArea() > 10000.:
```
erykoff, line 804: This needs to be a config value.  Also, how was this determined?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+
+                    # Get centroid of saturated core
+                    xCore, yCore = fpList[i].getCentroid()
+                    # Get the number of saturated columns around the centroid
+                    widthSat = numpy.sum(fpList[i].getSpans().asArray(), axis=1)[int(yCore)]
```
erykoff, line 809: This seems wasteful to do a whole summation and then take one value.  Why not `numpy.sum(fpList[i].getSpans().asArray()[:, int(yCore)]` (or I might have this transposed but you get the idea).

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                    for amp in ccdExposure.getDetector():
+                        # Select 2 amplifiers around the saturated
+                        # core with a potential edge bleed
+                        if amp.getBBox().contains(xCore, yCore) \
+                                or amp.getBBox().contains(xCore, numpy.abs(4000.-yCore)):
```
erykoff, line 814: `4000.` should be the height of the data region.  Or are we in an untrimmed image?  So `4000.0` isn't right, it's the height of the full raw region, right?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                            satLevel = metadata[f"LSST ISR SATURATION LEVEL {ampName}"]
+
+                            if ampName[:2] == 'C1':
+                                # Check there is an edge bleed
+                                if numpy.mean(maskedImage.image.array[:50, int(xCore)-5:int(xCore)+5]) \
+                                        > 0.8*satLevel:
+                                    # Approximately check it is
+                                    # not a saturated edge bleed
+                                    if numpy.any(maskedImage.mask.array[:50, int(xCore)-5:int(xCore)+5]
+                                                 != maskVal):
```
erykoff, line 827: I don't understand this part.  Don't we need this even if it's partially saturated mask?  Or I guess this is checking if all pixels are already marked as saturated then we can skip?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                        # Select 2 amplifiers around the saturated
+                        # core with a potential edge bleed
+                        if amp.getBBox().contains(xCore, yCore) \
+                                or amp.getBBox().contains(xCore, numpy.abs(4000.-yCore)):
+
+                            # Get the saturation level in that amp
+                            ampName = amp.getName()
+                            satLevel = metadata[f"LSST ISR SATURATION LEVEL {ampName}"]
+
+                            if ampName[:2] == 'C1':
```
erykoff, line 820: I don't love having the amp names baked in, but this is a specific ITL task so it may be fine.  Will want to ask Chris after we get through a bit more review.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                                # Check there is an edge bleed
+                                if numpy.mean(maskedImage.image.array[:50, int(xCore)-5:int(xCore)+5]) \
+                                        > 0.8*satLevel:
```
erykoff, line 823: Where did 0.8 come from?  Is this empirical?  I'm not loving this because right now we define satLevel as the PTC turnoff, but we may increase it later depending.  Does this need to be smarter or configurable?

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                                                                                axis=1))
+
+                                        # Mask edge bleed with a
+                                        # decaying exponential model
+                                        for y in range(200):
+                                            edgeBleedHalfWidth = \
+                                                int(((maxWidthEdgeBleed)*numpy.exp(-0.05*y) + widthSat)/2.)
+                                            maskedImage.mask.array[y, int(xCore)-edgeBleedHalfWidth:
+                                                                   int(xCore)+edgeBleedHalfWidth] = \
+                                                maskedImage.mask.getPlaneBitMask("BAD")
```
erykoff, line 845: The mask bit we set here should be configurable.  Set to `BAD` now, but in the future we can change to a specific bit.

erykoff: How long does this loop take?  For loops make me nervous about performance.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                                        xmax = amp.getBBox().getX()
+                                        subImage = maskedImage.image.array[:150, xmax.min:xmax.max]
+                                        # Get an estimate of the width
+                                        # of the edge bleed
+                                        maxWidthEdgeBleed = numpy.max(numpy.sum(subImage > 0.5*satLevel,
+                                                                                axis=1))
+
+                                        # Mask edge bleed with a
+                                        # decaying exponential model
+                                        for y in range(200):
```
erykoff, line 840: Does this value need to be configurable?  Also the exp constant below in the model.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
+                                            edgeBleedHalfWidth = \
+                                                int(((maxWidthEdgeBleed)*numpy.exp(-0.05*y) + widthSat)/2.)
+                                            maskedImage.mask.array[y, int(xCore)-edgeBleedHalfWidth:
+                                                                   int(xCore)+edgeBleedHalfWidth] = \
+                                                maskedImage.mask.getPlaneBitMask("BAD")
+
+                            elif ampName[:2] == 'C0':
```
erykoff, line 847: I'm trying to figure out the difference between these two code blocks which are very similar.  I think we should be able to decide based on top/bottom a couple of signs and then merge the two.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -786,6 +792,82 @@ def maskSaturatedPixels(self, badAmpDict, ccdExposure, detector, detectorConfig,
...
 
+        # Edge bleed masking
+        if self.config.doITLEdgeBleedMask and detector.getPhysicalType() == 'ITL':
+
+            # Get footprint of saturated pixels
+            thresh = afwDetection.Threshold(limits[self.config.saturatedMaskName])
+            fpList = afwDetection.FootprintSet(maskedImage, thresh).getFootprints()
+
+            # Go through footprints to select saturated cores
+            for i in range(len(fpList)):
```
erykoff, line 803: I worry about the speed of this for loop ... though it may be fine.  But maybe it would be easier to pull out all the areas from fpList and then numpy check the array for those over 10000?

e.g.
```
areas = numpy.asarray([fp.getArea() for fp in fpList])
large, = numpy.where(areas >= minArea)
if len(large) == 0:
    return
```

erykoff (commented):
Here are some comments to start with.  This should be moved to its own function in `isrFunctions` and it needs a test in `test_isrFunctions.py`.  Though getting the camera geometry correct is going to be annoying, we may need to hack something together and I can help with that.
Also, you need to make sure this is turned off for calibration production, please set this to `False` in https://github.com/lsst/cp_pipe/blob/main/python/lsst/cp/pipe/_configure.py (in the correct operational order).

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +216,75 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
         fpSet.setMask(mask, maskValue)
 
 
+def maskEdgeBleed(ccdExposure, edgeBleedSatMinArea, saturatedMaskName):
+
+    maskedImage = ccdExposure.getMaskedImage()
```
erykoff, line 221: `ccdExposure.maskedImage`

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,73 +216,84 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
         fpSet.setMask(mask, maskValue)
 
 
-def maskEdgeBleed(ccdExposure, edgeBleedSatMinArea, saturatedMaskName):
+def maskITLEdgeBleed(ccdExposure, ITLEdgeBleedSatMinArea,
```
erykoff, line 219: These cannot lead with a capital letter.  Also we need a docstring.  Thanks!

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+    """Mask edge bleeds in ITL detectors.
+
+    Parameters
+    ----------
+    ccdExposure : `lsst.afw.image.Exposure`
+        Exposure to apply masking to.
+    itlEdgeBleedSatMinArea : scalar
+        Minimal saturated footprint area where the presence of edge bleeds
+        will be checked.
+    itlEdgeBleedSatFracLevel : scalar
```
erykoff, line 231: This should be "\`float\`"

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+       Fraction of the saturation level at the detector edge
+       above which there is an edge bleed.
+    itlEdgeBleedModelConstant : scalar
```
erykoff, line 234: Also "\`float\`".

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+        Constant in the decaying exponential in the edge bleed masking.
+    saturatedMaskName : `str`
+        Mask name for saturation.
+    badAmpDict : `str`[`bool`]
```
erykoff, line 238: This should be "\`dict\` [\`str\`, \`bool\`]"

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+    # Get footprint list of saturated pixels
+    thresh = afwDetection.Threshold(satLevel)
+    fpList = afwDetection.FootprintSet(maskedImage, thresh).getFootprints()
+
+    satAreas = numpy.asarray([fp.getArea() for fp in fpList])
+    largeAreas, = numpy.where(satAreas >= itlEdgeBleedSatMinArea)
+
+    if len(largeAreas) == 0:
+        return
+    else:
```
erykoff, line 257: No need for an `else` here.  The below can be shifted one level left, the other check just kicks back out anyway.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+        for largeAreasIndex in largeAreas:
+
+            # Get centroid of saturated core
+            xCore, yCore = fpList[largeAreasIndex].getCentroid()
+            # Turn the Y detector coordinate
+            # into Y footprint coordinate
+            yCoreFP = yCore - fpList[largeAreasIndex].getBBox().getMin().getY()
```
erykoff, line 264: I think `getBBox().getMinY()` does the same thing?

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+            # Get the number of saturated columns around the centroid
+            widthSat = numpy.sum(fpList[largeAreasIndex].getSpans().asArray()[int(yCoreFP), :])
+
+            for amp in ccdExposure.getDetector():
+                # Select 2 amplifiers around the saturated
```
erykoff, line 269: Please specify that these are two amplifiers "top and bottom" and not "side to side" (or a random two I suppose)

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+                # core with a potential edge bleed
+                yBox = amp.getBBox().getCenter()[1]
+                if amp.getBBox().contains(xCore, yBox):
```
erykoff, line 272: I'm missing how this picks up two amplifiers.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+                        sliceImage = numpy.flipud(maskedImage.image.array[-200:, :])
+                        sliceMask = numpy.flipud(maskedImage.mask.array[-200:, :])
+
+                    # The middle columns of edge bleeds are often quite close
+                    # to the saturation level so
+                    # we check there is an edge bleed by looking
+                    # at a small image up to 50 pixels from the edge
+                    # and around the saturated columns
+                    # of the saturated core, and checking its mean is
+                    # above 80 percent of the saturation level.
```
erykoff, line 298: Update the comment to say that 80 is the default.

tests/test_isrFunctions.py
```diff
@@ -104,6 +155,38 @@ def test_makeThresholdMask(self):
 
         self.assertEqual(len(defectList), 1)
 
+    def test_ITLEdgeBleedMask(self):
+        """Expect number of masked pixels according to edge bleed masking.
+        """
+        ITLdetector = TestITLDetector()
```
erykoff, line 161: Please don't capitalize object instances.  This needs to be `itlDetector` or even just `detector`.

erykoff: Similarly for the exposure below.

tests/test_isrFunctions.py
```diff
@@ -51,6 +51,57 @@ def computeImageMedianAndStd(image):
...
+
+class TestITLExposure(afwImage.ExposureF):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    # Monkey-patch our fake detector.
+    def setDetector(self, detector):
+        self._detector = detector
+
+    def fillMetadata(self, saturationLevel):
```
erykoff, line 95: Maybe `fillSaturationMetadata`?

tests/test_isrFunctions.py
```diff
@@ -104,6 +155,38 @@ def test_makeThresholdMask(self):
...
+            edgeBleedWidth = 50*np.exp(-0.05*y)+8
+            ITLexposure.image.array[y, 1100-int(edgeBleedWidth/2.)] = 15000.
+            ITLexposure.image.array[y, 1100+int(edgeBleedWidth/2.)] = 15000.
+            ITLexposure.image.array[y,
+                                    1100-int(edgeBleedWidth/2.)+1:1100+int(edgeBleedWidth/2.)
+                                    ] = 0.85*saturationLevel
+
+        badAmpDict = {}
+        for amp in ITLexposure.getDetector():
+            badAmpDict[amp.getName()] = False
```
erykoff, line 183: You should make one of the innocent amps bad, why not?

tests/test_isrTaskLSST.py
```diff
@@ -1069,6 +1069,7 @@ def test_isrFloodedSaturatedE2V(self):
         isr_config.doFlat = False
         # Tun off saturation masking to simulate a PTC flat.
         isr_config.doSaturation = False
+        isr_config.doITLEdgeBleedMask = False
```
erykoff, line 1072: After seeing all these overrides I wonder if the default should be False?  Though it may be fine but there needs to be a PR for cp_pipe.

tests/test_isrFunctions.py
```diff
@@ -104,6 +155,38 @@ def test_makeThresholdMask(self):
...
+        ipIsr.maskITLEdgeBleed(ITLexposure, itlEdgeBleedSatMinArea=10000.,
+                               itlEdgeBleedSatFracLevel=0.8, itlEdgeBleedModelConstant=0.05,
+                               saturatedMaskName='SAT', badAmpDict=badAmpDict)
+
+        self.assertEqual(ipIsr.countMaskedPixels(ITLexposure, 'SAT'), 2298)
```
erykoff, line 188: This needs some tests to make sure (a) there are saturated pixels at each edge; (b) I guess there was no original SAT mask set?

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,108 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
         fpSet.setMask(mask, maskValue)
 
 
+def maskITLEdgeBleed(ccdExposure, itlEdgeBleedSatMinArea,
```
erykoff, line 220: Looking at the tests below I think that everything except ccdExposure should be a keyword with a default.  It's a bit redundant with the task configs, but I think it'll be clearer here what the intended values are.

python/lsst/ip/isr/isrFunctions.py
aferte: It will pick up the 2 amplifiers that have the same x coordinate. We don't care about the y coordinate so we put the center of the amp bbox for y to simplify.

erykoff: Oh I see.  Can you explicitly add this to the comment?  Thanks!

python/lsst/ip/isr/isrFunctions.py
aferte: done except for ccdExposure and badAmpDict

tests/test_isrFunctions.py
```diff
@@ -104,6 +196,166 @@ def test_makeThresholdMask(self):
...
+                               saturatedMaskName='SAT', badAmpDict=badAmpDict)
+        # Number of saturated pixels at the bottom edge
+        # after applying edge bleed masking
+        numPixSatBottomEdgeAfter = len(np.where(exposure.mask.array[0, :] == satMaskBit)[0])
+        numPixSatTopEdgeAfter = len(np.where(exposure.mask.array[-1, :] == satMaskBit)[0])
+
+        # Check the number of saturated pixels
+        self.assertEqual(ipIsr.countMaskedPixels(exposure, 'SAT'), 171434)
+        # Check there are more pixels along the bottom edge after masking
+        self.assertGreaterEqual(numPixSatBottomEdgeAfter, numPixSatBottomEdgeBefore)
```
erykoff, line 355: If there should be more you want assertGreater not assertGreaterEqual

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,113 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+    """Mask edge bleeds in ITL detectors.
+
+    Parameters
+    ----------
+    ccdExposure : `lsst.afw.image.Exposure`
+        Exposure to apply masking to.
+    badAmpDict : `dict` [`str`, `bool`]"
+        Dictionary of amplifiers, keyed by name, value is True if
+        amplifier is fully masked.
+    itlEdgeBleedSatMinArea : int
```
erykoff, line 233: This should be "\`int\`, optional".  And no quote above on the dict.  Typing the backticks into the github comment box is a challenge.
Similar for the optional ones below.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,113 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
         fpSet.setMask(mask, maskValue)
 
 
+def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
+                     itlEdgeBleedSatFracLevel=0.8,
+                     itlEdgeBleedModelConstant=0.03,
+                     saturatedMaskName='SAT'):
```
erykoff, line 223: This should be double-quotes (matching the rest of the code in the file).

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,113 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+
+    # Get footprint list of saturated pixels
+    thresh = afwDetection.Threshold(satLevel)
+    fpList = afwDetection.FootprintSet(maskedImage, thresh).getFootprints()
+
+    satAreas = numpy.asarray([fp.getArea() for fp in fpList])
+    largeAreas, = numpy.where(satAreas >= itlEdgeBleedSatMinArea)
+
+    if len(largeAreas) == 0:
+        return
```
erykoff, line 258: An empty line after `return` would look better, but isn't strictly necessary.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -216,6 +217,113 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+                                                            axis=1))
+
+                    # Mask edge bleed with a
+                    # decaying exponential model
+                    for y in range(200):
+                        edgeBleedHalfWidth = \
+                            int(((maxWidthEdgeBleed)*numpy.exp(-itlEdgeBleedModelConstant*y)
+                                 + widthSat)/2.)
+                        sliceMask[y, int(xCore)-edgeBleedHalfWidth:int(xCore)+edgeBleedHalfWidth] = \
+                            maskedImage.mask.getPlaneBitMask(saturatedMaskName)
```
erykoff, line 324: You want to take `maskedImage.mask.getPlaneBitMask(saturatedMaskName)` and bring it up and out of the loop and set it to a variable so we don't have to do the same operation hundreds of times (which can be surprisingly slow in some cases).

tests/test_isrFunctions.py
```diff
@@ -51,6 +51,98 @@ def computeImageMedianAndStd(image):
     return (median, std)
 
 
+def setExposureSatCore(exposure, x, y, halfWidthX, halfWidthY,
```
erykoff, line 54: If this is an internal helper function without a docstring, put a leading underscore on the name.  (Even so it should have a docstring.  It'll help 6 months from now!)

tests/test_isrFunctions.py
```diff
@@ -51,6 +51,98 @@ def computeImageMedianAndStd(image):
...
+                       satVal, satMaskBit):
+    exposure.image.array[y-halfWidthY:y+halfWidthY,
+                         x-halfWidthX:x+halfWidthX] = satVal
+    exposure.mask.array[y-halfWidthY:y+halfWidthY,
+                        x-halfWidthX:x+halfWidthX] = satMaskBit
+
+
+def setExposureSatColumns(exposure, x, y, halfWidthX, limY,
+                          satVal, satMaskBit, isTop=False):
```
erykoff, line 63: Similarly here.

tests/test_isrFunctions.py
```diff
@@ -51,6 +51,98 @@ def computeImageMedianAndStd(image):
...
+def makeEdgeBleed(exposure, x, extentY, edgeBleedWidth,
+                  edgeBleedConstant, edgeBleedWidthLimit,
+                  saturationLevel, saturationFrac,
+                  satVal, satMaskBit, isTop=False):
+    for y in range(extentY):
+        edgeBleedWidthY = edgeBleedWidth*np.exp(-edgeBleedConstant*y) \
+            + edgeBleedWidthLimit
+        if isTop:
+            # For edge bleed in top amplifier
+            y = 3999-y
```
erykoff, line 85: 3999 doesn't need to be hard coded, it can come from the exposure, right?

tests/test_isrFunctions.py
```diff
@@ -104,6 +196,166 @@ def test_makeThresholdMask(self):
...
+
+        # Case 1:
+        # A saturated core in the bottom amplifier
+        # that has a bottom edge bleed
+        satVal = 15000.
+        x = 1100
+        y = 1800
+        halfWidthX = 40
+        halfWidthY = 70
+        HalfWidthColX = 3
```
erykoff, line 226: Please no leading capitals.

tests/test_isrFunctions.py
```diff
@@ -104,6 +196,166 @@ def test_makeThresholdMask(self):
...
+        exposure.fillSaturationMetadata(saturationLevel)
+
+        badAmpDict = {}
+        for amp in exposure.getDetector():
+            if amp.getName() == 'C10':
+                badAmpDict[amp.getName()] = True
+            else:
+                badAmpDict[amp.getName()] = False
+
+        # Case 1:
```
erykoff, line 218: The three cases are three separate tests, with separate checks that they work.  I know from the image that these work, but the tests would succeed even if only one of the test cases worked.

tests/test_isrFunctions.py
```diff
@@ -51,6 +51,172 @@ def computeImageMedianAndStd(image):
...
+
+    Parameters
+    ----------
+    exposure : `lsst.afw.image.Exposure`
+        Input exposure.
+    x : `int`
+        X coordinate for the center of the rectangular saturated core.
+    y : `int`
+        Y coordinate for the center of the rectangular saturated core.
+    halfWidthX: `int`
```
erykoff, line 66: Needs space before colon

erykoff: And below

erykoff (approved):
The numpy docstrings need to be fixed; otherwise looks great!

erykoff (approved):

PR #379, opened 2025-02-20
DM-48783: Add additional metadata keys for ISR stages.

erykoff:


python/lsst/ip/isr/isrTask.py
```diff
@@ -1401,6 +1401,13 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
         # updateVariance, applyGains
         # Check if needed during/after BFE correction, CTI correction.
         exposureMetadata["LSST ISR UNITS"] = "adu"
+        exposureMetadata["LSST ISR CROSSTALK APPLIED"] = False
+        exposureMetadata["LSST ISR LINEARIZER APPLIED"] = False
+        exposureMetadata["LSST ISR CTI APPLIED"] = False
+        exposureMetadata["LSST ISR BIAS APPLIED"] = False
+        exposureMetadata["LSST ISR DARK APPLIED"] = False
+        exposureMetadata["LSST ISR BF APPLIED"] = False
+        exposureMetadata["LSST ISR FLAT APPLIED"] = False
```
TallJimbo, line 1410: Would it be useful to make this last option a multiple-choice string instead of a boolean, e.g. "DOME" or "TWILIGHT" instead of `True`?  I think that's probably best as a separate ticket if you can make it backwards-compatible, since I think we'd want that propagated through the flats themselves.

erykoff: @TallJimbo I'm happy to file a follow-up ticket, but this would require some more metadata added to the flat itself on the type of flat since we don't actually record that anywhere (the twilight flats are processed through the same cp_pipe pipelines as dome flats, for example).  I'd call it `LSST ISR FLAT TYPE` or something.
@timj Yes it's in the provenance if the task was run via pipetask/bps/etc.  But sometimes the task has been run via notebook/other methods, and we have been tripped up by this many times in the past and there was much rejoicing in cp-land when I showed off this PR just to have something simple recorded in the metadata.

timj: Thanks. Provenance is available to notebooks so long as someone remembers to call `prov.add_input` to record it themselves.

TallJimbo: The full provenance here will let us figure out the mistake we made once we've identified that we made one.  I was hoping for something that would let us prevent the mistake from happening.  But this is very low-priority, because the alternate configurations I'm worried about guarding against are pretty far off.

PR #380, opened 2025-02-21
DM-49092: Fix ip_isr test failure on Linux ARM due to incorrect 32-bit float comparison

enourbakhsh:


erykoff (approved):

PFLeget (approved):
LGTM based on slack discussion.

PR #381, opened 2025-02-21
DM-49094: Convert brighterFatterCorrection to 64-bit operations

Alex-Broughton:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -597,13 +597,14 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain, ga
         kernelImage = afwImage.ImageD(kLx, kLy)
         kernelImage.getArray()[:, :] = kernel
         tempImage = image.clone()
+        tempImage = afwImage.ImageD(image.clone(), deep=True)
```
erykoff, line 600: I don't think the `clone()` is necessary here.  Can you check in a notebook?  Specifically, create a single-precision afw image, and make an `ImageD` like this without the clone, and make sure that a copy happened (setting a value isn't synchronized).

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -597,13 +597,14 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain, ga
...
 
         nanIndex = numpy.isnan(tempImage.getArray())
         tempImage.getArray()[nanIndex] = 0.
 
-        outImage = afwImage.ImageF(image.getDimensions())
-        corr = numpy.zeros_like(image.getArray())
-        prev_image = numpy.zeros_like(image.getArray())
+        outImage = afwImage.ImageD(image.getDimensions())
+        corr = numpy.zeros_like(image.getArray(), dtype=numpy.float64)
+        prev_image = numpy.zeros_like(image.getArray(), dtype=numpy.float64)
```
erykoff, line 607: These two lines should be `image.array` (the modern API)

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -642,6 +643,7 @@ def brighterFatterCorrection(exposure, kernel, maxIter, threshold, applyGain, ga
 
             if iteration > 0:
                 diff = numpy.sum(numpy.abs(prev_image - tmpArray), dtype=numpy.float64)
+                print(f"{iteration}: {diff}")
```
erykoff, line 646: Please remove debugging prints.

python/lsst/ip/isr/isrFunctions.py
erykoff: Or maybe (similar to below) `numpy.zeros(image.array.shape, dtype=numpy.float64)`.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -818,14 +820,14 @@ def fluxConservingBrighterFatterCorrection(exposure, kernel, maxIter, threshold,
         kLy, kLx = kernel.shape
         kernelImage = afwImage.ImageD(kLx, kLy)
         kernelImage.getArray()[:, :] = kernel
-        tempImage = image.clone()
+        tempImage = afwImage.ImageD(image.clone(), deep=True)
```
erykoff, line 823: I think this is an extra `clone`.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -818,14 +820,14 @@ def fluxConservingBrighterFatterCorrection(exposure, kernel, maxIter, threshold,
...
 
         nanIndex = numpy.isnan(tempImage.getArray())
         tempImage.getArray()[nanIndex] = 0.
 
-        outImage = afwImage.ImageF(image.getDimensions())
-        corr = numpy.zeros_like(image.getArray())
-        prevImage = numpy.zeros_like(image.getArray())
+        outImage = afwImage.ImageD(image.getDimensions())
+        corr = numpy.zeros_like(image.getArray(), dtype=numpy.float64)
+        prevImage = numpy.zeros_like(image.getArray(), dtype=numpy.float64)
```
erykoff, line 830: Same comment as above.

python/lsst/ip/isr/isrTask.py
```diff
@@ -1607,7 +1607,9 @@ def run(self, ccdExposure, *, camera=None, bias=None, linearizer=None,
                     self.config.brighterFatterApplyGain,
                     bfGains
                 )
-            if bfResults[1] == self.config.brighterFatterMaxIter - 1:
+            bfCorrIters = bfResults[1]
+            self.metadata["LSST ISR BF CORR ITERS"] = bfCorrIters
```
erykoff, line 1611: I think this should be `NITER` or maybe `N_ITER`

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1853,8 +1854,11 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
                                  "gains in the kernel are not the same as the gains stored"
                                  "in the PTC. Using the kernel gains.")
 
-            ccdExposure = self.applyBrighterFatterCorrection(ccdExposure, flat, dark, bfKernelOut,
-                                                             brighterFatterApplyGain, bfGains)
+            ccdExposure, bfCorrIters = self.applyBrighterFatterCorrection(ccdExposure, flat, dark,
+                                                                          bfKernelOut,
+                                                                          brighterFatterApplyGain,
+                                                                          bfGains)
+            metadata["LSST ISR BF CORR ITERS"] = bfCorrIters
```
erykoff, line 1861: Same here for the name.

python/lsst/ip/isr/isrTask.py
erykoff: Oh, I realize on my PR I call this `LSST ISR BFK` and we should make sure these are the same.  I'd rather not have an extra word here, so either `LSST ISR BFK NITER` (following DM-48783) or `LSST ISR BFCORR NITER` and I can update my PR.  (Same comment below)

erykoff (commented):
This needs tests in (at least) `test_isrTaskLSST` to check that the new number of iterations metadata is properly set.

python/lsst/ip/isr/isrFunctions.py
Alex-Broughton: I think it is, when I tried this in a notebook before, but I'll double-check

erykoff (approved):
I assume you've confirmed that the extra `clone` wasn't necessary?

erykoff (approved):

PR #383, opened 2025-03-06
DM-49308: update edge bleed detection and masking

aferte:


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -382,22 +382,23 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
...
         default=10000,
     )
     itlEdgeBleedSatMaxArea = pexConfig.Field(
         dtype=int,
-        doc="Threshold of saturated cores footprint area.",
+        doc="Maximum limit for saturated cores footprint area.",
         default=100000,
     )
-    itlEdgeBleedSatFracLevel = pexConfig.Field(
+    itlEdgeBleedThreshold = pexConfig.Field(
```
erykoff, line 393: This is a tricky one.  Renaming a config is basically not allowed; the old one has to be deprecated and removed after a version.  But since we're going to backport this ticket into v29, and this config will only have lasted a couple weeks and not been "released", this may be allowed.  I'm pinging @yalsayyad .

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -382,22 +382,23 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
...
         dtype=float,
-        doc="Threshold of saturated cores footprint area.",
-        default=0.8,
+        doc="Sky background threshold for "
+        "edge bleed detection.",
```
erykoff, line 396: This doc line does not need to be split.  We're allowed 110 characters.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: I spoke to @yalsayyad and she said to yolo and replace and not deprecate; we should post in pipelines-embargo that you may see this problem if you run weekly 11+ on top of weekly 9 outputs, but since this is in ISR anybody doing that would probably be starting from raws anyway.  But normally we'd have to do a full deprecation cycle for any renamed configs.

erykoff (approved):

PR #386, opened 2025-03-25
DM-49500: Add ITL dip masking to IsrTaskLSST

erykoff:


python/lsst/ip/isr/overscanAmpConfig.py
```diff
@@ -114,6 +114,46 @@ class OverscanDetectorConfig(pexConfig.Config):
             "NONE": "No dithering is performed.",
         },
     )
+    itlDipMinHeight = pexConfig.Field(
```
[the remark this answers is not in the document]
czwa: `IsrDetectorConfig`.  These are operating at the detector level, not the amp level, right?  I think I confused myself from the filename.

tests/test_isrFunctions.py
```diff
@@ -521,6 +522,98 @@ def test_ITLEdgeBleedMask(self):
...
+        # This needs to be in several sections because of the max width.
+        for i in range(12):
+            exposure.mask.array[:, i*100: i*100 + 50] |= satMaskValue
+        exposure.mask.array[:, 100: 100 + detectorConfig.itlDipMaxColsPerImage + 1] |= satMaskValue
+
+        maskBak = exposure.mask.array.copy()
+        with self.assertLogs(level=logging.WARNING) as cm:
+            ipIsr.isrFunctions.maskITLDip(exposure, detectorConfig)
+        self.assertEqual(len(cm[1]), 1)
+        self.assertIn("Too many (519) columns would be masked", cm[1][0])
```
czwa, line 614: Have I confused myself in expecting that I would expect the number of columns to be masked to be at least 600?  You've added 12 sets of 50 SAT columns, so I don't see how this is coming up with 519 (as I would expect 600 to be the floor).  Maybe I'm just confused on lines 606-608.  Either way, can I ask for a short clarifying comment for this final test (lines 604 and onward)?

czwa (approved):
One point where I think I'm confused (in the unit tests), and would appreciate a comment as to why I shouldn't be if possible.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -347,6 +349,92 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
...
+    largeHeights, = numpy.where(heights >= detectorConfig.itlDipMinHeight)
+
+    if len(largeHeights) == 0:
+        return
+
+    # Get the approximate image background.
+    approxBackground = numpy.median(exposure.image.array)
+    maskValue = 0
+    for maskPlaneName in maskPlaneNames:
+        maskValue |= exposure.mask.getPlaneBitMask(maskPlaneName)
```
czwa, line 390: You should be able to get the complete bitmask by `maskValue = exposure.mask.getPlaneBitMask(maskPlaneNames)` (example: https://github.com/lsst/cp_verify/blob/8f5961f9077d81a93a6cad45f2fd3402cd469aaf/python/lsst/cp/verify/verifyStats.py#L337)

tests/test_isrFunctions.py
```diff
@@ -521,6 +522,104 @@ def test_ITLEdgeBleedMask(self):
...
         self.assertGreater(numPixSatTopEdgeAfter, numPixSatTopEdgeBefore)
 
+    def test_itlDipMasking(self):
+        """Test the ITL dip masking."""
+        detector = MockITLDetector()
+        exposure = MockITLExposure(detector.getBBox())
+        exposure.setDetector(detector)
+        exposure.mask.array[:, :] = 0
+        exposure.mask.addMaskPlane("ITL_DIP")
+        dipMaskValue = exposure.mask.getPlaneBitMask("SUSPECT") \
```
czwa, line 532: Same suggestion here.

czwa (approved):

PR #387, opened 2025-03-26
DM-49604: Update ITL edge bleed masking

aferte:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -219,8 +219,56 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
         fpSet.setMask(mask, maskValue)
 
 
-def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
-                     itlEdgeBleedSatMaxArea=100000,
+def getLargeSatMidPoint(fpCore):
```
erykoff, line 222: This should be an underscore method because it's not in the public interface.  (It's also not in `__all__` which is good, but that would make it extra clear)

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -219,8 +219,56 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+    Parameters
+    ----------
+    fpCore : `lsst.afw.detection._detection.Footprint`
+        Footprint of saturated core.
+
+    Returns
+    -------
+    trueSegments : int
+        Number of saturated cores in the footprint.
+    XMidPointBBox: `list` [`int`]
```
erykoff, line 235: All variables should lead with lower case (-> `xMidPointBBox`)

erykoff: This isn't a "bounding box" as such, and so I don't understand the variable name.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -219,8 +219,56 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+        if value:
+            if not inTrueSegment:
+                indexSwitchTrue.append(i)
+                trueSegments += 1
+                inTrueSegment = True
+        elif inTrueSegment:
+            indexSwitchFalse.append(i)
+            inTrueSegment = False
+    # if there are more than one true segment in the central slice
+    # then there are several saturated core and trails in the footprint.
```
erykoff, line 260: -> "more than one true segments"
-> "several saturated cores and trails"

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -219,8 +219,56 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+    XMidPointBBox = [0]
+    # TODO: generalize to N cores.
```
erykoff, line 262: This should have a ticket number attached for a TODO.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -219,8 +219,56 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+    if trueSegments == 2:
+        XMidPointBBox.append(int((indexSwitchTrue[1] - indexSwitchFalse[0])/2))
+        XMidPointBBox.append(fpCore.getSpans().asArray().shape[1])
```
erykoff, line 265: I think if there are more than two we should log a warning (so this should have a log passed in) so it's recorded that there are too many cores for the code right now.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -219,8 +219,56 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
...
+
+    return trueSegments, XMidPointBBox
+
+
+def maskITLEdgeBleed(ccdExposure, badAmpDict,
+                     fpCore, nbCore, XMidPointBBox,
```
erykoff, line 271: Here and above please fix the variable name.

erykoff: Also this isn't in the docstring; I still am not sure what it is.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -406,10 +412,10 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         doc="Sky background threshold for edge bleed detection.",
         default=5000.,
     )
-    itlEdgeBleedModelConstant = pexConfig.Field(
+    itlEdgeBleedModelCst = pexConfig.Field(
```
erykoff, line 415: Why is this renamed?  And to something that is less clear.  Renaming configs is not something that we can do simply; it formally needs deprecation, etc.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1835,16 +1841,39 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
-                and self.config.doSaturation:
-            isrFunctions.maskITLEdgeBleed(ccdExposure=ccdExposure,
-                                          itlEdgeBleedSatMinArea=self.config.itlEdgeBleedSatMinArea,
-                                          itlEdgeBleedSatMaxArea=self.config.itlEdgeBleedSatMaxArea,
-                                          itlEdgeBleedThreshold=self.config.itlEdgeBleedThreshold,
-                                          itlEdgeBleedModelConstant=self.config.itlEdgeBleedModelConstant,
-                                          saturatedMaskName=self.config.saturatedMaskName,
-                                          badAmpDict=badAmpDict)
+        if (self.config.doITLSatSagMask or self.config.doITLEdgeBleedMask) \
+                and detector.getPhysicalType() == 'ITL':
```
erykoff, line 1845: This is much more logic than should be in the parent run method; why isn't this a function?  (Either a self.method or an isrFunction; an isrFunction is more testable)

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -244,109 +293,171 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
...
-                        if upperRange > xmax:
-                            upperRange = xmax
-                        sliceMask[y, lowerRange:upperRange] = satMaskBit
+        widthSat = numpy.sum(fpCore.getSpans().asArray()[int(yCoreFP), :])
+        modelITLEdgeBleed(ccdExposure, xCore,
+                          satLevel, widthSat, itlEdgeBleedThreshold,
+                          itlEdgeBleedModelConstant, saturatedMaskName)
+
+
+def modelITLEdgeBleed(ccdExposure, xCore,
```
erykoff, line 331: Is this an internal method?  Underscore?

erykoff: I don't understand the naming or organization now.  This does the masking, not the modeling it looks to me.

erykoff (commented):
I've done a first pass.  I don't understand the new structure at all, as it turns out.  The functions and variable names seem to have gotten out of sync, and logic has been split in (to me) a surprising way.  Also, please add logging as we had discussed.

python/lsst/ip/isr/isrFunctions.py
aferte: What do you mean it is not in the public interface? It is called in isrTaskLSST

erykoff: Oh, I didn't realize that.  Then it should be in `__all__`.

python/lsst/ip/isr/isrTaskLSST.py
aferte: Because of the possibility of two cores in a footprint, these masking now work on footprints rather than on the exposure but I only want trigger this loop if the masking configs are true.

erykoff: I mean below this is way too much logic for the core run method of `IsrTaskLSST`, and should be moved either into a method or an isr function.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -233,120 +232,218 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
...
+    log : `logging.Logger`
+        Logger to handle messages.
     """
-    maskedImage = ccdExposure.maskedImage
-    xmax = maskedImage.image.array.shape[1]
 
-    # Get minimum of amplifier saturation level
-    satLevel = numpy.min([ccdExposure.metadata[f"LSST ISR SATURATION LEVEL {amp.getName()}"]
-                          for amp in ccdExposure.getDetector() if not badAmpDict[amp.getName()]])
+    # Get median of amplifier saturation level
```
czwa, line 248: `log = log if log else logging.getLogger(__name__)`  either here, or when you call the log, to ensure that the warning isn't lost silently.

python/lsst/ip/isr/isrTaskLSST.py
czwa: I'd vote method, with the unit test being just a direct call to that method with a faked footprint that needs to be split.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1835,16 +1841,37 @@ def run(self, ccdExposure, *, dnlLUT=None, bias=None, deferredChargeCalib=None,
...
+            largeAreas, = numpy.where((satAreas >= self.config.itlEdgeBleedSatMinArea)
+                                      & (satAreas < self.config.itlEdgeBleedSatMaxArea))
+
+            for largeAreasIndex in largeAreas:
+
+                fpCore = fpList[largeAreasIndex]
+
+                # Edge bleed masking
+                if self.config.doITLEdgeBleedMask:
+                    isrFunctions.maskITLEdgeBleed(ccdExposure=ccdExposure,
```
czwa, line 1864: I think I'd suggest just putting `config = self.config` in this block, and using the 5 character shorter version to get around the lint.

czwa (commented):
I think my github window and the code got out of sync, so I want to submit this now.  I think all of these comments are still helpful.

python/lsst/ip/isr/isrTaskLSST.py
erykoff: If this goes into a method it'll be reduced at least one indentation level which will also help.

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -233,120 +233,224 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
...
-                        upperRange = int(xCore)+edgeBleedHalfWidth
-                        # If the edge bleed model goes outside the detector
-                        # we set the limit for the masking
-                        # to the edge of the detector
-                        if lowerRange < 0:
-                            lowerRange = 0
-                        if upperRange > xmax:
-                            upperRange = xmax
-                        sliceMask[y, lowerRange:upperRange] = satMaskBit
+    # TODO DM-49736
```
erykoff, line 442: Can you add a comment about what this ticket is a TODO for?

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -233,120 +233,224 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
...
+    fpCore : `lsst.afw.detection._detection.Footprint`
+        Footprint of saturated core.
     itlEdgeBleedThreshold : `float`, optional
         Threshold above median sky background for edge bleed detection
         (electron units).
     itlEdgeBleedModelConstant : `float`, optional
         Constant in the decaying exponential in the edge bleed masking.
     saturatedMaskName : `str`, optional
         Mask name for saturation.
+    log : `logging.Logger`
```
erykoff, line 245: This should have a `, optional`

python/lsst/ip/isr/isrFunctions.py
```diff
@@ -233,120 +233,224 @@ def maskITLEdgeBleed(ccdExposure, badAmpDict, itlEdgeBleedSatMinArea=10000,
...
+    widthSat: `float`
+        Width of the saturated core.
+    itlEdgeBleedThreshold : `float`, optional
+        Threshold above median sky background for edge bleed detection
+        (electron units).
+    itlEdgeBleedModelConstant : `float`, optional
+        Constant in the decaying exponential in the edge bleed masking.
+    saturatedMaskName : `str`, optional
+        Mask name for saturation.
+    log : `logging.Logger`
```
erykoff, line 344: This needs an `, optional` (since default is None) and below:
```
log = log if log else logging.getLogger(__name__)
```

erykoff (approved):
Some minor comment comments and a missing log initialization, otherwise good to go.  Yes, the tests aren't beautiful but they do seem to be testing the functionality.

czwa (approved):

PR #389, opened 2025-03-28
DM-49749: Fix support for 2 cores in footprint in edge bleed

aferte:


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -221,7 +221,9 @@ def growMasks(mask, radius=0, maskNameList=['BAD'], maskValue="BAD"):
 
 
 def maskITLEdgeBleed(ccdExposure, badAmpDict,
-                     fpCore, itlEdgeBleedThreshold=5000.,
+                     fpCore, itlEdgeBleedSatMinArea=10000,
+                     itlEdgeBleedSatMaxArea=100000,
+                     itlEdgeBleedThreshold=5000.,
```
erykoff, line 226: It's usually bad form to insert new keywords rather than put them at the end; it probably doesn't matter but I do want to flag that.

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -854,6 +854,8 @@ def maskITLSatEdgesAndColumns(self, exposure, badAmpDict):
                 isrFunctions.maskITLEdgeBleed(ccdExposure=exposure,
                                               badAmpDict=badAmpDict,
                                               fpCore=fpCore,
+                                              itlEdgeBleedSatMinArea=self.config.itlEdgeBleedSatMinArea,
+                                              itlEdgeBleedSatMaxArea=self.config.itlEdgeBleedSatMaxArea,
```
erykoff, line 858: Were these accidentally removed in the last ticket?

erykoff (commented):
I would feel better if there were specific tests for these edge cases where it was crashing when hitting the edge to ensure we don't make this mistake again.

python/lsst/ip/isr/isrTaskLSST.py
aferte: I moved the check on the footprint area in isrTaskLSST in the last ticket because I need it both for edge bleed and saturation sag masking, but I had overlooked that I need that check for the sub"footprints" as well (so in isrFunctions) when doing the check for several cores in a footprint. There is probably a better way to organize.

erykoff (approved):
Just wanted to confirm that if you run these updated tests *without* the code fixes then it crashes with the index error that we saw on-sky?

aferte:
I confirm, the new edge case in the unit test now fails without the code fixes.

PR #395, opened 2025-04-07
DM-49926: Add list of input exposure MJDs to PTC

Alex-Broughton:


erykoff (approved):
I think this is correct, though I wonder if we should have just added mjd as an aux value?  But this is also fine.
One key question: have you confirmed that you can read an old PTC with this new code, and that the mjd is all nan as expected?  (We don't have a way of auto-testing this).

PR #396, opened 2025-04-10
DM-44113: Test and update electrometer ingest for FITS files as currently written.

czwa:
The main blockers on this is now how reliable the blocks/scheduler/etc are at generating GROUPID values.  I think this is ready for review, with the intent that we'll run this manually until we know what that reliability is.

### DIFF
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -537,7 +537,8 @@ class IsrCalib(abc.ABC):
             for k, v in table.meta.items():
                 if isinstance(v, fits.card.Undefined):
                     table.meta[k] = None
-
+        primaryHeader = dict(fits.open(filename)[0].header)
+        tableList[0].meta.update(primaryHeader)  # This retains things it shouldn't, like 'SIMPLE'.
         calibClass = cls.determineCalibClass(tableList[0].meta, "readFits")
         return calibClass.fromTable(tableList, **kwargs)
 
--- a/python/lsst/ip/isr/calibType.py
+++ b/python/lsst/ip/isr/calibType.py
@@ -28,6 +28,7 @@ import warnings
 import yaml
 import numpy as np
 
+from astro_metadata_translator import merge_headers
 from astropy.table import Table
 from astropy.io import fits
 
@@ -537,9 +538,15 @@ class IsrCalib(abc.ABC):
             for k, v in table.meta.items():
                 if isinstance(v, fits.card.Undefined):
                     table.meta[k] = None
-        primaryHeader = dict(fits.open(filename)[0].header)
-        tableList[0].meta.update(primaryHeader)  # This retains things it shouldn't, like 'SIMPLE'.
+
         calibClass = cls.determineCalibClass(tableList[0].meta, "readFits")
+        if calibClass._OBSTYPE in ("PHOTODIODE", ):
+            # Merge primary header, as these types store information
+            # there.
+            with fits.open(filename) as hdul:
+                primaryHeader = hdul[0].header
+            tableList[0].meta = merge_headers([tableList[0].meta, primaryHeader], mode="first")
+
         return calibClass.fromTable(tableList, **kwargs)
 
     def writeFits(self, filename):
--- a/python/lsst/ip/isr/photodiode.py
+++ b/python/lsst/ip/isr/photodiode.py
@@ -178,7 +178,7 @@ class PhotodiodeCalib(IsrCalib):
             metadata.pop(key)
 
         # Do translations:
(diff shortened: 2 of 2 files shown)


python/lsst/ip/isr/calibType.py
```diff
@@ -527,7 +527,8 @@ def readFits(cls, filename, **kwargs):
             for k, v in table.meta.items():
                 if isinstance(v, fits.card.Undefined):
                     table.meta[k] = None
-
+        primaryHeader = dict(fits.open(filename)[0].header)
+        tableList[0].meta.update(primaryHeader)  # This retains things it shouldn't, like 'SIMPLE'.
```
erykoff, line 531: Can you use `from astro_metadata_translator import merge_headers` for this?

python/lsst/ip/isr/photodiode.py
```diff
@@ -169,15 +169,38 @@ def fromTable(cls, tableList):
...
+        location = metadata.pop("LOCATN", "NO_LOCATION")
+
+        if instrument == "Electrometer_index_201" and location == "AuxTel":
+            metadata["INSTRUME"] = "LATISS"
+        elif instrument == "Electrometer_index_101" and location == "MainTel":
+            metadata["INSTRUME"] = "LSSTCam"
+        else:
+            # This will cause problems in ingest, but we don't know
+            # what to associate it with.
+            metadata["INSTRUME"] = "Unknown"
```
erykoff, line 191: Is this really the correct location for this code?  (Though I'm not entirely sure where else).  And does this create any problems with the old photodiode data from Run7?

python/lsst/ip/isr/calibType.py
timj: Yes. Please. That gives you much more control over how you want to deal with duplicate headers.

python/lsst/ip/isr/calibType.py
```diff
@@ -527,7 +527,8 @@ def readFits(cls, filename, **kwargs):
             for k, v in table.meta.items():
                 if isinstance(v, fits.card.Undefined):
                     table.meta[k] = None
-
+        primaryHeader = dict(fits.open(filename)[0].header)
```
timj, line 530: I think this leaks a filehandle. Please use

```python
with fits.open(filename) as hdul:
    primaryHeader = hudl[0].header
```

PS we should consider at some point changing this to using ResourcePath so that butler can read them directly from S3 without the local file. Something that is entirely possible with `ResourcePath.to_fsspec` with `astropy.io.fits`.

erykoff: Hmm.  Are the files in the LFA?  In which case we have to use resources?

python/lsst/ip/isr/calibType.py
czwa: Done.  Will push after repeating my tests.

python/lsst/ip/isr/photodiode.py
czwa: It's was the least bad place when I was thinking it through, and since we only have photodiode/electrometer data for LSSTCam.  I've tried to keep the ingest code able to accept the older versions, but I don't think those populated `INSTRUME`.  I have changed this to default to `None` on the initial `pop`, and use that value here, whatever it is, if we couldn't determine the instrument from the if/elses.

python/lsst/ip/isr/calibType.py
czwa: The files are in LFA, but there's a `ResourcePath.as_local()` in the `obs_lsst` code, so the calib class only needs to handle local files.  That sentence inverts the logic; the calib class never needed to know where the file actually lived prior to dealing with the photodiode, and so the photodiode ingest code hides that from the calib class.

python/lsst/ip/isr/photodiode.py
```diff
@@ -169,15 +169,38 @@ def fromTable(cls, tableList):
             The calibration defined in the tables.
         """
         dataTable = tableList[0]
-
         metadata = dataTable.meta
+
+        # Dump useless entries that are carried over from merging
+        # HDU[0]'s header with the header from HDU[1] (which has the
+        # data table).
```
erykoff, line 176: Is this still necessary with the `merge_headers` tooling?

erykoff (approved):

python/lsst/ip/isr/photodiode.py
timj: Yes, but Astropy provides an API for this so I think when the header is first read in we need to call `primaryHeader.strip()` on it and then these lines won't be needed.

erykoff: Is this comment still relevant?

PR #398, opened 2025-04-11
DM-46887: Fix calibType metadata clobbering

erykoff:


python/lsst/ip/isr/calibType.py
```diff
@@ -362,17 +363,26 @@ def search(haystack, needles):
...
-        self._slotName = search(dictionary, ["SLOTNAME"])
-        self._detectorId = search(dictionary, ["DETECTOR", "detectorId"])
-        self._detectorName = search(dictionary, ["DET_NAME", "DETECTOR_NAME", "detectorName"])
-        self._detectorSerial = search(dictionary, ["DET_SER", "DETECTOR_SERIAL", "detectorSerial"])
-        self._filter = search(dictionary, ["FILTER", "filterName"])
-        self._calibId = search(dictionary, ["CALIB_ID"])
-        self._seqfile = search(dictionary, ["SEQFILE"])
-        self._seqname = search(dictionary, ["SEQNAME"])
-        self._seqcksum = search(dictionary, ["SEQCKSUM"])
+        if (value := search(dictionary, ["INSTRUME", "instrument"])) is not None:
```
timj, line 366: At least half of these can come straight out of ObservationInfo if you want to consider that as a future way of not being directly dependent on every FITS header being exactly what you need.

czwa (approved):
I had to think this through a few times, but I think I'm happy here.

PR #399, opened 2025-04-14
DM-50153: Make IsrTaskLSST generate a more helpful error message when it is missing a PTC.

erykoff:


RobertLuptonTheGood:
Looks good, test and all

PR #400, opened 2025-04-14
DM-50189: Add exposure metadata keywords to indicate if there was a sequencer mismatch

erykoff:


czwa (approved):
This all makes sense to me.

PR #405, opened 2025-04-18
DM-50259: Add checks for amp noise and safe bail-out of a full detector is bad.

erykoff:


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -730,6 +747,85 @@ def maskFullAmplifiers(self, ccdExposure, detector, defects, gains=None):
...
+
+            doMask = False
+            if ptc.noise[ampName] > self.config.ampNoiseThreshold:
+                self.log.info(
+                    "Amplifier %s has a PTC noise level of %.2f e-, above threshold.",
+                    ampName,
+                    ptc.noise[ampName],
+                )
+                doMask = True
+            else:
```
Alex-Broughton, line 811: Do we want this to be another if statement? Don't we want to check if the PTC noise is greater than `ampNoiseThreshold` OR `LSST ISR OVERSCAN RESIDUAL SERIAL STDEV {ampName}` is greater than `ampNoiseThreshold` ?

tests/test_isrTaskLSST.py
```diff
@@ -1717,6 +1831,7 @@ def get_isr_config_minimal_corrections(self):
         """Get an IsrTaskLSSTConfig with minimal corrections.
         """
         isr_config = IsrTaskLSSTConfig()
+        isr_config.ampNoiseThreshold = np.inf
```
Alex-Broughton, line 1834: Won't this always pass?

python/lsst/ip/isr/isrTaskLSST.py
erykoff: This is written in this way because I think it should be logged as an INFO if it is a "known" high noise amp (from the PTC) and a WARNING if it is a "new" high noise amp (differ from the PTC).  Also we need to do the gain correction for the overscan noise.

tests/test_isrTaskLSST.py
erykoff: Yes!  This is the test config "with minimal corrections" and that means overriding the default so that the amp always passes.

Alex-Broughton (approved):

PR #407, opened 2025-04-21
DM-50336: Add new PTC attributes for overscanMedian and overscanMedianAbsDev

Alex-Broughton:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -150,7 +148,13 @@ class PhotonTransferCurveDataset(IsrCalib):
...
     overscanMedianLevelList : `dict`, [`str`, `np.ndarray`]
         Dictionary keyed by amp names containing the median overscan
-        level from eachinput flat pair (units: adu).
+        level from each input flat pair (units: adu).
+    overscanMedian : `dict `, [`str`, `float`]
+        Dictionary keyed by amp names containing the median of
+        overscanMedianLevelList[expIdMask] (units: adu).
+    overscanMedianAbsDev : `dict `, [`str`, `float`]
+        Dictionary keyed by amp names containing the median absolute
+        deviation of overscanMedianLevelList[expIdMask] (units: adu).
```
erykoff, line 157: To be clear, this isn't the MAD exactly, but the scaled MAD (to be Gaussian equivalent).  So I think that this should actually be called `overscanMedianSigma` as estimated from the MAD.  (The problem with this request here though is that would require yet another rerun ... so we can discuss ... but I want this to be as clear as possible for the future).

erykoff (changes requested):
Sorry to be difficult on the naming here...

PR #409, opened 2025-04-23
DM-50257: Add checks to IsrTaskLSST for detectors and amps that are turned off.

erykoff:


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -243,13 +243,35 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         doc="Per-detector and per-amplifier overscan configurations.",
     )
 
+    serialOverscanMedianSigmaThreshold = pexConfig.Field(
```
Alex-Broughton, line 246: Maybe change this to `serialOverscanMedianShiftSigmaThreshold` or something like that, to not confuse it with the actual serial overscan sigma

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -605,6 +627,8 @@ def validate(self):
                 raise ValueError("Cannot run task with doBootstrap=True and doCorrectGains=True.")
             if self.doCrosstalk and self.crosstalk.doQuadraticCrosstalkCorrection:
                 raise ValueError("Cannot apply quadratic crosstalk correction with doBootstrap=True.")
+            if numpy.isfinite(self.serialOverscanMedianSigmaThreshold):
+                raise ValueError("Cannot do amp overscan level checks with doBootstrap=True.")
```
Alex-Broughton, line 631: I'm not sure why we can't do this with doBootstrap=True

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -772,6 +796,65 @@ def checkAllBadAmps(self, badAmpDict, detector):
...
+
+        # We want to consolidate all the amps into one warning if necessary.
+        # This config should not be set to a finite threshold if the necessary
+        # data is not in the PTC.
+        missingWarnString = "No PTC overscan information for amplifier "
+        missingWarnFlag = False
+        for amp in exposure.getDetector():
+            ampName = amp.getName()
+
+            if not numpy.isfinite(ptc.overscanMedian[ampName]) or \
```
Alex-Broughton, line 832: It might be nice to add some metadata to the image that describes if it checked for overscan outliers. But probably not necessary if you can just check that doBootstrap=True

python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -1599,6 +1682,29 @@ def ditherCounts(self, exposure, detectorConfig, fallbackSeed=12345):
...
+        voltage = exposure.metadata.get(self.config.bssVoltageKeyword, None)
+        if voltage is None or not numpy.isfinite(voltage):
+            self.log.warning(
+                "Back-side bias voltage %s not found in metadata.",
+                self.config.bssVoltageKeyword,
+            )
+            return
+
+        if voltage < self.config.bssVoltageMinimum:
+            raise UnprocessableDataError("Back-side bias voltage is turned off; skipping ISR.")
```
Alex-Broughton, line 1706: Add the detector #/name here?

tests/test_isrTaskLSST.py
```diff
@@ -83,6 +83,10 @@ def setUp(self):
             self.ptc.gain[amp_name] = mock.config.gainDict.get(amp_name, mock.config.gain)
             self.ptc.noise[amp_name] = mock.config.readNoise
             self.ptc.ptcTurnoff[amp_name] = self.saturation_adu
+            pre_level = mock.config.clockInjectedOffsetLevel
+            overscan_level = mock.config.biasLevel + (pre_level / self.ptc.gain[amp_name])
+            self.ptc.overscanMedian[amp_name] = overscan_level
+            self.ptc.overscanMedianSigma[amp_name] = 10.0
```
Alex-Broughton, line 89: Not important, but this is big. I checked a few real amps and the overscanMedianSigma was arouns ~0.8

python/lsst/ip/isr/isrTaskLSST.py
erykoff: Hmm.  It's not crazy to add something about that, because it can be turned off.

tests/test_isrTaskLSST.py
erykoff: This is not a real amp.  In the mocks we have a very large serial slope and a small overscan region which really messes up the statistics vs reality.  I can add a comment here.

PR #412, opened 2025-05-13
DM-50838: Add monitoring of HVBIAS keyword for IsrTaskLSST.

erykoff:


tests/test_isrTaskLSST.py
```diff
@@ -2037,6 +2037,99 @@ def test_bssVoltageChecks(self):
...
+                crosstalk=self.crosstalk,
+                ptc=self.ptc,
+                linearizer=self.linearizer,
+                defects=self.defects,
+                deferredChargeCalib=self.cti,
+            )
+
+        # Check that processing runs with other way of turning checks off.
+        isr_config.doCheckUnprocessableData = True
+        isr_config.bssVoltageMinimum = 0.0
```
aferte, line 2079: Isn't the voltage for that mock 50 V? In that case, the voltage is always greater than bssVoltageMinimum wetheir it is 0 or 10 as set here, or 5, the default.

aferte (commented):
Looks good to me, I just have one comment on the mock testing.

craiglagegit (commented):
I also looked at it, just to try and get back up to speed with reviewing these.  I ran a quick ISR in a notebook with 2025051100591, detector 76, which is one of the ones that tripped.  With the existing code, the error message is:
   1731 if voltage < self.config.bssVoltageMinimum:
   1732     detector = exposure.getDetector()
-> 1733     raise UnprocessableDataError(
   1734         f"Back-side bias voltage is turned off for {detector.getName()}; skipping ISR.",
   1735     )

UnprocessableDataError: Back-side bias voltage is turned off for R20_S11; skipping ISR.

With DM-50838, the message is:
   1745 if voltage < self.config.bssVoltageMinimum or hv == "OFF":
   1746     detector = exposure.getDetector()
-> 1747     raise UnprocessableDataError(
   1748         f"Back-side bias voltage is turned off for {detector.getName()}; skipping ISR.",
   1749     )

UnprocessableDataError: Back-side bias voltage is turned off for R20_S11; skipping ISR.

so it looks to be working.  CCD 94 with HVBias on processed normally.

tests/test_isrTaskLSST.py
erykoff: This part of the test is to set this config to 0.0 which turns off all voltage checking (including HVBIAS).   I could add a `doCheckVoltage` but it seemed redundant.

PR #420, opened 2025-07-11
DM-46261: Build ingestion for shutter motion profiles

czwa:


python/lsst/ip/isr/shutterMotion.py
```diff
@@ -0,0 +1,576 @@
...
+            position is midway between its starting and ending
+            locations.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if the requested ``modelName`` is not found in the
+            calibration.
+        """
+        # I think these are all "six-parameter" models.
```
erykoff, line 106: Can you confirm?

python/lsst/ip/isr/shutterMotion.py
```diff
@@ -0,0 +1,576 @@
...
+        self.hall_isOn = []
+        self.fit_name = []
+        self.fit_start_time = []
+        self.fit_pivot1 = []
+        self.fit_pivot2 = []
+        self.fit_jerk0 = []
+        self.fit_jerk1 = []
+        self.fit_jerk2 = []
+
+        self.requiredAttributes.update(['time_tai', 'time_mjd', 'position',
```
erykoff, line 69: There is a mix of single-quote and double-quote in this file.  I would request that you settle on one (and sadly, double-quote has won out in general).

python/lsst/ip/isr/shutterMotion.py
```diff
@@ -0,0 +1,576 @@
...
+        """
+        # I think these are all "six-parameter" models.
+        modelIndex = -1
+        for idx, name in enumerate(self.fit_name):
+            if name == modelName:
+                modelIndex = idx
+        if modelIndex == -1:
+            raise RuntimeError(f"Unknown model {modelName} requested.")
+
+        # Alias to follow technote
```
erykoff, line 114: Can you put a link to the technote here and in the docstring?

python/lsst/ip/isr/shutterMotion.py
```diff
@@ -0,0 +1,576 @@
...
+
+        Returns
+        -------
+        dictionary : `dict`
+            Dictionary of properties.
+        """
+        self.updateMetadata()
+        formatVersion = self.metadata['version']
+
+        if formatVersion == 1.0:
```
erykoff, line 226: Can you put a bit more about what the different format versions are about?

python/lsst/ip/isr/shutterMotion.py
```diff
@@ -0,0 +1,576 @@
...
+            outDict = {"fileName": self.metadata["fileName"],
+                       "fileType": self.metadata["fileType"],
+                       "metadata": {
+                           "CALIBCLS": "lsst.ip.isr.ShutterMotionProfile",
+                           "OBSTYPE": self._OBSTYPE, },
```
erykoff, line 231: I'm not insisting on black formatting, but the `}` should be on the next line.

python/lsst/ip/isr/shutterMotion.py
```diff
@@ -0,0 +1,576 @@
...
+                       "obsId": self.metadata["obsId"],
+                       "version": self.metadata.get("version", -1),
+                       "motionProfile": {
+                           "startTime": {
+                               "tai": {
+                                   "isot": self.metadata["startTime_tai"],
+                                   "mjd": self.metadata["startTime_mjd"], }, },
```
erykoff, line 260: And here please move the `}`s down.

python/lsst/ip/isr/shutterMotion.py
erykoff: There are a bunch of these.

erykoff (commented):
I've made a first pass of comments.  One of my bigger questions is ... are we supposed to be adding more sections to `docs/` as we make these updates?

czwa:
> I've made a first pass of comments. One of my bigger questions is ... are we supposed to be adding more sections to `docs/` as we make these updates?

I have a technote (sitcomtn-167) to document the file formats for shutter motion profiles, electrometer/photodiode, and fiber spectrograph data.  I think the docs directory here and in the cp_ packages need revision in general (neither IsrTask shows up in the task list, among other issues).

erykoff (approved):
Please add comments somewhere in the file about v1 vs v2, otherwise looks good.

python/lsst/ip/isr/shutterMotion.py
czwa: ```            # Original format.```
and
```            # Update to clarify all times are in the TAI system.```
Possibly not the most detailed, but this was really just a switch to ensure that it's clear that both the ISO and MJD are TAI based.

File: python/lsst/ip/isr/shutterMotion.py [hunk 7]
```diff
@@ -0,0 +1,589 @@
...
+            The calibration defined in the tables.
+        """
+        samples = tableList[0]
+        transitions = tableList[1]
+        modelFits = tableList[2]
+
+        metadata = samples.meta
+
+        calib = cls()
+        calib.time_tai = samples["TIME_TAI"]
```
erykoff, line 316: Do all of these need a `[0]` at the end? Maybe.  But I don't understand how the round trip test works, unless it's incomplete.

PR #426, opened 2025-09-03
DM-46501: Generalize BinExposureTask to nxn bin Image and MaskedImage type data

jrmullaney:
I changed the title of the ticket to better reflect the solution I've landed on.


python/lsst/ip/isr/binImageDataTask.py
```diff
@@ -0,0 +1,214 @@
+# This file is part of ip_isr.
```
czwa, line 1: I think github is telling me this was a `git mv` rename of `binExposureTask.py`.  Is that correct (I'm hoping it's correct).

czwa (approved):

python/lsst/ip/isr/binImageDataTask.py
jrmullaney: Yes - that's correct. I split the edits and the git mv into two separate commits, so you can see only the edits against the current main.

I thought it would be pretty cruel make the edits and the mv in the same commit!!

PR #427, opened 2025-09-14
DM-52242: Add the ptcRolloff and ptcRolloffError attributes to ptcDataset

Alex-Broughton:


python/lsst/ip/isr/ptcDataset.py
```diff
@@ -1335,31 +1393,22 @@ def evalPtcModel(self, mu):
         Computes the covModel for all mu, and it returns
         cov[N, M, M], where the variance model is cov[:,0,0].
         Both mu and cov are in ADUs and ADUs squared. This
-        routine evaulates the n-degree polynomial model (defined
-        by polynomialFitDegree) if self.ptcFitType == POLYNOMIAL,
-        the approximation in Eq. 16 of Astier+19 (1905.08677)
+        routine evaulates the approximation in Eq. 16 of
```
erykoff, line 1396: -> evaluates

python/lsst/ip/isr/ptcDataset.py
```diff
@@ -1335,31 +1393,22 @@ def evalPtcModel(self, mu):
...
         covModel = {ampName: np.array([]) for ampName in ampNames}
 
-        if self.ptcFitType == "POLYNOMIAL":
-            pars = self.ptcFitPars
-
-            for ampName in ampNames:
-                c00 = poly.polyval(mu, [*pars[ampName]])
-                covModel[ampName] = c00
-
-        elif self.ptcFitType == "EXPAPPROXIMATION":
```
erykoff, line 1362: I actually don't think we can remove this ... we've deprecated it, and I'm happy with removing the ability to *solve* polynomial (this is a grey area) but if anybody has a polynomial ptc out there they need to be able to use it.  So this should raise a FutureWarning if somebody uses this, but it should still work.  And link to the TODO removal ticket post v30.

PR #429, opened 2025-09-19
DM-52350: Add A23 BF Correction

Alex-Broughton:


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -105,6 +105,13 @@ class IsrTaskLSSTConnections(pipeBase.PipelineTaskConnections,
         dimensions=["instrument", "detector"],
         isCalibration=True,
     )
+    electroBfDistortionMatrix = cT.PrerequisiteInput(
+        name="electroBfDistortionMatrix",
+        doc="Electrostatic BF solution",
+        storageClass="ElectrostaticBrighterFatter",
```
erykoff: This should be `IsrCalib`.  It was a bug that it ever was something else.

czwa (approved):
I don't see any problems, other than the incorrect docstrings.  If you want to punt those so this can merge, file a ticket for that an add me as a watcher.

PR #436, opened 2025-11-11
DM-52156: Add defect reason extension to defects

czwa:


python/lsst/ip/isr/defects.py
```diff
@@ -513,6 +611,13 @@ def _get_values(values, n=1):
 
         return values[:n]
 
+    def getReasonDict(self):
```
erykoff: I would prefer if there were too methods here ... `getReasonToBitMapping()` and `getBitToReasonMapping()` or something like that?
Also this doesn't have a docstring.

tests/test_defects.py
```diff
@@ -65,6 +67,52 @@ def assertMetadata(self, first, second):
...
+        # the same mask
+        reason = 'VAMPIRE_PIXEL'
+        defects.maskPixelsReason(ccdImage.mask, reason)
+        self.assertEqual(numpy.sum(ccdImage.mask.array), numpy.sum(ccdImageMaskBefore))
+        # test that masking with an inexistent reason raises
+        reason = 'TEST_PIXEL'
+        with self.assertRaises(RuntimeError):
+            defects.maskPixelsReason(ccdImage.mask, reason)
+
+        # test 2. test mask plane with reasons is set properly
```
erykoff: What's the story here?

PR #441, opened 2026-02-12
DM-54146: Add IntrinsicZernikes IsrCalib

jmeyers314:


erykoff (commented):
A few questions/comments; looks good overall.

python/lsst/ip/isr/intrinsicZernikes.py
```diff
@@ -0,0 +1,253 @@
...
+                    table[col].to("um").value for col in zcols
+                ]
+            )
+            self.values = zks.reshape(self.field_y.size, self.field_x.size, -1)
+            self._createInterpolator()
+
+        self.requiredAttributes.update(["field_x", "field_y", "values", "noll_indices"])
+
+    def _createInterpolator(self):
+        self.interpolator = RegularGridInterpolator(
```
erykoff: Doesn't this have the problem that it might do the transpose?  The other interpolator is what we want to use.

erykoff (commented):
I think we should update the interpolator unless you can guarantee it won't have the same problem as ts_wep?

erykoff (approved):

jmeyers314 (commented):
I think we should explicitly state in the ctor docstring and the getIntrinsicZernikes docstring the coordsys of the field angles.

python/lsst/ip/isr/intrinsicZernikes.py
```diff
@@ -100,7 +104,7 @@ def __init__(self, table=None, **kwargs):
 
     def _createInterpolator(self):
         self.interpolator = LinearNDInterpolator(
-            np.column_stack((self.field_y, self.field_x)),
+            np.column_stack((self.field_x, self.field_y)),
```
erykoff, line 107: Oh this is fun.   `ts_wep` has `y, x`.  https://github.com/lsst-ts/ts_wep/blob/f22275acc781daf324d204cb7158a1537099f827/python/lsst/ts/wep/task/calcZernikesTask.py#L221

PR #442, opened 2026-02-18
DM-54023: Implement conversion weight depths for electrostatic BF

Alex-Broughton:


erykoff (commented):
Same comment as the other PR about the name "color correction" which doesn't seem quite correct.

erykoff (approved):
Looks good.  Have you confirmed that you can read in the old EBF with this code and run ISR correctly without warnings?

And now that I think about it ... will that warn?  I guess it will only warn if applyFilterCorrection is True, which is what we want. So I believe this is all correct, but please confirm on the jira ticket.

PR #443, opened 2026-04-16
DM-54543: Set variance plane in NO_DATA vignetted regions to a very large value.

erykoff:
This prevents possible detections in the fully vignetted corners.

This also makes sure that NO_DATA pixels aren't also marked BAD, preventing accidental interpolation and filling of pixels that should not have data.

### DIFF
--- a/python/lsst/ip/isr/isrTaskLSST.py
+++ b/python/lsst/ip/isr/isrTaskLSST.py
@@ -666,6 +666,12 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         doc="If flatScalingType is 'USER' then scale flat by this amount; ignored otherwise.",
         default=1.0,
     )
+    noDataVariance = pexConfig.Field(
+        dtype=float,
+        doc="Value to fill for variance plane in NO_DATA regions (which are generally "
+            "inherited from the flat in fully vignetted regions.)",
+        default=1e10,
+    )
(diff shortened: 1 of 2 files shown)


python/lsst/ip/isr/isrTaskLSST.py
```diff
@@ -666,6 +666,12 @@ class IsrTaskLSSTConfig(pipeBase.PipelineTaskConfig,
         doc="If flatScalingType is 'USER' then scale flat by this amount; ignored otherwise.",
         default=1.0,
     )
+    noDataVariance = pexConfig.Field(
+        dtype=float,
+        doc="Value to fill for variance plane in NO_DATA regions (which are generally "
+            "inherited from the flat in fully vignetted regions.)",
+        default=1e10,
+    )
```
laurenam, line 674: You didn’t want to adopt the suggestion of @TallJimbo to use the maximum image value to set the fill value?

laurenam (approved):

python/lsst/ip/isr/isrTaskLSST.py
TallJimbo: I think a configurable constant is probably better; less coupling between unrelated things.

erykoff: Yup!  This was discussed at pair programming.

PR #445, opened 2026-09-18
DM-56104: Write DECam edge-bleed masking task

ebellm:
Co-Authored-By: SLAC AI


python/lsst/ip/isr/isrFunctions.py
```diff
@@ -280,6 +281,144 @@ def maskE2VEdgeBleed(exposure, e2vEdgeBleedSatMinArea=10000,
...
+    saturatedMaskName : `str`, optional
+        Mask plane name for saturation.
+    log : `logging.Logger`, optional
+        Logger to handle messages.
+    """
+    log = log if log else logging.getLogger(__name__)
+
+    detector = exposure.getDetector()
+    if detector is None or len(detector) == 0:
+        log.warning("Cannot mask DECam edge bleeds: exposure has no detector with amplifiers.")
```
erykoff: Is this a warn or a raise?

python/lsst/ip/isr/isrTask.py
```diff
@@ -632,6 +632,55 @@ class IsrTaskConfig(pipeBase.PipelineTaskConfig,
         doc="Widen bleed trails based on their width?",
         default=True
     )
+    doDECamEdgeBleedMask = pexConfig.Field(
```
erykoff: I do not love adding so many configs to the parent isr task.  Does this need to be a subtask with a sub-config instead?

erykoff (commented):
I'm not comfortable adding so many config options to the main task which isn't DECam-specific; I think this needs to be a sub-task.

The other question I have is if this has been fully tested on detectors both on the top-half and the bottom-half of the DECam focal plane (upside-down).  Getting the readout corner stuff correct has tripped us up many times on LSSTCam at least.

python/lsst/ip/isr/isrFunctions.py
ebellm: Changed to a `RuntimeError` in 1eadb91b. The masking now only runs when the DECam subtask is explicitly retargeted in, so an exposure with no detector at that point is a misconfiguration rather than something to skip silently. That also matches the other `isrFunctions` that assume a detector is attached.

python/lsst/ip/isr/isrTask.py
ebellm: Moved into a subtask in 1eadb91b. The eight tuning parameters and the saturated mask plane name now live in `DECamEdgeBleedMaskConfig`, and `DECamEdgeBleedMaskTask` (in `masking.py`) is a drop-in retarget for the existing `IsrTaskConfig.masking` slot, whose docstring already says it is meant to be retargeted by a camera-specific task. `IsrTaskConfig` gains no new fields; the earlier `doDECamEdgeBleedMask` flag and `validate` check are gone. obs_decam will enable it with `config.doCameraSpecificMasking = True` and `config.masking.retarget(DECamEdgeBleedMaskTask)`.

ebellm:
On the config concern: the DECam-specific parameters are now in a `DECamEdgeBleedMaskTask` subtask that retargets the existing `masking` slot, so `IsrTaskConfig` is unchanged (see the inline thread and 1eadb91b).

On the readout corner question: the function keys on `amp.getReadoutCorner()` for each amplifier. For `UL`/`UR` the read edge is the top of the amp bbox in assembled coordinates and the row profile is reversed so index 0 is at that edge; for `LL`/`LR` it is the bottom. Both the candidate-footprint test (does it come within `approachRows` of the read edge) and the final mask (rows from the read edge inward) use that same orientation. In the obs_decam camera, all S detectors have amps with `UR`/`UL` corners and all N detectors have `LL`/`LR`, so both halves of the focal plane exercise different branches. The unit tests build a detector with a `UL` amp beside an `LL` amp and check that a bleed at each amp's read edge is masked, and that a dip at an amp's non-read edge is not.

ebellm:
Hi @erykoff, this is ready for another look.  The unit tests should cover the upper-half/lower-half concerns (see eg `test_dipInOtherAmpIsNotMasked`).

